PR #22, opened 2019-06-21
network: PushHeaders and BlockEvent to request missing blocks

mzabaluev:
These additions to the protocol enable retrieval of chain from the gRPC client upon solicitation from the service.

### DIFF
--- a/network-core/src/server/block.rs
+++ b/network-core/src/server/block.rs
@@ -76,11 +76,14 @@ pub trait BlockService: P2pService {
     /// implementation to produce a server-streamed response.
     type GetHeadersFuture: Future<Item = Self::GetHeadersStream, Error = Error> + Send + 'static;
 
+    /// The type of asynchronous futures returned by method `push_headers`.
+    type PushHeadersFuture: Future<Item = (), Error = Error> + Send + 'static;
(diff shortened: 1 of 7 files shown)


network-core/src/server/block.rs
```diff
@@ -111,8 +114,8 @@ pub trait BlockService: P2pService {
     /// to the server's tip.
     fn pull_blocks_to_tip(&mut self, from: &[Self::BlockId]) -> Self::PullBlocksToTipFuture;
 
-    /// Get block headers, walking forward in a range between any of the given
-    /// starting points, and the ending point.
+    /// Get block headers, walking the chain forward in a range between any
+    /// of the given starting points, and the ending point.
```
edolstra: Maybe this should be more specific about which starting point is used, e.g. the closest ancestor of the ending point that the server has available. Also, what happens if none of the starting points is known?

mzabaluev: If the filtered checkpoint list is empty, an empty stream is returned. In the push direction, we make a bogus `PushHeaders` call with an empty stream, too. Perhaps I should change this to send a `NotFound` error instead, good point.

PR #27, opened 2019-06-24
Generic ledger iterator

edolstra:
This iterator returns the entire contents of Ledger as a sequence of `Event`s, which can be Utxos, Accounts, and so on.

### DIFF
--- a/chain-impl-mockchain/src/accounting/account.rs
+++ b/chain-impl-mockchain/src/accounting/account.rs
@@ -272,3 +272,11 @@ impl<ID: Clone + Eq + Hash, Extra: Clone> Ledger<ID, Extra> {
         Iter(self.0.iter())
     }
 }
+
+impl<ID: Clone + Eq + Hash, Extra: Clone> std::iter::FromIterator<(ID, AccountState<Extra>)>
+    for Ledger<ID, Extra>
+{
+    fn from_iter<I: IntoIterator<Item = (ID, AccountState<Extra>)>>(iter: I) -> Self {
+        Ledger(Hamt::from_iter(iter))
+    }
+}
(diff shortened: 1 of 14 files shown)


chain-impl-mockchain/src/accounting/account.rs
```diff
@@ -39,10 +39,10 @@ impl From<InsertError> for LedgerError {
 
 #[derive(Clone)]
 pub struct AccountState<Extra> {
-    counter: SpendingCounter,
-    delegation: Option<StakePoolId>,
-    value: Value,
-    extra: Extra,
+    pub counter: SpendingCounter,
```
mzabaluev: Could we stop doing this everywhere and provide accessor methods instead?
As per API guidelines: https://rust-lang-nursery.github.io/api-guidelines/future-proofing.html#structs-have-private-fields-c-struct-private

chain-impl-mockchain/src/ledger.rs
```diff
@@ -853,3 +853,118 @@ fn input_account_verify(
         }
     }
 }
+
+pub enum Entry<'a> {
```
mzabaluev: This looks good, although `Globals` could just be a property.

chain-impl-mockchain/src/accounting/account.rs
edolstra: I don't know, hiding everything behind accessors strikes me a bit as Java-itis. If the internal representation of these types change, almost certainly the external interface will need to be changed as well. So the accessors are just boilerplate.

A more fundamental issue is whether the ledger iterator should provide access to these structs at all, since they're arguably somewhat tied to the current implementation.

PR #29, opened 2019-06-25
Property tests for address and ledger discrimination

### LINKED ISSUE #16: Testing Lvl1: check address discrimination in the ledger are checked
This is more like a unit test. We need to check that the ledger, when it applies a new fragment (a new transaction), verify the discrimination.

This is the code in here:

https://github.com/input-output-hk/rust-cardano/blob/master/chain-impl-mockchain/src/ledger.rs#L672-L674

We need to check that:

* if address 's discrimination is the same as the one in the params, it passes;
* if it differs it fails

Relates to #727

dkijania:
resolves #16

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
 /Cargo.lock
-**/target/
+/target/
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
 /Cargo.lock
-/target/
+**/target/
--- a/chain-impl-mockchain/tests/ledger/discrimination_tests.rs
+++ b/chain-impl-mockchain/tests/ledger/discrimination_tests.rs
@@ -1,0 +1,115 @@
+use crate::common::address::AddressData;
+use crate::common::ledger;
+use crate::common::ledger::ConfigBuilder;
+use crate::common::tx_builder::TransactionBuilder;
+use chain_addr::{Discrimination, Kind, KindType};
+use chain_impl_mockchain::transaction::*;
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -1,2 +1,2 @@
 /Cargo.lock
-/target/
+**/target/
```
mzabaluev: Do you regularly get a `target/` subdir created somewhere else with builds in subdirectories?
This might indicate an improperly configured workspace.

chain-impl-mockchain/tests/ledger/discrimination_tests.rs
```diff
@@ -0,0 +1,115 @@
...
+use chain_addr::{Discrimination, Kind, KindType};
+use chain_impl_mockchain::transaction::*;
+use chain_impl_mockchain::value::*;
+use quickcheck::{Arbitrary, Gen, TestResult};
+use quickcheck_macros::quickcheck;
+
+#[derive(Clone, Debug)]
+pub struct ArbitraryAddressKind(KindType);
+
+impl Arbitrary for ArbitraryAddressKind {
```
mzabaluev: There is already an `Arbitrary` impl for `KindType`, conditionally enabled on the `chain-addr` crate with feature `property-test-api`. Consider using that instead of a shim type?

dkijania: Yes, but this property is throwing `not implement yet` exception for `Multisig` value, so i created another without messing value

PR #30, opened 2019-06-26
Property test for initial transaction

dkijania:


chain-impl-mockchain/tests/common/address.rs
```diff
@@ -1,14 +1,30 @@
 use chain_addr::{Address, Discrimination, Kind, KindType};
-use chain_impl_mockchain::account::SpendingCounter;
-use chain_impl_mockchain::key::{EitherEd25519SecretKey, SpendingPublicKey};
+use chain_impl_mockchain::{
+    account::SpendingCounter,
+    key::{EitherEd25519SecretKey, SpendingPublicKey},
+    transaction::{Input, Output, UtxoPointer},
+    value::Value,
+};
+use std::fmt::{Debug, Formatter, Result};
```
mzabaluev: Please don't import `fmt::Result`, it makes the code confusing.
It's more common to import `std::fmt::self` and qualify with `fmt::Result` in usage.

chain-impl-mockchain/tests/common/address.rs
```diff
@@ -1,14 +1,30 @@
...
     pub spending_counter: Option<SpendingCounter>,
     pub address: Address,
 }
 
+impl Debug for AddressData {
+    fn fmt(&self, f: &mut Formatter) -> Result {
+        write!(
+            f,
+            "[{:?},{:?},{:?}]",
+            self.public_key, self.spending_counter, self.address
```
mzabaluev: Typically structs are written out using the `debug_struct` helper method on `Formatter` so that the fields are named in the output.

chain-impl-mockchain/tests/common/address.rs
```diff
@@ -24,6 +40,19 @@ impl AddressData {
         }
     }
 
+    pub fn as_input(&self, value: Value, utxo: UtxoPointer) -> Input {
```
mzabaluev: Conventionally, `as_` methods denote lightweight by-reference conversions.
This should be named `to_input` or something more meaningful to the subject matter, maybe `make_input`?

chain-impl-mockchain/tests/common/address.rs
```diff
@@ -24,6 +40,19 @@ impl AddressData {
...
+        match self.address.kind() {
+            Kind::Account { .. } => {
+                Input::from_account_public_key(self.public_key.clone(), value.clone())
+            }
+            _ => Input::from_utxo(utxo),
+        }
+    }
+
+    pub fn as_output(&self, value: Value) -> Output<Address> {
```
mzabaluev: Same as above here, `make_output`.

chain-impl-mockchain/tests/common/address.rs
```diff
@@ -24,6 +40,19 @@ impl AddressData {
         }
     }
 
+    pub fn as_input(&self, value: Value, utxo: UtxoPointer) -> Input {
+        match self.address.kind() {
+            Kind::Account { .. } => {
+                Input::from_account_public_key(self.public_key.clone(), value.clone())
+            }
+            _ => Input::from_utxo(utxo),
```
mzabaluev: This looks dodgy. The match should be exhaustive, so that adding a new address variant would make you revisit the implementation here. All the other kinds can be listed in the match arm separated with ` | `.

chain-impl-mockchain/tests/common/arbitrary/address.rs
```diff
@@ -0,0 +1,38 @@
...
+            0 => ArbitraryAddressKind(KindType::Single),
+            1 => ArbitraryAddressKind(KindType::Group),
+            2 => ArbitraryAddressKind(KindType::Account),
+            _ => unreachable!(),
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct ArbitraryAddressesData(pub Vec<AddressData>);
```
mzabaluev: This name seems like it will have a greppability problem when someone searches for `AddressData`.

chain-impl-mockchain/tests/common/arbitrary/address.rs
```diff
@@ -0,0 +1,38 @@
...
+#[derive(Clone, Debug)]
+pub struct ArbitraryAddressKind(pub KindType);
+
+impl Arbitrary for ArbitraryAddressKind {
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        match u8::arbitrary(g) % 3 {
+            0 => ArbitraryAddressKind(KindType::Single),
+            1 => ArbitraryAddressKind(KindType::Group),
+            2 => ArbitraryAddressKind(KindType::Account),
+            _ => unreachable!(),
```
mzabaluev: So what's the strategy for when a new `KindType` (ugh) appears? The test could silently fall into not producing really arbitrary values.

I would back this with an exhaustive match of the generated value, to make sure this impl is updated when the enum definition changes. If `Multisig` can't be properly handled at the moment, this impl can just ignore it and throw the dice again.

chain-impl-mockchain/tests/common/arbitrary/transaction.rs
```diff
@@ -0,0 +1,45 @@
+use crate::common::address::AddressData;
+use chain_addr::Address;
+use chain_impl_mockchain::transaction::Output;
+use chain_impl_mockchain::value::Value;
+use quickcheck::{Arbitrary, Gen};
+
+#[derive(Clone, Debug)]
+pub struct ArbitraryValue(pub Value);
+
+impl Arbitrary for ArbitraryValue {
```
mzabaluev: This could be defined for `Value` itself in the mockchain crate, under `mod testing` (does not have to be public) conditionally enabled on `cfg(feature = "property-test-api")` like in the other crates.

chain-impl-mockchain/tests/common/arbitrary/transaction.rs
```diff
@@ -0,0 +1,45 @@
...
+    fn arbitrary<G: Gen>(gen: &mut G) -> Self {
+        ArbitraryValue(Value(u64::arbitrary(gen)))
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct ArbitraryOutput(pub Output<Address>);
+
+impl Arbitrary for ArbitraryOutput {
```
mzabaluev: Same as above, no need to wrap these types at all if the property test API is implemented in the library crate.

chain-impl-mockchain/tests/common/address.rs
```diff
@@ -1,14 +1,30 @@
...
+use chain_impl_mockchain::{
+    account::SpendingCounter,
+    key::{EitherEd25519SecretKey, SpendingPublicKey},
+    transaction::{Input, Output, UtxoPointer},
+    value::Value,
+};
+use std::fmt::{Debug, Formatter, Result};
 
+#[derive(Clone)]
 pub struct AddressData {
```
mzabaluev: Can you document the purpose this type is serving for tests, which could not be addressed by the code under test?

chain-impl-mockchain/tests/common/address.rs
dkijania: I change it to `use std::fmt;` somehow `std::fmt::self;` throws error

mzabaluev: Sure; I meant in a grouped shorthand like `use std::fmt::{self, Debug};`.

chain-impl-mockchain/tests/common/arbitrary/address.rs
dkijania: Hmm but impl of  `Arbitrary for ArbitraryAddressKind` should handle situation with `Multisig` or code which is using it?

mzabaluev: I think you mentioned that the library itself currently panics on seeing `Multisig` in actual use, or something like that. If such is the case, this implementation, to be future-proof, needs to:
1. roll a random number in mod 4;
2. match it to all four values of `KindType`;
3. match the resulting enum value with an exhaustive set of patterns:

```rust
    use KindType::*;
    match v {
         Single | Group | Account => v,
         Multisig => {
             // Currently not supported by the library
             ... // use one of the above, or have a loop around the whole thing
                 // and use the continue statement here
         }
    }
```

chain-impl-mockchain/tests/common/arbitrary/output.rs
```diff
@@ -0,0 +1,28 @@
+use super::KindWithoutMultisig;
+use chain_addr::{Address, Discrimination};
+use chain_impl_mockchain::transaction::Output;
+use quickcheck::{Arbitrary, Gen};
+
+#[derive(Clone, Debug)]
+pub struct OutputsWithoutMultisig(pub Vec<Output<Address>>);
```
mzabaluev: Again, there could be a filter fn to get this without added types, in iterator style:

```rust
fn kind_without_multisig(item: Kind) -> Option<Kind> {
    use Kind::*; 
    match input {
        Multisig => None,
        x => Some(x),
    }
}
```

```rust
iter::from_fn(|| Some(Kind::arbitrary(gen)))
    .filter_map(kind_without_multisig)
    .take(n)
```

mzabaluev:
Build fails in CircleCI.

chain-impl-mockchain/tests/common/arbitrary/output.rs
mzabaluev: The generator might be difficult to use in this style, if needed in multiple closures.

chain-impl-mockchain/src/lib.rs
```diff
@@ -30,5 +30,13 @@ pub mod update;
...
 pub mod value;
 
-#[cfg(test)]
-mod tests {}
+#[macro_use]
+extern crate cfg_if;
+
+cfg_if! {
+   if #[cfg(test)] {
+        pub mod testing;
```
mzabaluev: I wonder why doesn't this work for in-tree test builds.

mzabaluev: There is an open issue about this: rust-lang/cargo#2911

A workaround is to also only enable tests that depend on the property api if the `property-test-api` feature is enabled (or maybe it's how it should be, since the property test API is an optional part of the crate and should be tested as such.) Then in CI, we can use the feature flag.

mzabaluev: On a better thought, we should not bother with test feature flags before looking into what may be the root of the problem here: wrong placement of test code. Any test code that depends on the property test impls in the source tree of the crate should belong in _unit_ tests (in `#[cfg(test)] mod tests { ... }`). Any generally useful test utilities that look like they might be used from other places than the in-crate tests should be moved into `mod testing` and made public.

chain-impl-mockchain/src/testing/genesis/genesis_tests.rs
```diff
@@ -1,14 +1,16 @@
-use crate::common::ledger as ledger_mock;
+#![cfg(test)]
```
mzabaluev: That makes it work, but now the unit tests are part of the `testing` module and their modules are also public, even though they won't be compiled for any users of crate.

The `#[quickcheck]` unit tests should be moved to `mod tests` in the library source, next to the types they are testing.

chain-impl-mockchain/src/ledger/mod.rs
```diff
@@ -0,0 +1,9 @@
+pub mod ledger;
+
+pub use ledger::*;
+
+cfg_if! {
+   if #[cfg(test)] {
+        pub mod tests;
```
mzabaluev: This looks like quite an elaborate way to use the conditional `cfg` attribute on a module :) `cfg_if!` is typically used when there are alternative definitions that otherwise would be tedious.

PR #33, opened 2019-07-03
[Tests] Property test:  the blockchain state total value is constant

### LINKED ISSUE #15: Testing Lvl1: the blockchain state total value is constant
We need to be able to check that, for any new transaction applied to the state, the total value remains constant.

This task is a bit more complex as you need to create only valid transaction for a given state of the blockchain. but still possible:

* for an arbitrary state of the blockchain (random UTxO, accounts, ledger parameters...);
* for an arbitrary *valid* transaction;

apply the transaction on the blockchain state;

* the total value should not have changed

relates to epic #727

dkijania:
Resolves #15

### DIFF
--- a/chain-impl-mockchain/Cargo.toml
+++ b/chain-impl-mockchain/Cargo.toml
@@ -39,6 +39,5 @@ chain-crypto = { path = "../chain-crypto", features=["property-test-api"]}
 chain-addr = { path = "../chain-addr", features=["property-test-api"]}
 ed25519-bip32 = "0.1"
 rand_chacha = "0.1"
-rand = "0.6.0"
 lazy_static = "1.3.0"
 rand = "0.6"
--- a/chain-impl-mockchain/Cargo.toml
+++ b/chain-impl-mockchain/Cargo.toml
@@ -39,5 +39,6 @@ chain-crypto = { path = "../chain-crypto", features=["property-test-api"]}
 chain-addr = { path = "../chain-addr", features=["property-test-api"]}
 ed25519-bip32 = "0.1"
 rand_chacha = "0.1"
+rand = "0.6.0"
 lazy_static = "1.3.0"
 rand = "0.6"
--- a/chain-impl-mockchain/src/ledger/tests/ledger_tests.rs
+++ b/chain-impl-mockchain/src/ledger/tests/ledger_tests.rs
@@ -1,16 +1,19 @@
 #![cfg(test)]
 
 use crate::testing::address::AddressData;
+use crate::testing::arbitrary::AribtraryValidTransactionData;
 use crate::testing::ledger;
 use crate::testing::ledger::ConfigBuilder;
 use crate::testing::tx_builder::TransactionBuilder;
-use chain_addr::Discrimination;
-use chain_impl_mockchain::ledger::{
-    Error::{NotEnoughSignatures, TransactionHasTooManyOutputs},
-    Ledger,
+use crate::{
+    ledger::{
+        Error::{NotEnoughSignatures, TransactionHasTooManyOutputs},
+        Ledger,
+    },
+    transaction::*,
+    value::*,
 };
-use chain_impl_mockchain::transaction::*;
-use chain_impl_mockchain::value::*;
+use chain_addr::Discrimination;
 use crate::ledger::{Entry, Ledger};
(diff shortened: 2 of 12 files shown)


chain-impl-mockchain/Cargo.toml
```diff
@@ -39,4 +39,5 @@ chain-crypto = { path = "../chain-crypto", features=["property-test-api"]}
 chain-addr = { path = "../chain-addr", features=["property-test-api"]}
 ed25519-bip32 = "0.1"
 rand_chacha = "0.1"
+rand = "0.6.0"
```
mzabaluev: I think @vincenthz split the rand deps partly to improve build times. Is it possible to depend on smaller sub-crates here?

chain-impl-mockchain/src/ledger/tests/ledger_tests.rs
```diff
@@ -31,158 +39,124 @@ macro_rules! assert_err {
...
-    let (message, utxos) = ledger::create_initial_transaction(Output::from_address(
-        faucet.address.clone(),
-        Value(42000),
-    ));
+#[quickcheck]
+pub fn ledger_accepts_correct_transaction(
+    faucet: AddressData,
+    receiver: AddressData,
+) -> TestResult {
+    let value = Value(100);
```
mzabaluev: Can this be generated with `Arbitrary`?

chain-impl-mockchain/src/testing/arbitrary/transaction.rs
```diff
@@ -0,0 +1,197 @@
...
+        let output_addresses_values =
+            Self::choose_random_output_subset(&addresses_values, total_input_value);
+        AribtraryValidTransactionData::new(
+            addresses_values,
+            input_addresses_values,
+            output_addresses_values,
+        )
+    }
+
+    fn generate_random_values() -> impl Iterator<Item = Value> {
```
mzabaluev: Why do this and below functions need to be methods on `ArbitraryValidTransactionDataBuilder`? They don't depend on this type on any way.

chain-impl-mockchain/src/testing/arbitrary/transaction.rs
```diff
@@ -0,0 +1,197 @@
...
+        AribtraryValidTransactionDataBuilder { source: source }
+    }
+
+    pub fn build(&self) -> AribtraryValidTransactionData {
+        let values: Vec<Value> = Self::generate_random_values()
+            .take(self.source.len())
+            .collect();
+
+        let addresses_values: Vec<AddressDataValue> =
+            Self::zip_addresses_and_values(&self.source, values);
```
mzabaluev: So the list of addresses is always the same and in the same order, only the values are random between invocations of `build` on the same builder instance? Is this the intent here?

chain-impl-mockchain/src/testing/arbitrary/transaction.rs
```diff
@@ -0,0 +1,197 @@
...
+use chain_addr::Address;
+use quickcheck::{Arbitrary, Gen};
+use std::iter;
+
+extern crate rand;
+use super::ArbitraryAddressDataVec;
+use rand::Rng;
+
+#[derive(Clone, Debug)]
+pub struct AribtraryValidTransactionData {
```
vincenthz: typo in Arbitrary and all the other below

chain-impl-mockchain/Cargo.toml
vincenthz: yes, but more to the case, this should use quickcheck arbitrary so that we make sure to have a deterministic result, not `rand`

chain-impl-mockchain/src/testing/arbitrary/transaction.rs
dkijania: Hm, no. list of addresses is also arbitraty because its generated from ArbitraryValidTransactionData

mzabaluev: Yes, but it will be always the same from this instance of `...Builder`, because it is generated once and then baked in? If you don't use the builder to generate transaction data repeatedly, why even have the separate builder type?

dkijania: ok. removed the builder.

Issue #21, opened 2019-06-21
Active Slot coefficient value is not check before applying it to the settings


NicolasDP:
https://github.com/input-output-hk/chain-libs/blob/e89e48a973f6fdfc5789a14d07a29dac4dddb0e4/chain-impl-mockchain/src/setting.rs#L83-L85

and it is not checked from decoding the value from the config param either because we use `Milli`:

https://github.com/input-output-hk/chain-libs/blob/e89e48a973f6fdfc5789a14d07a29dac4dddb0e4/chain-impl-mockchain/src/config.rs#L328-L336

PR #35, opened 2019-07-04
Hermetize ActiveSlotsCoeff

CodeSandwich:
Fixes https://github.com/input-output-hk/chain-libs/issues/21

It turned out that the structure API was safe, but because it wasn't hermetic, it was bypassed causing security threat

### DIFF
--- a/chain-impl-mockchain/src/leadership/genesis/vrfeval.rs
+++ b/chain-impl-mockchain/src/leadership/genesis/vrfeval.rs
@@ -51,7 +51,7 @@ impl Error for ActiveSlotsCoeffError {}
 /// Described in Ouroboros Praos paper, also referred to as parameter F of phi function
 /// Always in range (0, 1]
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
-pub struct ActiveSlotsCoeff(pub(crate) Milli);
+pub struct ActiveSlotsCoeff(Milli);
 
 impl TryFrom<Milli> for ActiveSlotsCoeff {
     type Error = ActiveSlotsCoeffError;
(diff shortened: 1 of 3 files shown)


chain-impl-mockchain/src/setting.rs
```diff
@@ -79,7 +79,7 @@ impl Settings {
                     new_state.epoch_stability_depth = *d;
                 }
                 ConfigParam::ConsensusGenesisPraosActiveSlotsCoeff(d) => {
-                    new_state.active_slots_coeff = ActiveSlotsCoeff(*d);
```
NicolasDP: much better

chain-impl-mockchain/src/leadership/genesis/vrfeval.rs
```diff
@@ -65,6 +65,12 @@ impl TryFrom<Milli> for ActiveSlotsCoeff {
     }
 }
 
+impl Into<Milli> for ActiveSlotsCoeff {
```
NicolasDP: I always thought the consensus was to implement `From<ActiveSlotCoeff> for Milli` rather than implementing `Into<Milli> for ActiveSlotCoeff`. Since one implies the other.

mzabaluev: Yes, prever implementing `From` whenever you can. It can always be done also in the crate defining the parameter type.

chain-impl-mockchain/src/leadership/genesis/vrfeval.rs
```diff
@@ -65,6 +65,12 @@ impl TryFrom<Milli> for ActiveSlotsCoeff {
...
+    fn into(self) -> Milli {
+        self.0
+    }
+}
+
 impl Into<f64> for ActiveSlotsCoeff {
```
mzabaluev: Same here, actually.

PR #38, opened 2019-07-05
Address removal of environment prefix

vincenthz:
Remove hardcoded address bech32 prefix

* no more testing/production difference
* display output addr in base32 form instead of bech32
* add arbitrary prefix 'qaddr'
* make 'ca' prefix in unit test to prevent rewriting
* remove environment gathering build.rs

### DIFF
--- a/chain-addr/src/lib.rs
+++ b/chain-addr/src/lib.rs
@@ -300,17 +300,15 @@ fn is_valid_data(bytes: &[u8]) -> Result<(Discrimination, KindType), Error> {
 pub struct AddressReadable(String);
 
 impl AddressReadable {
-    const PRODUCTION_ADDRESS_PREFIX: &'static str = env!("PRODUCTION_ADDRESS_PREFIX");
-
     pub fn as_string(&self) -> &str {
         &self.0
     }
 
     /// Validate from a String to create a valid AddressReadable
-    pub fn from_string(s: &str) -> Result<Self, Error> {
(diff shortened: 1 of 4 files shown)


chain-addr/src/lib.rs
```diff
@@ -300,40 +300,36 @@ fn is_valid_data(bytes: &[u8]) -> Result<(Discrimination, KindType), Error> {
...
-        let (discrimination, _) = is_valid_data(&dat[..])?;
-        if discrimination != expected_discrimination {
-            return Err(Error::MismatchPrefix);
-        }
+        let _ = is_valid_data(&dat[..])?;
+
+        Ok(AddressReadable(s.to_string()))
+    }
+
+    pub fn from_string_anyprefix(s: &str) -> Result<Self, Error> {
```
mzabaluev: This method and the one above should be named with `str` instead of `string`, which  conventionally refers to dealing with owned `String`.

Bikeshed: `from_str_with_prefix` and `from_str_any_prefix`?

NicolasDP: can we have a function to extract the _prefix_ too from the `AddressReadable`? This will make things useful for displaying a better error message to the users (and checking they are utilising the appropriate _prefix_ in a given context).

vincenthz: too late to rename things, but I added a duplicate function for now dealing with the `str/String` and deprecate the other function later.

PR #47, opened 2019-07-10
Add (somewhat) generic Digest type

vincenthz:


chain-crypto/src/digest.rs
```diff
@@ -0,0 +1,244 @@
...
+pub struct TryFromSliceError(());
+
+impl fmt::Display for TryFromSliceError {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Display::fmt("could not convert slice to digest", f)
+    }
+}
+
+pub trait HashingAlg {
```
mzabaluev: Bikeshed: name `DigestAlg` to avoid any confusion with `std::hash::Hash`?

vincenthz:
The typing in rust-crypto is very weak

PR #48, opened 2019-07-10
Message to fragment rename

vincenthz:
rename most Message to Fragment


chain-impl-mockchain/src/block/builder.rs
```diff
@@ -66,24 +66,24 @@ impl BlockBuilder {
...
         &mut self,
         signed_transaction: AuthenticatedTransaction<Address, NoExtra>,
     ) -> &mut Self {
-        self.message(Message::Transaction(signed_transaction))
+        self.message(Fragment::Transaction(signed_transaction))
     }
 
     /// add a message in the block to build
-    pub fn message(&mut self, message: Message) -> &mut Self {
+    pub fn message(&mut self, message: Fragment) -> &mut Self {
```
mzabaluev: Should this method and `messages` below be also renamed?

vincenthz: yes, but I don't want to do too many changes at once

PR #50, opened 2019-07-11
Role crypto

vincenthz:
* Add role for public key and secret key under the `role` module (to be kept qualified)
* Add `DigestOf` which qualify what is applied to, similar to `Signature`

### DIFF
--- a/chain-crypto/src/digest.rs
+++ b/chain-crypto/src/digest.rs
@@ -137,6 +137,7 @@ impl DigestAlg for Sha3_256 {
     }
 }
 
+/// A Digest Context for the H digest algorithm
 pub struct Context<H: DigestAlg>(H::DigestContext);
 
 impl<H: DigestAlg> Clone for Context<H> {
@@ -146,14 +147,17 @@ impl<H: DigestAlg> Clone for Context<H> {
 }
 
 impl<H: DigestAlg> Context<H> {
+    /// Create a new digest context
     pub fn new() -> Self {
         Self(H::new())
     }
 
+    /// Append data in the context
     pub fn append_data(&mut self, data: &[u8]) {
         H::append_data(&mut self.0, data)
     }
 
+    /// Finalize a context and create a digest
     pub fn finalize(self) -> Digest<H> {
         Digest(H::finalize(self.0))
     }
--- a/chain-crypto/src/digest.rs
+++ b/chain-crypto/src/digest.rs
@@ -246,3 +246,80 @@ impl<H: DigestAlg> fmt::Debug for Digest<H> {
         f.write_str(")")
     }
 }
+
+impl<H: DigestAlg> Digest<H> {
+    /// Get the digest of a slice of data
+    pub fn digest(slice: &[u8]) -> Self {
+        let mut ctx = Context::new();
+        ctx.append_data(slice);
+        ctx.finalize()
+    }
+}
+
+use std::marker::PhantomData;
+
+/// A typed version of Digest
+pub struct DigestOf<H: DigestAlg, T> {
+    inner: Digest<H>,
+    marker: PhantomData<T>,
+}
+
+impl<H: DigestAlg, T> Clone for DigestOf<H, T> {
+    fn clone(&self) -> Self {
+        DigestOf {
+            inner: self.inner.clone(),
+            marker: self.marker.clone(),
+        }
+    }
+}
+
+impl<H: DigestAlg, T> From<DigestOf<H, T>> for Digest<H> {
+    fn from(d: DigestOf<H, T>) -> Self {
+        d.inner
+    }
+}
+
+impl<H: DigestAlg, T> From<Digest<H>> for DigestOf<H, T> {
+    fn from(d: Digest<H>) -> Self {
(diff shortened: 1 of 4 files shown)


chain-crypto/src/digest.rs
```diff
@@ -242,3 +246,80 @@ impl<H: DigestAlg> fmt::Debug for Digest<H> {
...
+    }
+}
+
+impl<H: DigestAlg, T> From<DigestOf<H, T>> for Digest<H> {
+    fn from(d: DigestOf<H, T>) -> Self {
+        d.inner
+    }
+}
+
+impl<H: DigestAlg, T> From<Digest<H>> for DigestOf<H, T> {
```
mzabaluev: What is this transformation for? Presumably `DigestOf` is used to calculate digest of a particular type of value, but here it can also be obtained from an arbitrary digest?

chain-crypto/src/digest.rs
```diff
@@ -242,3 +246,80 @@ impl<H: DigestAlg> fmt::Debug for Digest<H> {
...
+        DigestOf {
+            inner: d,
+            marker: PhantomData,
+        }
+    }
+}
+
+impl<H: DigestAlg, T> DigestOf<H, T> {
+    /// Coerce a digest of T, to a digest of U
+    pub fn coerce<U>(&self) -> DigestOf<H, U> {
```
mzabaluev: This also looks odd. Should any untyped comparisons and the like be done on untyped `Digest`?

chain-crypto/src/digest.rs
```diff
@@ -242,3 +246,80 @@ impl<H: DigestAlg> fmt::Debug for Digest<H> {
...
+        DigestOf {
+            inner: ctx.finalize(),
+            marker: PhantomData,
+        }
+    }
+
+    /// Get the digest of object T, given its serialization function in closure
+    pub fn digest_with<F>(obj: &T, f: F) -> Self
+    where
+        F: FnOnce(&T) -> &[u8],
```
mzabaluev: I think this is problematic for several reasons.

The closure's signature assumes there is a contiguous byte slice inside the object and that is the only content that needs to be hashed. There really should be a trait for that and the types define the way they are hashed with.

Also, with this function there is no invariant to how `DigestOf` has been calculated, potentially leading to bugs down the line where digests are compared.

I once designed [a little crate](https://crates.io/crates/digest-hash) to deal with cryptographically stable hashing; check out the design.

chain-crypto/src/digest.rs
NicolasDP: a `DigestOf<T>` is a `Digest` but tied to a given type. You state this is the digest of a `T`.
For example, the FragmentId is a `DigestOf<Fragment>`. Here we allow users to tag any arbitrary `Digest` with any arbitrary type `T`.

chain-crypto/src/digest.rs
vincenthz: this is an escape hatch when type-level situation is not ideal, just like `unsafe` is useful sometimes. This is not about untyped comparaison, it's about making it quack like a duck in some, hopefully rare, situation.

chain-crypto/src/digest.rs
mzabaluev: OK, but then I don't see a point in this complication.

chain-crypto/src/digest.rs
NicolasDP: How does this look odd? You may compute the `DigestOf` an array of bytes (i.e. `Vec<u8>`) and now you want to lift the tag to `Fragment` as `Vec<u8>` was only the serialised version of `Fragment`.

This is useful to have an explicit conversion from one type to another as we grep `coerce` and see where the type tag has been changed to another. Allowing `From<Digest<T, H>> for Digest<T, U>` would have been a bad idea, at least the unsafety is explicit in the sense it is not a standard API feature that can be silently applied.

chain-crypto/src/digest.rs
vincenthz: yes, it's far from perfect to assume the presence of a slice of byte, this can be improved later; At the same time, it's really a type poorness to assume that any object can be hashed here, and not only sliced of data (think `MessageRaw`, `HeaderRaw`)

chain-crypto/src/digest.rs
mzabaluev: This sounds scary in conjunction with cryptographic hashes and the notion of a type that looks to a code reader like "this is _the_ cryptographic digest of a T".

mzabaluev: @NicolasDP the `coerce` might be hidden in a totally different part of the code.

chain-crypto/src/digest.rs
mzabaluev: There is no technical issue here that can't be solved with a `Hash`-like trait on `T`. Unless you really want different custom ways of hashing the same type.

NicolasDP:
There are limitations and improvements that can be done down the line. But overall this is going the right direction to improv our cryptographic tooling (typing here) in the higher level interfaces of the chain library

PR #51, opened 2019-07-11
clear_on_drop has a C dependency which blows up a wasm compilation

sjmackenzie:
make curve25519-dalek use the nightly feature which passes the
no_cc feature to clear_on_drop


mzabaluev:
The nightly feature will not work on the stable compiler.

Can we discard `clear_on_drop` instead? It's a large hack anyway.

vincenthz:
This is what you need:
https://github.com/input-output-hk/jormungandr/blob/master/testing/mjolnir/Cargo.toml#L42-L44

and no, we can't drop someone else dependency

sjmackenzie:
Ah didn't realise you could shadow transitive dependencies this way. neat.

PR #55, opened 2019-07-12
Add Send implementation and quickcheck arbitrary

vincenthz:


chain-crypto/src/digest.rs
```diff
@@ -62,7 +62,7 @@ impl fmt::Display for TryFromSliceError {
 
 pub trait DigestAlg {
     const HASH_SIZE: usize;
-    type DigestData: Clone + PartialEq + Hash + AsRef<[u8]>;
+    type DigestData: Clone + PartialEq + Hash + Send + Sync + AsRef<[u8]>;
```
NicolasDP: Remove `Sync`

```suggestion
    type DigestData: Clone + PartialEq + Hash + Send + AsRef<[u8]>;
```

PR #56, opened 2019-07-12
Transaction-ID removal

### LINKED ISSUE #25: transaction-id shortcoming related to accounts
the uniqueness property of utxo is hard to satisfy when the component are just accounts, since the nonce is associated with the witness only.

vincenthz:
remove transaction-id for transaction, and replace by:

* `TransactionSignDataHash` for the data that need to be signed for authenticating the transaction
* `FragmentId` which is used to store the transaction and identify it uniquely in the ledger

fix #25

### DIFF
--- a/chain-impl-mockchain/src/ledger/tests/discrimination_tests.rs
+++ b/chain-impl-mockchain/src/ledger/tests/discrimination_tests.rs
@@ -1,5 +1,6 @@
 #![cfg(test)]
 
+use crate::fragment::Fragment;
 use crate::testing::{
     address::AddressData,
     arbitrary::KindTypeWithoutMultisig,
@@ -9,6 +10,7 @@ use crate::testing::{
 use crate::transaction::*;
 use crate::value::*;
 use chain_addr::Discrimination;
+use chain_core::property::Message as _;
 use quickcheck::TestResult;
 use quickcheck_macros::quickcheck;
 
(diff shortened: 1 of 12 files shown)


chain-impl-mockchain/src/ledger/tests/discrimination_tests.rs
```diff
@@ -9,6 +10,7 @@ use crate::testing::{
 use crate::transaction::*;
 use crate::value::*;
 use chain_addr::Discrimination;
+use chain_core::property::Message as _;
```
mzabaluev: I'll start renaming the network things as well, in a separate PR.

chain-impl-mockchain/src/transaction/transaction.rs
```diff
@@ -1,12 +1,26 @@
...
 use crate::value::{Value, ValueError};
 use chain_addr::Address;
 use chain_core::mempack::{read_vec, ReadBuf, ReadError, Readable};
 use chain_core::property;
+use chain_crypto::{digest::DigestOf, Blake2b256};
+use std::boxed::Box;
 
-// FIXME: should this be a wrapper type?
-pub type TransactionId = Hash;
+pub struct TransactionSignData(Box<[u8]>);
```
mzabaluev: Good, a frozen `Vec`.

chain-impl-mockchain/src/transaction/transaction.rs
```diff
@@ -112,10 +126,10 @@ impl<Extra: property::Serialize> Transaction<Address, Extra> {
         self.serialize_body(&mut codec.into_inner())
     }
 
-    pub fn hash(&self) -> TransactionId {
+    pub fn hash(&self) -> TransactionSignDataHash {
         let mut bytes = Vec::new();
         self.serialize_body(&mut bytes).unwrap();
-        TransactionId::hash_bytes(&bytes)
+        DigestOf::digest(&TransactionSignData(bytes.into()))
```
mzabaluev: I think there should be a better looking API, like `Digest::of` to produce `DigestOf`.
Something to think of for a future refactoring.

chain-impl-mockchain/src/ledger/tests/discrimination_tests.rs
mzabaluev: See input-output-hk/jormungandr#523

chain-impl-mockchain/src/transaction/transaction.rs
vincenthz: ah that's a good name, will add it

PR #59, opened 2019-07-15
[Tests] Utxo verification for test: total_funds_are_total_in_ledger

dkijania:


chain-impl-mockchain/src/testing/arbitrary/transaction.rs
```diff
@@ -269,3 +270,58 @@ fn filter_accounts(x: &AddressDataValue) -> bool {
         _ => false,
     }
 }
+
+fn filter_utxo(x: &AddressDataValue) -> bool {
+    match x.address_data.kind() {
+        Kind::Single { .. } | Kind::Group { .. } => true,
```
CodeSandwich: Why filter them?

dkijania: Because i want to calculate expected utxos only. For account there is another verifier.

PR #64, opened 2019-07-18
Certificate overhaul

### LINKED ISSUE #18: overhaul certificate signing
general improvement to signing

* [x] cleanup signing capability
* [x] overhaul signing to be flexible to multi owners and different algorithms
* [x] transaction binding mechanism
* [ ] document the security guarantees

vincenthz:
* Introduce OwnerStakeDelegation which allows for a special case which is more compact
* ~Add a generic non-replayable / binded structure for certificate construction~ 
* Add special fragments for each certificate
* Add new fields in stake pool registration:
  * management_threshold: threshold of signature to meet for doing updates
  * start_time: for certificate validity


chain-crypto/src/sign.rs
```diff
@@ -120,8 +121,19 @@ impl<A: VerificationAlgorithm, T> Signature<T, A> {
             phantom: PhantomData,
         }
     }
+
+    pub fn safe_coerce<U: SafeSignatureCoerce<T>>(self) -> Signature<U, A> {
```
mzabaluev: Looks baroque, but safer than coercing `DigestOf` for arbitrary types.
Should we retire that `coerce` method as well?

vincenthz: `coerce` is still useful until we converted everything and added a safe version of all the possible conversion we need. At the moment the codebase is slowly converting to typed versions and keeping the raw representation alive, but it's far from being all converted, nor it's a good idea to convert everything in one go.

PR #65, opened 2019-07-23
[Tests] Address data enhancement

dkijania:


chain-impl-mockchain/src/testing/address.rs
```diff
@@ -37,25 +38,66 @@ impl PartialEq for AddressData {
     }
 }
 
+lazy_static! {
+    static ref DETERMINISTIC_ID: AtomicU64 = AtomicU64::new(1);
```
mzabaluev: This is a start on deterministic test runs, but now it's too deterministic? I would try to get the random seed from an environment variable, or if it is not set, use the seed of `OsRng`.

PR #66, opened 2019-07-23
network: Add Handshake method

mzabaluev:
The method to use on the client when connected initially, to get information on protocol version and genesis block. This will allow to bailing out of connections to wrong nodes.

### DIFF
--- a/network-grpc/src/lib.rs
+++ b/network-grpc/src/lib.rs
@@ -16,3 +16,5 @@ pub mod client;
 mod convert;
 pub mod server;
 mod service;
+
+pub const PROTOCOL_VERSION: u32 = 0;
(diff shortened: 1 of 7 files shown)


network-grpc/src/lib.rs
```diff
@@ -16,3 +16,5 @@ pub mod client;
 mod convert;
 pub mod server;
 mod service;
+
+pub const PROTOCOL_VERSION: u32 = 0;
```
NicolasDP: I would not use `0` as an initial version and keep `0` for error/garbage cases

mzabaluev: Now that you mention it, 0 is the no-field default in protobuf 3.
I will change it to 1, but we'll keep breaking the protocol without bumping the version until we declare it stable.

PR #73, opened 2019-07-31
[Tests] Enabled Property-test-api annotation for arbitraries

dkijania:
Moved arbitrary to testing module (in order to use it in jormungand e2e tests). Small refactoring for AddressData

### DIFF
--- a/chain-impl-mockchain/src/testing/address.rs
+++ b/chain-impl-mockchain/src/testing/address.rs
@@ -134,16 +134,19 @@ impl AddressData {
         self.address.discrimination().clone()
     }
 
-    pub fn address_as_string(&self) -> String {
+    pub fn delegation_key(&self) -> PublicKey<Ed25519> {
+        match self.kind() {
+            Kind::Group(_, delegation_key) => delegation_key,
+            _ => panic!("cannot get delegation key from non group addres kind"),
+        }
+    }
+
+    pub fn to_bech32_str(&self) -> String {
         let prefix = match self.discrimination() {
             Discrimination::Production => "ta",
             Discrimination::Test => "ca",
         };
         AddressReadable::from_address(prefix, &self.address).to_string()
-    }
-
-    pub fn public_key_as_string(&self) -> String {
-        self.public_key().to_bech32_str()
     }
 
     pub fn generate_key_pair<A: AsymmetricKey>() -> KeyPair<A> {
--- a/chain-impl-mockchain/src/fee.rs
+++ b/chain-impl-mockchain/src/fee.rs
@@ -55,7 +55,7 @@ impl FeeAlgorithm<tx::Transaction<Address, Certificate>> for LinearFee {
     }
 }
 
-#[cfg(test)]
+#[cfg(any(test, feature = "property-test-api"))]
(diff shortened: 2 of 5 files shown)


chain-impl-mockchain/src/testing/address.rs
```diff
@@ -122,6 +120,13 @@ impl AddressData {
         }
     }
 
+    pub fn delegation_key(&self) -> PublicKey<Ed25519> {
+        match self.kind() {
+            Kind::Group(_, delegation_key) => delegation_key,
+            _ => panic!("cannto get delegation key from non group addres kind"),
```
CodeSandwich: `cannto` is Japanese name of famous philosopher :P

chain-impl-mockchain/src/testing/arbitrary/mod.rs
```diff
@@ -64,3 +63,25 @@ impl Arbitrary for Output<Address> {
...
+
+impl Arbitrary for LinearFee {
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        Self {
+            constant: Arbitrary::arbitrary(g),
+            coefficient: Arbitrary::arbitrary(g),
+            certificate: Arbitrary::arbitrary(g),
+        }
+    }
+}
```
CodeSandwich: I'm not really convinced that moving impls of `Arbitrary` to a central point is a good idea.

1. This file is going to become a behemoth and a messy jungle
2. We're decoupling type definition and instance generation. It's going to be hard to find what type `X`'s `Arbitrary` looks like, because this isn't going to be defined where `X` is.
1. Some `Arbitrary` impls are probably not going to be movable at all, because they can depend on access to private fields.

dkijania: 1) it can be split accross testing module. the only requirement was that it needs to be in testing module
2) Yes i have similar doubts. but i couldn't find any other solution for enabling jormungadr integration tests use testing structures in `chain-impl-mockchain`. I copied "solution" for chain-crypto. The thing is, in order to use code from `chain-impl-mockchain` in test cfg in `jormungandr-integration-tests`  i need to import it with :

chain-impl-mockchain = { path = "../chain-addr", features=["property-test-api"]}

and this requires all arbitrary to be visible in testing mod. i couldn;t find any other idea how to fix it.

3) Unfortunately yes, i fixed that by `pub(crate)` but dunno if this is a great idea

CodeSandwich: The 3. alone is a red flag for me. We can't dehermetize things to allow foreign hacky constructors. At most we can provide normal constructors to build hacky ones on top of them.

The trait impl doesn't have to be visible or imported, its existence is enough for it to be used. I think that making them exist by passing a feature isn't a bad idea. The `impl Arbitrary for ...` will have to be annotated (or put into an annotated module) with something like `#[cfg(any(test, feature = "test_utils_please"))]`

dkijania: Oh, great. your solution is working as expected. pushed changes with annotations

PR #74, opened 2019-08-02
[Tests] Property test for account ledger

dkijania:
Property test for  removing account / removing value operations since they present some logic.

1) First test case adds account and then tries to remove it (which is only successful if account value = 0)
2) Second test case adds account and then tries to sub value (which is only successful if value subtracted from account is less than actual balance)

### DIFF
--- a/chain-impl-mockchain/src/accounting/account/mod.rs
+++ b/chain-impl-mockchain/src/accounting/account/mod.rs
@@ -212,23 +212,24 @@ pub mod tests {
         let result = ledger.remove_account(&id);
         let expected_result = account_state.get_value() == Value::zero();
         match (result, expected_result) {
-            (Err(_), false) => verify_account_exists(&ledger, &id, true),
+            (Err(_), false) => verify_account_exists(&ledger, &id),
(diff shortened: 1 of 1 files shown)


chain-impl-mockchain/src/accounting/account/mod.rs
```diff
@@ -156,3 +156,79 @@ impl<ID: Clone + Eq + Hash, Extra: Clone> std::iter::FromIterator<(ID, AccountSt
...
+        let expected_result = account_state.get_value() == Value::zero();
+        match (result, expected_result) {
+            (Err(_), false) => verify_account_exists(&ledger, &id, true),
+            (Ok(_), false) => TestResult::failed(),
+            (Err(_), true) => TestResult::failed(),
+            (Ok(ledger), true) => verify_account_exists(&ledger, &id, false),
+        }
+    }
+
+    fn verify_account_exists(ledger: &Ledger, id: &Identifier, should_exist: bool) -> TestResult {
```
CodeSandwich: It's probably better for readability to split this function into `verify_account_exists` and `verify_account_does_not_exists`

PR #76, opened 2019-08-06
[Tests] Block builder property test

dkijania:
Property tests for block builder :
- genesis block
- bft block
- genesis praos block

### DIFF
--- a/chain-impl-mockchain/src/block/mod.rs
+++ b/chain-impl-mockchain/src/block/mod.rs
@@ -226,10 +226,6 @@ mod test {
         fn block_serialization_bijection(b: Block) -> TestResult {
             property::testing::serialization_bijection(b)
         }
-
-        fn block_is_consient(block: Block) -> TestResult {
-            TestResult::from_bool(block.is_consistent())
-        }
     }
 
     impl Arbitrary for HeaderRaw {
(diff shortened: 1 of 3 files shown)


chain-impl-mockchain/src/block/mod.rs
```diff
@@ -222,6 +222,10 @@ mod test {
         fn block_serialization_bijection(b: Block) -> TestResult {
             property::testing::serialization_bijection(b)
         }
+
+        fn block_is_consient(block: Block) -> TestResult {
```
vincenthz: typo

vincenthz: I couldn't find a use for this function, is there something missing maybe ?

dkijania: Removing

chain-impl-mockchain/src/testing/arbitrary/utils.rs
```diff
@@ -71,3 +72,39 @@ pub fn choose_random_indexes<G: Gen>(gen: &mut G, upper_bound: usize) -> HashSet
...
+    pub fn verify(&mut self, is_true: bool, desc: &str) {
+        let result = match is_true {
+            false => TestResult::error(format!("expected {} to be true, but got false", desc)),
+            true => TestResult::passed(),
+        };
+        self.0.push(result);
+    }
+
+    pub fn get_result(&self) -> TestResult {
+        match self.0.iter().cloned().find(|x| x.is_failure()) {
```
CodeSandwich: Nitpick: `self.0.iter.find(TestResult::is_failure).cloned().unwrap_or_else(TestResult::passed)`?

PR #83, opened 2019-08-20
[Tests] Property test for Transaction with Certificate builder

dkijania:


chain-impl-mockchain/src/txbuilder.rs
```diff
@@ -472,4 +473,31 @@ mod tests {
...
         }
     }
+
+    #[quickcheck]
+    pub fn cert_builder_never_creates_unbalanced_tx(
+        input: Input,
+        fee: LinearFee,
+        certificate: Certificate,
+    ) -> TestResult {
+        let mut builder = TransactionBuilder::new();
```
vincenthz: you have to replace `::new` by `::payload(certificate)` IIRC.

PR #84, opened 2019-08-27
move ledger iterator into its own module

vincenthz:


vincenthz:
spurious failure of statistical test on appveyor

PR #89, opened 2019-09-03
FragmentService API for network protocol client

mzabaluev:
Add the missing definitions for client-side network protocol to support fragment methods.


CodeSandwich:
2 minutes late, but LGTM

PR #91, opened 2019-09-06
Ledger apply TX clean up

CodeSandwich:
Part of https://github.com/input-output-hk/jormungandr/pull/748:
- Exposed Transaction validation tools
- Splits and cleans up ledger internal_apply_transaction

### DIFF
--- a/chain-impl-mockchain/src/ledger/check.rs
+++ b/chain-impl-mockchain/src/ledger/check.rs
@@ -113,3 +113,59 @@ pub(super) fn valid_pool_update_certificate(
     )?;
     Ok(())
 }
+
+custom_error! {
+    #[derive(Clone, PartialEq, Eq)]
+    pub TxVerifyError
+        TooManyInputs {expected: usize, actual: usize }
+            = "too many inputs, expected maximum of {expected}, but received {actual}",
+        TooManyOutputs {expected: usize, actual: usize }
+            = "too many outputs, expected maximum of {expected}, but received {actual}",
+        TooManyWitnesses {expected: usize, actual: usize }
+            = "too many witnesses, expected maximum of {expected}, but received {actual}",
+        NumberOfSignaturesInvalid { expected: usize, actual: usize }
+            = "invalid number of signatures, expected {expected}, but received {actual}",
+}
+
+pub struct TxVerifyLimits {
+    pub max_inputs_count: usize,
+    pub max_outputs_count: usize,
+    pub max_witnesses_count: usize,
+}
+
+impl<OutAddress, Extra> AuthenticatedTransaction<OutAddress, Extra> {
+    pub fn verify_well_formed(&self, limits: &TxVerifyLimits) -> Result<(), TxVerifyError> {
+        let inputs = &self.transaction.inputs;
+        if inputs.len() > limits.max_inputs_count {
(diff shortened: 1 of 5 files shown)


chain-impl-mockchain/src/transaction/mod.rs
```diff
@@ -21,6 +21,62 @@ pub struct AuthenticatedTransaction<OutAddress, Extra> {
...
+}
+
+impl<OutAddress, Extra> AuthenticatedTransaction<OutAddress, Extra> {
+    pub fn verify_well_formed(&self, limits: &VerificationLimits) -> Result<(), VerifyError> {
+        let inputs = &self.transaction.inputs;
+        if inputs.len() > limits.max_inputs_count {
+            Err(VerifyError::TooManyInputs {
+                expected: limits.max_inputs_count,
+                actual: inputs.len(),
+            })?
```
NicolasDP: please, don't do that. use `return` so it is obvious you are leaving the function at this opint

chain-impl-mockchain/src/transaction/mod.rs
```diff
@@ -21,6 +21,62 @@ pub struct AuthenticatedTransaction<OutAddress, Extra> {
...
+        if inputs.len() != witnesses.len() {
+            Err(VerifyError::NumberOfSignaturesInvalid {
+                expected: inputs.len(),
+                actual: witnesses.len(),
+            })?
+        }
+
+        Ok(())
+    }
+}
```
NicolasDP: all of this should be in `ledger/check.rs.` Vincent has put a macro to help with the check making the test more consise (can be extended obviously).

chain-impl-mockchain/src/transaction/mod.rs
```diff
@@ -21,6 +21,62 @@ pub struct AuthenticatedTransaction<OutAddress, Extra> {
...
+            = "too many witnesses, expected maximum of {expected}, but received {actual}",
+        NumberOfSignaturesInvalid { expected: usize, actual: usize }
+            = "invalid number of signatures, expected {expected}, but received {actual}",
+}
+
+pub struct VerificationLimits {
+    pub max_inputs_count: usize,
+    pub max_outputs_count: usize,
+    pub max_witnesses_count: usize,
+}
```
NicolasDP: This is a good idea because we might want to make some restrictions configurable at the protocol level. Don't do that know, but that's a good starting point.

chain-impl-mockchain/src/transaction/mod.rs
CodeSandwich: That's the first place I've tried, but it felt wrong:
1. `check.rs` is private. It looks like a bag of internal helpers and probably shouldn't be made public.
2. This is strictly `AuthenticatedTransaction` logic. It should be as close to its definition as possible, making it method is also convenient and easy to find for users.

NicolasDP: 1. make it public.
2. we are the user. Making it in one module is convenient too.

CodeSandwich: I've made `check.rs` public. `verify_well_formed` is moved to `check.rs` completely. Because it's in the same crate, it can still be a method of `AuthenticatedTransaction`.

PR #94, opened 2019-09-13
Delayed resolution for upload sinks

mzabaluev:
In network-core, change the methods producing upload sinks
to return a future that resolves to the sink, instead of the sink
directly. Despite more complexity, this simplifies application code
in case that needs asynchronous setup for inbound processing.

### DIFF
--- a/network-grpc/src/service.rs
+++ b/network-grpc/src/service.rs
@@ -18,7 +18,6 @@ use network_core::{
 
 use futures::future::{self, FutureResult};
 use futures::prelude::*;
-use futures::stream::{Forward, Fuse};
 use futures::try_ready;
 use tower_grpc::{self, Code, Request, Response, Status, Streaming};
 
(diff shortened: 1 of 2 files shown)


network-grpc/src/service.rs
```diff
@@ -259,43 +258,80 @@ where
...
+    pub enum State<St, F>
+    where
+        St: Stream<Error = tower_grpc::Status>,
+        F: Future,
+        F::Item: Sink,
+        <F::Item as Sink>::SinkItem: FromProtobuf<St::Item>,
+    {
+        WaitingSink(F, St),
+        Forwarding(Forward<Fuse<RequestStream<<F::Item as Sink>::SinkItem, St>>, F::Item>),
+        Intermediate,
```
CodeSandwich: What does `Intermediate` state mean? Doesn't `WaitingSink` future take care of all states while building the sink?

network-grpc/src/service.rs
```diff
@@ -259,43 +258,80 @@ where
...
+                let stream = RequestStream::new(stream).fuse();
+                self.state = Forwarding(stream.forward(sink));
+            } else {
+                unreachable!()
+            }
+        }
+    }
+}
+
+mod stream_forward {
```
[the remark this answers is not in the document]
mzabaluev: It might, but I couldn't think of a good module to put it in. The `mod` is mostly to avoid an unwieldy long name for `State`, it has no other purpose for now.

network-grpc/src/service.rs
mzabaluev: It's needed to be able to do `mem::replace` on the borrowed future state that is passed to `poll`, in order to move the stream object out for further processing.

PR #100, opened 2019-10-01
Apply empty block prop test

dkijania:
Property test which check if incoming block has correct chain length and block date

### DIFF
--- a/chain-impl-mockchain/src/ledger/ledger.rs
+++ b/chain-impl-mockchain/src/ledger/ledger.rs
@@ -969,7 +969,7 @@ mod tests {
     use chain_addr::Discrimination;
     use quickcheck::{Arbitrary, Gen, TestResult};
     use quickcheck_macros::quickcheck;
-    use std::iter;
+    use std::{fmt, iter};
 
     impl Arbitrary for LedgerStaticParameters {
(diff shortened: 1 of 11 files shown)


chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -991,6 +991,60 @@ mod tests {
...
+    }
+
+    #[quickcheck]
+    pub fn apply_empty_block_prop_test(
+        context: HeaderContentEvalContext,
+        ledger: ArbitraryEmptyLedger,
+    ) -> TestResult {
+        let ledger: Ledger = ledger.into();
+        let should_succeed =
+            context.chain_length == ledger.chain_length.next() && context.block_date > ledger.date;
```
NicolasDP: is this ever true in the property test?

dkijania: i don't fully understand. if we apply empty block then both conditions are the only conditions determine success. On the other hand should_succedd is only try if arbitrary ledger and arbitrary context have this combination on chain_length and block_data, or am I missing something?

vincenthz: that looks fine, although I had to read this thrice

PR #113, opened 2019-10-15
breaking change: Header refactor

vincenthz:
* use mostly like a c structure packed in memory
* improve memory/speed handling and the amount of deserialization and serialization needed to cut, slice and dice
* verify VRF more thoroughly
* improve authenticated data to include more than the common parts
* retire block builder
* add a header builder that have a specific state machine: prevent user issue better and introduce rigour in creation


chain-impl-mockchain/src/fragment/content.rs
```diff
@@ -0,0 +1,82 @@
...
+        self.compute_hash_size() == rhs.compute_hash_size()
+    }
+}
+impl Eq for Contents {}
+
+impl From<ContentsBuilder> for Contents {
+    fn from(content_builder: ContentsBuilder) -> Self {
+        Contents(content_builder.fragments.into())
+    }
+}
```
NicolasDP: consensus is to have `ContentsBuilder::build()` function instead

chain-impl-mockchain/src/key.rs
```diff
@@ -234,6 +238,12 @@ impl From<Hash> for [u8; 32] {
     }
 }
 
+impl<'a> From<&'a Hash> for &'a [u8; 32] {
+    fn from(h: &'a Hash) -> Self {
+        (&h.0).into()
+    }
+}
+
```
NicolasDP: You mean to implement `AsRef<[u8;32]> for Hash` ?

chain-impl-mockchain/src/leadership/bft.rs
```diff
@@ -14,6 +14,12 @@ pub type BftVerificationAlg = Ed25519;
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct LeaderId(pub(crate) PublicKey<BftVerificationAlg>);
 
+impl From<[u8; 32]> for LeaderId {
+    fn from(v: [u8; 32]) -> LeaderId {
+        LeaderId(PublicKey::from_binary(&v[..]).expect("leader-id invalid format"))
+    }
+}
```
NicolasDP: what about `TryFrom` instead of `From` to avoid a partial function here?

chain-impl-mockchain/src/header/cstruct.rs
```diff
@@ -0,0 +1,419 @@
...
+        unsafe { self.gp[HEADER_OFFSET_GP_KES_SIG..HEADER_GP_SIZE].copy_from_slice(s) }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum HeaderError {
+    InvalidSize,
+    UnknownVersion,
+    SizeMismatch { expected: usize, got: usize },
+}
```
NicolasDP: missing `Display` and `Error` implementation. This type is not going to propagate well to the user of the APIs, creating.

chain-impl-mockchain/src/fragment/content.rs
vincenthz: will keep that in mind, definitely better

chain-impl-mockchain/src/leadership/bft.rs
vincenthz: the conversion cannot fail. this is what I've been saying that I need new interfaces in chain-crypto to do these things

PR #119, opened 2019-10-21
Expose key derivation for WASM

SebastienGllmt:
See https://github.com/input-output-hk/js-chain-libs/pull/58 for usage

Currently the Shelley WASM bindings don't allow for child key derivations because the capability isn't exposed by chain-libs. This PR fixes this. At the same time I also add the ability to create a key from a bip39 seed

##### 1) Why create a `derive.rs` file

Currently `XPrv` and `XPub` are private so you can't access them from the WASM bindings. Instead of changing this, I add a new file called `derive.rs` that forwards the derivation to the `XPrv`. I put it in a new file because

1) It doesn't belong in `Ed25519_derive.rs` since it needs to access `SecretKey<A>`
2) It doesn't belong in `key.rs` because it needs to access an algorithm-specific function

It seems the current pattern for when (1) and (2) hold is to put them in separate files (such as `sign.rs` and `asymlock.rs`

##### 2) Not exposing DerivationScheme

Since we don't support support the v1 derivation scheme, I manually hardcode using v2 in `derive.rs`. We could decide to still take `DerivationScheme` as an argument to the function in case we want to define a v3 later. I'm open to either for this.

### DIFF
--- a/chain-crypto/src/key.rs
+++ b/chain-crypto/src/key.rs
@@ -150,6 +150,12 @@ impl<A: AsymmetricPublicKey> AsRef<[u8]> for PublicKey<A> {
     }
 }
 
+impl<A: AsymmetricKey> AsRef<[u8]> for SecretKey<A> {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
+
 impl<A: AsymmetricKey> From<SecretKey<A>> for KeyPair<A> {
     fn from(secret_key: SecretKey<A>) -> Self {
         let public_key = secret_key.to_public();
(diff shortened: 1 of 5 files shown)


chain-crypto/src/key.rs
```diff
@@ -150,6 +150,12 @@ impl<A: AsymmetricPublicKey> AsRef<[u8]> for PublicKey<A> {
     }
 }
 
+impl<A: AsymmetricKey> AsRef<[u8]> for SecretKey<A> {
```
SebastienGllmt: This is unrelated to the rest of this PR but it's just a functionality I wanted that way you can implement `as_bytes` for private keys using

```rust
pub fn as_bytes(&self) -> Vec<u8> {
        self.0.as_ref().to_vec()
    }
```

vincenthz:
While I'm fine exposing all that need to be here, all the wallet related stuff have moved to https://github.com/input-output-hk/chain-wallet-libs/blob/master/hdwallet/src/bip44.rs

PR #120, opened 2019-10-22
Fix CBOR (de)serialization for legacy ExtendedAddr

CodeSandwich:
Fixes https://github.com/input-output-hk/jormungandr/issues/968

### DIFF
--- a/cardano-legacy-address/src/address.rs
+++ b/cardano-legacy-address/src/address.rs
@@ -18,7 +18,7 @@ use cryptoxide::sha3::Sha3;
 use ed25519_bip32::XPub;
 
 use std::{
-    convert::TryFrom,
+    convert::{TryFrom, TryInto},
     fmt,
     io::{BufRead, Write},
 };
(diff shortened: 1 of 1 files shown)


cardano-legacy-address/src/address.rs
```diff
@@ -314,7 +316,14 @@ impl cbor_event::de::Deserialize for ExtendedAddr {
         let bytes = cbor::util::raw_with_crc32(reader)?;
         let mut raw = Deserializer::from(std::io::Cursor::new(bytes));
         raw.tuple(3, "ExtendedAddr")?;
-        let addr = cbor_event::de::Deserialize::deserialize(&mut raw)?;
+        let addr_bytes = raw.bytes()?;
+        let addr = addr_bytes.as_slice().try_into().map_err(|_| {
+            cbor_event::Error::WrongLen(
+                addr_bytes.len() as u64,
+                cbor_event::Len::Len(EXTENDED_ADDR_LEN as u64),
+                "invalid extended addr length",
```
mzabaluev: nitpick: It's better to not abbreviate in user readable messages, so: "invalid length of extended address".

PR #122, opened 2019-10-24
[Tests] Scenario chain libs

dkijania:
Mini framework for e2e ledger tests.
We can build scenario just like jormungandr-scenario-tests library.Supported functionalities:

create ledger with initial wallets, stake pool delegation, stake pool ownership,
create transaction
create certificates (Register,Delegate,Retire)
various verifications (e.q. Stake distribution check)


dkijania:
Rebased and fixed after breaking changes

PR #124, opened 2019-10-25
Implement old UTXO witness creation and verification

CodeSandwich:
Part of https://github.com/input-output-hk/jormungandr/issues/1005


NicolasDP:
remove the the verification as this will never work and is not as simple as that. Vincent will work on the verification side of this as it affects the whole ledger and this path is never called anyway.

CodeSandwich:
Ok, verification removed

PR #128, opened 2019-10-28
Transaction rework

### LINKED ISSUE #18: overhaul certificate signing

vincenthz:
* Keep the transaction in binary format and slice it efficiently on demand
* Get rid of AuthenticatedTransaction (now just Transaction)
* Add InputOutputs builder which is not parametrised to the type anymore, but expect it explicitly when calculating fees.
* Redo the transaction builder to be a state machine: 1) set payload 2) set input-outputs 3) set witnesses 4) set payload-auth
* Bind certificate with transaction
* simplify tests and add more efficient kitchen sink construct (e.g. TestLedger) to allow less repetition and more state tracking

fix #18


chain-impl-mockchain/src/transaction/io.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// Seal the transaction by passing fee rule and the output policy
+    ///
+    /// Along with the transaction, this return the balance unassigned to output policy
+    /// if any
+    pub fn seal_with_output_policy<P: Payload, F: FeeAlgorithm<P>>(
+        mut self,
+        payload: &P,
+        fee_algorithm: F,
+        policy: OutputPolicy,
+    ) -> Result<(Balance, Vec<Output<Address>>, InputOutput), Error> {
```
SebastienGllmt: 👍

PR #130, opened 2019-10-29
Add spending counter in witness

SebastienGllmt:
The Jormungandr docs say that the spending counter should be part ofthe account witness. However, it turns out it is not present and so this PR adds it.

Before: 
```
0277b757da13eed8c5c934541782d389b1e955577a071f0a2c400bdecc538b34cd9632f38416e7b1022b57f6ac72dabebd89f6d946354d2ea0a349c7e0e5e8b705

02df368fa3a6d699d67e16f3d1da986574ea3eda06262cc0b5477ab16608cf3d0df2b09cad1030b674761c34325418c188972ea51c71c603c99f67568a93e8e604
```
After: 
```
0277b757da13eed8c5c934541782d389b1e955577a071f0a2c400bdecc538b34cd9632f38416e7b1022b57f6ac72dabebd89f6d946354d2ea0a349c7e0e5e8b70500000000

02df368fa3a6d699d67e16f3d1da986574ea3eda06262cc0b5477ab16608cf3d0df2b09cad1030b674761c34325418c188972ea51c71c603c99f67568a93e8e60400000001
```

### DIFF
--- a/chain-impl-mockchain/doc/format.md
+++ b/chain-impl-mockchain/doc/format.md
@@ -149,8 +149,8 @@ With the following serialization:
   * Type=2 utxo witness scheme (64 bytes):
     * ED25519 Signature (64 bytes)
   * Type=3 Account witness (68 bytes):
-    * Account Counter (4 bytes : TODO-ENDIANNESS)
     * ED25519 Signature (64 bytes)
+    * Account Counter (4 bytes in big endian)
 
 The message, w.r.t the cryptographic signature, is generally of the form:
 
--- a/chain-impl-mockchain/src/ledger/ledger.rs
+++ b/chain-impl-mockchain/src/ledger/ledger.rs
@@ -776,7 +776,7 @@ impl Ledger {
         witness: &Witness,
     ) -> Result<Self, Error> {
         match witness {
-            Witness::Account(_) => Err(Error::ExpectingUtxoWitness),
+            Witness::Account(_, _) => Err(Error::ExpectingUtxoWitness),
             Witness::Multisig(_) => Err(Error::ExpectingUtxoWitness),
             Witness::OldUtxo(xpub, signature) => {
                 let (old_utxos, associated_output) = self
@@ -885,7 +885,8 @@ fn match_identifier_witness<'a>(
     match witness {
         Witness::OldUtxo(_, _) => Err(Error::ExpectingAccountWitness),
         Witness::Utxo(_) => Err(Error::ExpectingAccountWitness),
-        Witness::Account(sig) => {
+        Witness::Account(sig, counter) => {
+            // TODO: use counter in verification?
             // refine account to a single account identifier
             let account = account
                 .to_single_account()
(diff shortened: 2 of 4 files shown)


chain-impl-mockchain/doc/format.md
```diff
@@ -149,8 +149,8 @@ With the following serialization:
   * Type=2 utxo witness scheme (64 bytes):
     * ED25519 Signature (64 bytes)
   * Type=3 Account witness (68 bytes):
-    * Account Counter (4 bytes : TODO-ENDIANNESS)
     * ED25519 Signature (64 bytes)
+    * Account Counter (4 bytes in big endian)
```
SebastienGllmt: hopefully nobody minds that I moved this around (no particular reason to swap the order. I just happen to code it this way)

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -885,7 +885,8 @@ fn match_identifier_witness<'a>(
     match witness {
         Witness::OldUtxo(_, _) => Err(Error::ExpectingAccountWitness),
         Witness::Utxo(_) => Err(Error::ExpectingAccountWitness),
-        Witness::Account(sig) => {
+        Witness::Account(sig, counter) => {
+            // TODO: use counter in verification?
```
SebastienGllmt: TODO

chain-impl-mockchain/src/transaction/witness.rs
```diff
@@ -78,7 +78,7 @@ impl WitnessAccountData {
         transaction_id: &TransactionSignDataHash,
         spending_counter: &account::SpendingCounter,
     ) -> Self {
-        let mut v = Vec::with_capacity(65);
+        let mut v = Vec::with_capacity(69);
```
SebastienGllmt: I have no idea if this should be `65` or `69`. `tag || block0` || `transaction_id` is  65 bytes so if you add `spending_counter` then it should be 69 so I thought maybe it should be increased

SebastienGllmt: Similarly, I noticed we do `let mut v = Vec::with_capacity(65);` for `WitnessUtxoData` but this seems wrong since we're only writing 64 bytes (for some reason the tag byte isn't added)

vincenthz:
The doc need some thorough update, as I can see why this could be very confusing.

* The counter doesn't appear in the serialized witness in the transaction (we don't need it) but we use it implicitly to generate and verify that the sequentiality of the operation is respected.
* `WitnessUtxoData` should have the TAG pushed just like all other witness, for consistency reason. account have been added after the fact, which is why I forgot about the TAG. (for posterity, everything other category of signature live in a different "plane of data to verify", so it's not a safety problem)
* `with_capacity` seems wrong in place, but the only thing that it does is prevent a memory re-allocation as it might now; this need correction just for memory efficiency.
* currently the spending counter is serialized in little endian; I suspected that it might be when I updated the format, and hence the TODO item. might be good to unify this with the rest of the serialization choices.

PR #134, opened 2019-11-03
Add CertificatePayload enum

CodeSandwich:
This allows infallible conversion from `Certificate` to `CertificateSlice`, which is needed for fee calculation in JCLI TX info

### DIFF
--- a/chain-impl-mockchain/src/certificate/mod.rs
+++ b/chain-impl-mockchain/src/certificate/mod.rs
@@ -4,7 +4,7 @@ mod pool;
 #[cfg(test)]
 mod test;
 
-use crate::transaction::{Payload, PayloadSlice};
+use crate::transaction::{Payload, PayloadData, PayloadSlice};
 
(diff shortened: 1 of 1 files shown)


chain-impl-mockchain/src/certificate/mod.rs
```diff
@@ -64,6 +64,49 @@ impl<'a> CertificateSlice<'a> {
...
+                CertificatePayload::PoolRetirement(payload.payload_data())
+            }
+            Certificate::PoolUpdate(payload) => {
+                CertificatePayload::PoolUpdate(payload.payload_data())
+            }
+        }
+    }
+}
+
+impl<'a> From<&'a CertificatePayload> for CertificateSlice<'a> {
```
vincenthz: I don't think it's a good idea to call "owned-data-to-borrow-slice" through `From`. I'ld much rather have a specific method (unless there's one generic trait that allow this, like `Borrow` which I don't think exist here due to <'a>)

PR #150, opened 2019-11-12
(#52) Optimize unspent utxo

### LINKED ISSUE #52: Optimise Utxo representation
At the moment the unspent is implemented using a `BtreeMap`:

    struct TransactionUnspents<OutAddress>(BTreeMap<TransactionIndex, Output<OutAddress>>);

whereas it could do with an optimised contiguous implementation with a bitmap and special cases (likely 1 index, 2 indices, 3 indices)

eugene-babichenko:
- Implement `SparseArray`.
- Implement wrapper (`FastSparse`) on top of sparse array that allows to share data between its instances and thus allows for faster clones.

Data under `FastSparseArray` is shared until `set` or `shrink` function is called. It is also possible to avoid copying the underlying structure when calling `set`, but actual use cases do not require that.

**TBD**

- [x] Faster popcount (utilize `POPCNT` instruction)
- [x] Faster trailing zeros count (`TZCNT`)
- [ ] Special bitmaps for small arrays (8, 16, ... elements) #157 

Resolves #52

### DIFF
--- a/sparse-array/src/fast.rs
+++ b/sparse-array/src/fast.rs
@@ -1,0 +1,187 @@
+///! Wrapper for sparse arrays that doesn't delete anything from the memory
+/// unless `shrink` is called.
+
+use std::sync::Arc;
+use crate::{
+    bitmap::BitmapIndex,
+    SparseArray,
+};
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct FastSparseArray<V> {
+    index: BitmapIndex,
+    data: Arc<SparseArray<V>>,
+}
+
+impl<V: Clone> FastSparseArray<V> {
+    pub fn new() -> Self {
+        Self {
+            index: BitmapIndex::new(),
+            data: Arc::new(SparseArray::new()),
+        }
+    }
+
+    pub fn with_capacity(capacity: u8) -> Self {
+        Self {
+            index: BitmapIndex::new(),
+            data: Arc::new(SparseArray::with_capacity(capacity)),
+        }
+    }
+
+    pub fn get(&self, idx: u8) -> Option<&V> {
+        if !self.index.get_index(idx) {
+            None
+        } else {
+            self.data.get(idx)
+        }
+    }
+
+    pub fn set(&mut self, idx: u8, value: V) {
+        match Arc::get_mut(&mut self.data) {
+            Some(d) => d.set(idx, value),
+            None => {
+                let mut new_sparse_array = SparseArray::new();
+                for (idx, value) in self.iter() {
+                    new_sparse_array.set(idx, (*value).clone());
+                }
+                new_sparse_array.set(idx, value);
+                self.data = Arc::new(new_sparse_array);
+            }
+        }
+        self.index.set_index(idx);
+    }
+
+    pub fn remove(&mut self, idx: u8) -> Option<V> {
+        if self.index.get_index(idx) {
+            self.index.remove_index(idx);
+            return self.data.get(idx).map(|x| (*x).clone());
+        }
+
+        None
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.index.is_empty()
+    }
+
+    pub fn contains_key(&self, idx: u8) -> bool {
+        self.index.get_index(idx)
+    }
+
+    pub fn iter(&self) -> FastSparseArrayIter<V> {
+        FastSparseArrayIter::new(&self)
+    }
+
+    pub fn shrink(&mut self) {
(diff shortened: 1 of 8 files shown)


sparse-array/src/fast.rs
```diff
@@ -0,0 +1,187 @@
...
+use std::sync::Arc;
+use crate::{
+    bitmap::BitmapIndex,
+    SparseArray,
+};
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct FastSparseArray<V> {
+    index: BitmapIndex,
+    data: Arc<SparseArray<V>>,
```
NicolasDP: I don't get the reason for the presence of the `Arc`?

eugene-babichenko: Initially I wanted to use `Rc` to share this field between instances of `FastSparseArray`, but apparently it is used by different threads in the UTXO module.

The idea behind sharing this data is making faster clones, I discussed this one with Vincent.

NicolasDP: ok got it

sparse-array/src/fast.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+impl<V: Clone> FastSparseArray<V> {
```
vincenthz: I don't think `Clone on V` should be needed for every method. I can only imagine `remove` and `shrink` need it since it's immutable, but that's about it.

sparse-array/src/fast.rs
```diff
@@ -0,0 +1,187 @@
...
+impl<'a, V> FastSparseArrayIter<'a, V> {
+    pub fn new(sparse_array: &'a FastSparseArray<V>) -> Self {
+        Self {
+            bitmap: sparse_array.index.clone(),
+            sparse_array,
+        }
+    }
+}
+
+impl<'a, V: Clone> Iterator for FastSparseArrayIter<'a, V> {
```
vincenthz: only needed it because `get` requires it too, but same thing it shouldn't be needed

sparse-array/src/sparse_array.rs
```diff
@@ -0,0 +1,133 @@
+use crate::bitmap::BitmapIndex;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SparseArray<V> {
+    index: BitmapIndex,
+    data: Vec<V>,
```
vincenthz: this can be reduced to a `Box<[V]>` to save 8 bytes

eugene-babichenko:
@vincenthz I used `where` clauses for functions that really require the `Clone` trait, it allowed to removed the trait bound from the majority of structs. The only structure in utxo that still require `Clone` is `Ledger`, but each method here require `Clone`, so I think this is fine.

sparse-array/src/sparse_array.rs
eugene-babichenko: I'm not sure about that. Will we still be able to add and remove elements at arbitrary positions?

vincenthz: no, the `Box<[V]>` is for when the array is immutable/frozen. when in the tree, it's useful to reduce the cost of the array by 8 bytes by leaf. But once you want to update it, you can thaw it into a `Vec` with `into_vec`

eugene-babichenko: This will change the interface a bit. Instead of `set(&mut self, idx: u8, value: V)` and `remove(&mut self, idx: u8) -> Option<V>` we will have `set(self, idx: u8, value: V) -> Self` and `remove(self, idx: u8) -> (Self, Option<V>)`.

Anyway, I think that's ok for an immutable struct.

vincenthz: yes, I think that's fine. if you want to have a mutable interface, I usually like the pattern of having a `Builder` type which use data structure that are more friendly to modification like Vec in this case.

eugene-babichenko:
@vincenthz I suggest moving this

> Special bitmaps for small arrays (8, 16, ... elements)

into a separate issue and then create a reusable bitmap that will be used by both `sparse-array` and `imhamt`.

vincenthz:
yes, we can move the special cases to another low priority issue.

Otherwise everything has been addressed and now ready for merge, but I'm less than conviced about the latest commit to move from `u8` to `usize`. it's safer to keep as the lowest possible "size" since it's always safe to cast to bigger size, whereas the opposite direction `usize -> u8` is not.

eugene-babichenko:
dropped this one

vincenthz:
nice one !

Issue #153, opened 2019-11-13
Replace Vec<T> by Box<[T]>


vincenthz:
In many immutable places, we keep `Vec<T>` instead of using `Box<[T]>` which save 8 bytes and also represent immutability better than Vec.

Issue #152, opened 2019-11-13
Optimise HAMT internal


vincenthz:
* mutable mode should kept the tree mutable and modify the tree on demand (Cow)
* freezing: should replace all mutated cells, back to frozen nodes
* replace recursive functions by explicit iterative functions
* optimise node internals (unsafe / pointers / ..)

PR #154, opened 2019-11-13
Optimise HAMT inner representation to use Box instead of Vec

vincenthz:
related to #153 #152


imhamt/src/content.rs
```diff
@@ -28,7 +28,7 @@ impl<K: Clone, V> KV<K, V> {
 pub enum SmallVec<T> {
     One(T),
     //Two(T, T),
-    Many(Vec<T>),
+    Many(Box<[T]>),
 }
```
mzabaluev: You could soon (now?) be able try to use const generics for the inline variant.

imhamt/src/hash.rs
```diff
@@ -37,7 +37,7 @@ pub struct LevelIndex(pub usize);
 
 impl LevelIndex {
     #[inline]
-    pub fn mask(&self) -> u32 {
+    pub const fn mask(&self) -> u32 {
         1u32.wrapping_shl(self.0 as u32)
```
mzabaluev: Is this meant to silently produce a zero bit mask for levels above 32? There is probably no other alternative if you want to have it as a const fn, just wanted to make sure.

imhamt/src/helper.rs
```diff
@@ -1,15 +1,15 @@
 #[inline]
-pub fn clone_array_and_insert_at_pos<A: Clone>(v: &Vec<A>, a: A, pos: usize) -> Vec<A> {
+pub fn clone_array_and_insert_at_pos<A: Clone>(v: &Box<[A]>, a: A, pos: usize) -> Box<[A]> {
```
mzabaluev: Why does the parameter type have to be a `Box` if you clone it anyway?
I think you can just pass a slice that `Box`, `Vec`, or `Bytes` can be deref-converted to.

imhamt/src/content.rs
mzabaluev: Is the branching between two variants worth having an inline form, BTW?
I had a similar [discussion](https://github.com/tokio-rs/bytes/issues/268) about `Bytes` recently, and they decided to cut down on internal dynamic polymorphism to speed up the main use cases.

NicolasDP: Here the main use case is often one output to track.

imhamt/src/helper.rs
vincenthz: ah yes, you're right. I got carried away

imhamt/src/hash.rs
vincenthz: yes, it's fine in this case, it should never need more

PR #158, opened 2019-11-19
Add tag for utxo witness

SebastienGllmt:
This PR is the result of @vincenthz 's comment: https://github.com/input-output-hk/chain-libs/pull/130#issuecomment-547757578

Up to you to decide if you want a tag for UTXO witnesses also. Just wanted to make the PR so that we don't forget to make the decision

### DIFF
--- a/chain-impl-mockchain/src/transaction/witness.rs
+++ b/chain-impl-mockchain/src/transaction/witness.rs
@@ -56,8 +56,13 @@ impl std::fmt::Display for Witness {
 pub struct WitnessUtxoData(Vec<u8>);
 
 impl WitnessUtxoData {
-    pub fn new(block0: &HeaderId, transaction_id: &TransactionSignDataHash) -> Self {
+    pub fn new(block0: &HeaderId, transaction_id: &TransactionSignDataHash, is_legacy: bool) -> Self {
         let mut v = Vec::with_capacity(65);
+        let tag = match is_legacy {
+            true => WITNESS_TAG_OLDUTXO,
+            false => WITNESS_TAG_UTXO,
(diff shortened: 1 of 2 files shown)


chain-impl-mockchain/src/transaction/witness.rs
```diff
@@ -56,8 +56,13 @@ impl std::fmt::Display for Witness {
 pub struct WitnessUtxoData(Vec<u8>);
 
 impl WitnessUtxoData {
-    pub fn new(block0: &HeaderId, transaction_id: &TransactionSignDataHash) -> Self {
+    pub fn new(block0: &HeaderId, transaction_id: &TransactionSignDataHash, is_legacy: bool) -> Self {
```
vincenthz: this shouldn't be a bool, something like (not keen on "Normal" but couldn't think of anything else):

```pub enum UtxoWitnessType { Normal, Legacy } ```

File: chain-impl-mockchain/src/transaction/witness.rs [hunk 2]
```diff
@@ -56,8 +56,13 @@ impl std::fmt::Display for Witness {
...
         let mut v = Vec::with_capacity(65);
+        let tag = match is_legacy {
+            true => WITNESS_TAG_OLDUTXO,
+            false => WITNESS_TAG_UTXO,
+        };
+        v.push(tag);
```
vincenthz: This is a breaking cryptographic change, so this need some coordination with others here.

vincenthz:
Thanks, this is sensible to apply before the release (but later than now, otherwise that create issues with the already released test run)

PR #160, opened 2019-11-19
Rename Ledger::settings to settings_mut and add settings getter

CodeSandwich:
Part of https://github.com/input-output-hk/jormungandr/issues/1137

This is technically a breaking change, but it breaks no code in chain-libs or Jormungandr.

### DIFF
--- a/chain-impl-mockchain/src/ledger/ledger.rs
+++ b/chain-impl-mockchain/src/ledger/ledger.rs
@@ -34,6 +34,7 @@ pub struct LedgerStaticParameters {
 pub struct LedgerParameters {
     pub fees: LinearFee,
     pub reward_params: Option<RewardParams>,
+    pub max_number_of_transactions_per_block: u32,
 }
 
 /// Overall ledger structure.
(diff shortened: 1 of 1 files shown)


NicolasDP:
While I appreciate the API change. I am not sure that exposing the settings is still a good thing. Everything has been done, so far, to not expose these values as they have meaning that are not meant to be shared.

For working on #input-output-hk/jormungandr#1137 you need to use only the `LeaderParameters`. They are the only valid values worth using. And you can only get it from the appropriate `tip`.

CodeSandwich:
It turns out that `Ledger::settings()` isn't used anywhere anymore. I've removed it completely.

NicolasDP:
Thanks. I think this is going toward the right direction. However, there is also a reason why this value was not exposed in the first place: we are not sure sure how to enforce it just yet. Here we are counting in number of transactions, but we want to count in number of fragments per block.

Also:

* this value is not checked by the ledger at the moment;
* all fragments are not equals, some have 255 inputs and 255 outputs for example, Would it not be better to have a limit in byte size instead?

CodeSandwich:
I'm going to close this PR, it makes no sense to push this functionality any more, it's broken and slated for replacement

PR #161, opened 2019-11-20
Add fee calculation functions to InputOutput

SebastienGllmt:
# Background

In Yoroi, building and signing the transaction are totally separate steps (multiple prompts later). I will denote this break with `-----`

The old transaction creation flow had two main steps

1) Create a `TransactionBuilder` and add inputs/outputs
`-----`
2) Add certificate data/witnesses and sign

Since the new transaction creation flow, you have to do the following

1) Create an `InputOutputBuilder` and add inputs/outputs
2) Once finished, turn it into an immutable `InputOutput`
`-----`
3) Add certificate data/witnesses and sign

# The problem

- `InputOutputBuilder` does NOT let you get a list of inputs/outputs added so far but lets you calculate the fee
` InputOutput` does let you get a list of inputs/outputs but does NOT let you calculate the fee

That means to show the user a confirmation screen before signing the transaction, we need to use both (1) and (2) so we can show both the inputs/outputs in the transaction and the fee for the transactions. It's kind of ugly to have to do this.

# The solution

I add the methods to estimate the fee from `InputOutputBuilder` to  `InputOutput`.

### DIFF
--- a/chain-impl-mockchain/src/transaction/io.rs
+++ b/chain-impl-mockchain/src/transaction/io.rs
@@ -61,6 +61,49 @@ pub enum OutputPolicy {
     Forget,
 }
 
+impl InputOutput {
+    pub fn balance(&self, fee: Value) -> Result<Balance, ValueError> {
+        let inputs = Value::sum(self.inputs.iter().map(|i| i.value()))?;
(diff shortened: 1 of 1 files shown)


SebastienGllmt:
Closing this since I realized you don't need these functions because you can get the fee by just doing `output-input`

PR #165, opened 2019-11-22
Add legacy daedalus key type

SebastienGllmt:
# Why do we need this

The Shelly genesis block contains legacy Daedalus UTXOs in it and we want to be able to create transactions that transfer these funds to a Shelley wallet from Yoroi (basically port the existing "transfer from a Daedalus wallet" to be Shelley-compatible)

# Problem

Why can't use the existing `ed25519bip32` key type because Daedalus keys do not follow the normalization rules for bip32 keys. 

You can see this yourself by doing

```
echo '6d54e6ecbed6ecb5a163597f96a38127f8d1a12406837c0e778a2dc7e63395051391241669ba90ae293b8d7a0c8c654549fc3f5b97bed722dda5382e834a508506b6bd7a7baa2a5dc191cd08f0ca81ada7298cfa20db44d7eda31e7777b4bbe0' > foo

jcli key from-bytes --type ed25519bip32 foo

> Invalid secret key
```

We can't use `ed25519extended` either because we need to include the chaincode into the witness (similar to `ed25519bip32`.

This means we need to create a new key type for this case

# Why can you not just bypass the validation

There are two main functions for constructing an `XPrv`

- `from_bytes_verified` (which verifies the bip32 normalization rules)
- `from_bytes` (which DOESN'T verify the normalization rules)

In the v2 bindings, we had access to `from_bytes` so we called it directly in the Daedalus case. However, in `chain-libs` we no longer have access to `from_bytes` .

Fortunately we don't actually need any functions from `XPrv` so instead I just create a `pub struct LegacyPriv([u8; XPRV_SIZE]);` which does the job

### DIFF
--- a/chain-crypto/src/algorithms/legacy_daedalus.rs
+++ b/chain-crypto/src/algorithms/legacy_daedalus.rs
@@ -1,0 +1,179 @@
+use crate::key::{
+    AsymmetricKey, AsymmetricPublicKey, PublicKeyError, SecretKeyError, SecretKeySizeStatic,
+};
+use crate::sign::{SignatureError, SigningAlgorithm, Verification, VerificationAlgorithm};
+
+use cryptoxide::digest::Digest;
+use cryptoxide::hmac::Hmac;
+use cryptoxide::mac::Mac;
+use cryptoxide::sha2::Sha512;
+
+use super::ed25519 as ei;
+use cryptoxide::ed25519;
+use ed25519_bip32::{XPrv, XPub, XPRV_SIZE, XPUB_SIZE};
+use rand_core::{CryptoRng, RngCore};
+
+const CHAIN_CODE_SIZE: usize = 32;
+const SEED_SIZE: usize = 64;
+
+/// Legacy Daedalus algorithm
+pub struct LegacyDaedalus;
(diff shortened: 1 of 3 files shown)


chain-crypto/src/algorithms/legacy_daedalus.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+impl AsymmetricKey for LegacyDaedalus {
+  type Secret = LegacyPriv;
+  type PubAlg = LegacyDaedalus;
+
+  const SECRET_BECH32_HRP: &'static str = "legacy_xprv";
+
+  fn generate<T: RngCore + CryptoRng>(mut _rng: T) -> Self::Secret {
+      panic!("generation not implemented for legacy Daedalus wallets")
```
SebastienGllmt: The panic here causes the test at the bottom of this file to fail. Not sure if there's a good way to provide an arbitrary instance only for when the test is called.

vincenthz:
The key in example is extremely problematic and really shouldn't "go through"

vincenthz:
genuinely insecure key shouldn't be used.

PR #167, opened 2019-11-22
Specify old_utxo only works for Icarus-style old UTXOs

SebastienGllmt:
It may be that we don't want to merge https://github.com/input-output-hk/chain-libs/pull/165 to avoid introducing legacy functionality into this codebase. If that's the case, I propose this alternative PR that simply does a rename to make it clear that old UTXOs only work for Icarus-style UTXOs (since `ed25519bip32` does not work for legacy Daedalus-style UTXOs)

This should make the limitation clearer for anybody who reads this later


vincenthz:
the code path is generic to any `XPrv`

PR #168, opened 2019-11-25
Implement fees for different certificate types

### LINKED ISSUE #9: per-certificate fees
Certificate might have different overhead on the chain, and thus might require more fees.

Introduce a certificate trait that allow to define extra fees depending on the certificate type and content, effectively `Certificate -> ExtraFee`

related to #427

eugene-babichenko:
Allow setting different fees for `PoolRegistration`, `StakeDelegation` and `OwnerStakeDelegation`. Fees for other types of certificates remain unchanged.

Resolves #9

### DIFF
--- a/chain-impl-mockchain/src/fee.rs
+++ b/chain-impl-mockchain/src/fee.rs
@@ -9,6 +9,14 @@ pub struct LinearFee {
     pub constant: u64,
     pub coefficient: u64,
     pub certificate: u64,
+    pub per_certificate_fees: Option<PerCertificateFee>,
+}
+
+#[derive(PartialEq, Eq, PartialOrd, Debug, Clone, Copy)]
+pub struct PerCertificateFee {
+    pub certificate_pool_registration: u64,
+    pub certificate_stake_delegation: u64,
+    pub certificate_owner_stake_delegation: u64,
 }
 
 impl LinearFee {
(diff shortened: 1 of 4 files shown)


vincenthz:
sorry for the misunderstanding, I think the idea is to maybe leave `LinearFee` as it is and add a PerCertificateFee optional ConfigParams that will selectively override some value, but keep the certificate fee in the LinearFee as a value that apply if nothing else has overrided the value.

eugene-babichenko:
Got it. Will rework it a bit then.

eugene-babichenko:
@vincenthz Added a separate `ConfigParam` for certificate fees. Thus, `LinearFee` serialisation remained unchanged.

chain-impl-mockchain/src/fee.rs
```diff
@@ -45,13 +85,20 @@ impl FeeAlgorithm for LinearFee {
...
         )
     }
-    fn fees_for_certificate<'a>(&self, _: CertificateSlice<'a>) -> Value {
+
+    fn fees_for_certificate<'a>(&self, cert_slice: CertificateSlice<'a>) -> Value {
+        if let Some(per_certificate_fees) = self.per_certificate_fees {
+            if let Some(fee) = per_certificate_fees.fees_for_certificate(cert_slice) {
+                return fee;
+            }
+        }
```
NicolasDP: it does not seem to be right to me. @vincenthz said: https://github.com/input-output-hk/chain-libs/pull/168#issuecomment-558119527

> keep the certificate fee and override some values (simplified).

Here you override the whole certificate fee if this value is set. While actually what was asked is to override some of the values. Not necessarily all of them.

i.e. We may still want to have a `LinearFees::certificate` set to a value for all. Except for the stake pool registration certificate where we want to have a different value.

eugene-babichenko: > Here you override the whole certificate fee if this value is set

No. There are just 3 types of certificate in `PerCertificateFee`. Probably, it would be a good idea to make every field of `PerCertificateFee` optional.

Note that `PerCertificateFee::fees_for_certificate` may return `None`, in that case we just fall back to `LinearFees::certificate`.

PR #172, opened 2019-11-26
Remove TransactionSlice::into_owned

### LINKED ISSUE #162: optimisation: don't own transaction from slice to verify balance
https://github.com/input-output-hk/chain-libs/blob/master/chain-impl-mockchain/src/ledger/ledger.rs#L484

eugene-babichenko:
Resolves #162

### DIFF
--- a/chain-impl-mockchain/src/fee.rs
+++ b/chain-impl-mockchain/src/fee.rs
@@ -32,9 +32,9 @@ pub trait FeeAlgorithm {
             .saturating_add(cert.map_or(Value::zero(), |c| self.fees_for_certificate(c)))
     }
 
-    fn calculate_tx<P: tx::Payload>(&self, tx: &tx::Transaction<P>) -> Value {
+    fn calculate_tx<P: tx::Payload>(&self, tx: &tx::TransactionSlice<P>) -> Value {
         self.calculate(
-            tx.as_slice().payload().to_certificate_slice(),
+            tx.payload().to_certificate_slice(),
             tx.nb_inputs(),
             tx.nb_outputs(),
         )
--- a/chain-impl-mockchain/src/ledger/ledger.rs
+++ b/chain-impl-mockchain/src/ledger/ledger.rs
@@ -481,7 +481,7 @@ impl Ledger {
     {
         check::valid_transaction_ios_number(tx)?;
         let fee = calculate_fee(tx, dyn_params);
-        tx.into_owned().verify_strictly_balanced(fee)?;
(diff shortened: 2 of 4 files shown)


eugene-babichenko:
What was the purpose of introducing `TransactionSlice`? I can't see what are the benefits. Given current use cases we can just pass `&Transaction` instead of `&TransactionSlice`. At the same time, dealing with both of those types leads to unnecessary type conversions and code duplication.

vincenthz:
The idea behind all this (and that you cannot replicate with a simple `&`), is that you have recursively
different slice types like:

```
|Block............................................................................................................................|
|HdrSlice|ContentSlice........................................................................................................|
                 |FragmentSlice..............................................|
                  | TransactionSlice 1 |    | TransactionSlice 2 |
```

If you had untyped generic `Vec` and `&[]` it would be more apparent why this is structured this way, and would less code also, but the shape of Borrow/Owned trait make this difficult without using `mem::transmute`

`into_owned` is really about given a complete api (which may or not be used currently), so that you can turn any of this object which represent valid something into a valid `T` (TransactionSlice -> Transaction) in the most typed way

note that `&Object` where `Object(Box<[a]>)` is also different from `ObjectSlice(&[a])` since one is a reference (pointer of 8 bytes) to a fat pointer, where as the second one is a fat pointer directly (16 bytes). This is not a big optimisation since we don't keep thousand of those things live, but one nonetheless too.

SebastienGllmt:
I ran into this a while back when I was trying to add a deserialize function to `Transaction` (which required you to use `into_owned` from a slice) but there wasn't any obvious way to deduce what the template parameter for `Transaction` should be so it instead made for sense to just use the deserialize logic for `Fragment` which has a tag in it which lets you deduce the template parameter of the contained `Transaction`.

tl;dr conversion from `Vec` -> `Transaction` not so meaningful because it's better to go `Vec` -> `Fragment` -> `Transaction`

eugene-babichenko:
@vincenthz Ok, then I'll revert my last commit since it breaks the entire idea of `TransactionSlice`. The question that still remains is how to avoid excessive code duplication.

I noticed that almost all methods refer to the `tstruct` field and don't use any other fields. My suggestion here is to reimplement those methods in `TransactionStruct` and then implement the delegation pattern in `Transaction`/`TransactionSlice` or (may look as a bit of a hack, but I actually would prefer this one) implement `Deref<Target = TransactionStruct>` for `Transaction` and `TransactionSlice`.

PR #175, opened 2019-11-27
Incentive config

vincenthz:
* Add the ability to set all the parameters in the config
* Cleanup the types to be more explicit (tuples -> structs)
* Tweak documentation & add some example

### DIFF
--- a/chain-impl-mockchain/src/config.rs
+++ b/chain-impl-mockchain/src/config.rs
@@ -1,3 +1,4 @@
+use crate::header::Epoch;
 use crate::leadership::bft::LeaderId;
 use crate::milli::Milli;
 use crate::rewards::TaxType;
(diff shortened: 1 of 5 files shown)


chain-impl-mockchain/src/config.rs
```diff
@@ -78,8 +79,20 @@ pub enum ConfigParam {
 
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum RewardParams {
-    Linear(u64, u64, NonZeroU64),
-    Halving(u64, u64, NonZeroU64),
+    Linear {
+        constant: u64,
+        ratio_num: u64,
+        ratio_denom: NonZeroU64,
```
NicolasDP: why not re-using the same `Ratio` type that you already have?

PR #176, opened 2019-11-28
On some architecture, the get_first_index_impl fails to returns a valid value

NicolasDP:
see error in https://travis-ci.org/input-output-hk/jormungandr/jobs/618020686

The error:

```
Error: thread 'main' panicked at 'FastSparseArray does not contains item at index 65', chain-deps/sparse-array/src/fast.rs:130:21
```

There was only one entry in the utxos

Removing `bmi1` fixes the problem

### DIFF
--- a/sparse-array/src/bitmap.rs
+++ b/sparse-array/src/bitmap.rs
@@ -75,7 +75,7 @@ impl BitmapIndex {
     }
 
     #[inline]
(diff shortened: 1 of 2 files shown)


eugene-babichenko:
Might be related to input-output-hk/jormungandr#1144

> There was only one entry in the utxos

That's not reproducible if the array has more than one entry, right?

NicolasDP:
may be. There was one entry. If it crashes on travis-ci, it may crash on other system.

PR #178, opened 2019-11-28
Stop using target_feature from the code

eugene-babichenko:
Users who want to build the project should use rustc flags instead.

Example: `-C target-feature=+popcnt,+bmi1`

This also reduces the amount of potentially unsafe code.

### DIFF
--- a/imhamt/src/bitmap.rs
+++ b/imhamt/src/bitmap.rs
@@ -68,14 +68,9 @@ impl SmallBitmap {
         self.0 == 0
     }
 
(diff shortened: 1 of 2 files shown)


eugene-babichenko:
To me an open question is how to control the usage of particular features. The simplest approach would be using `RUSTFLAGS="-C target-feature=+popcnt,+bmi2" cargo build ...`.

vincenthz:
might be nice to add some note of the RUSTFLAG magic in the README for enquiring users

PR #183, opened 2019-11-29
Add missing cert methods

SebastienGllmt:
Two small things added in this PR

# 1 - Missing `serialize` methods on certificates

Some certificates had a `serialize` function and others only had `serizalize_in`. We need `serialize` because `typed_bytes` isn't available from the WASM bindings (which is why I assume the other certs have `serialize_in` in the first place)

# 2 - GenesisProasLeaderHash as a type

I need this because it's an argument for creating a `PoolUpdate` certificate from the WASM.

### DIFF
--- a/chain-impl-mockchain/src/certificate/pool.rs
+++ b/chain-impl-mockchain/src/certificate/pool.rs
@@ -19,7 +19,7 @@ use typed_bytes::{ByteArray, ByteBuilder};
 pub type PoolId = DigestOf<Blake2b256, PoolRegistration>;
 
 /// Hash of keys used for pool
-pub type GenesisProasLeaderHash = DigestOf<Blake2b256, GenesisPraosLeader>;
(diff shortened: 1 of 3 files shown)


chain-impl-mockchain/src/certificate/pool.rs
```diff
@@ -18,6 +18,9 @@ use typed_bytes::{ByteArray, ByteBuilder};
 /// Pool ID
 pub type PoolId = DigestOf<Blake2b256, PoolRegistration>;
 
+/// Hash of keys used for pool
+pub type GenesisProasLeaderHash = DigestOf<Blake2b256, GenesisPraosLeader>;
```
vincenthz: typo `Proas`

SebastienGllmt: Oops 🤦 fixed

PR #185, opened 2019-12-02
fix double update of the stake distribution

NicolasDP:
Found with input-output-hk/jormungandr#1242


vincenthz:
@dkijania please look on adding some tests relating to value and stake

PR #186, opened 2019-12-02
Starting removing ed25519 bip32 for direct use

vincenthz:
* move towards requiring only public interface in chain implementation
* ed25519 public key are all the same

### DIFF
--- a/cardano-legacy-address/Cargo.toml
+++ b/cardano-legacy-address/Cargo.toml
@@ -17,7 +17,7 @@ edition = "2018"
 [dependencies]
 cryptoxide = "0.1"
 cbor_event = "^2.1.3"
-ed25519-bip32 = "^0.1.1"
+ed25519-bip32 = "^0.3.0"
 
 [features]
 default = []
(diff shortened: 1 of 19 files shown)


SebastienGllmt:
Is it possible that you please not merge this until after the first release of Shelley this month? Otherwise if you make any last-second breaking changes we won't be able to update our fork of chain-libs without ugly merge conflicts that we don't have time to deal with right now 🙏

vincenthz:
I can indeed old off for a bit, but the only right option is to move to the provider of xprv/xpub for any generation/derivation related business (ed25519-bip32 package) and only deal in term of ed25519 signature and public keys which have absolutely no bearing on whether or not they are normal ed25519, extended ed25519 or xprv.

vincenthz:
btw, without a serious plan to either merge back and have a path forward though, this will be a useless wait.

SebastienGllmt:
Yes I will deal with whatever changes need to be made after the release goes out. At that point you can merge this PR and we will make it work.

SebastienGllmt:
You can merge this now if you want to since we have time to deal with re-adding this on our end

vincenthz:
fix #258

PR #201, opened 2019-12-04
Fix linear rewards reset to inintial after reduction to 0

CodeSandwich:
I'm quite certain that this is a bug


rinor:
Since you are fixing rewards, can you check also
https://github.com/input-output-hk/chain-libs/blob/212300db42f12d26d4cd6df46a5de974097cb0af/chain-impl-mockchain/src/rewards.rs#L153
and
https://github.com/input-output-hk/chain-libs/blob/212300db42f12d26d4cd6df46a5de974097cb0af/chain-impl-mockchain/src/rewards.rs#L190
if they need fixing like in https://github.com/input-output-hk/chain-libs/pull/197/commits/ec1cb622b6c81503217415f9604c2556237b031c

NicolasDP:
@rinor , they do 😱 In another PR though

PR #202, opened 2019-12-04
rewards: replace 'binary XOR' by the actually meant value

rinor:
https://github.com/input-output-hk/chain-libs/pull/201 comments


NicolasDP:
thanks a lot @rinor

PR #203, opened 2019-12-04
changes to the incentive model

JaredCorduan:
This PR is code from @kevinhammond that updates the incentives model. There will be a corresponding change for jormungandr.

### DIFF
--- a/chain-impl-mockchain/Cargo.toml
+++ b/chain-impl-mockchain/Cargo.toml
@@ -11,8 +11,8 @@ edition = "2018"
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-serde = { version = "^1.0", optional = true }
-serde_derive = { version = "^1.0", optional = true }
+serde = { version = "1.0", features = ["derive"] }
+serde_derive = { version = "^1.0" }
 chain-core = { path = "../chain-core" }
 chain-addr = { path = "../chain-addr" }
 chain-crypto = { path = "../chain-crypto" }
--- a/chain-impl-mockchain/src/config.rs
+++ b/chain-impl-mockchain/src/config.rs
@@ -1,7 +1,7 @@
 use crate::header::Epoch;
 use crate::leadership::bft::LeaderId;
 use crate::milli::Milli;
-use crate::rewards::{Ratio, TaxType};
+use crate::rewards::{PoolLimit, Ratio, RewardLimitByStake, TaxType};
 use crate::value::Value;
 use crate::{
     block::ConsensusVersion,
@@ -75,6 +75,8 @@ pub enum ConfigParam {
     RewardParams(RewardParams),
     PerCertificateFees(PerCertificateFee),
     FeesInTreasury(bool),
+    RewardLimitByStake(RewardLimitByStake),
+    PoolLimit(PoolLimit),
 }
 
 #[derive(Clone, Debug, Eq, PartialEq)]
@@ -134,6 +136,10 @@ pub enum Tag {
     PerCertificateFees = 21,
     #[strum(to_string = "fees-in-treasury")]
     FeesInTreasury = 22,
+    #[strum(to_string = "reward-limit-by-stake")]
+    RewardLimitByStake = 23,
+    #[strum(to_string = "pool-limit")]
+    PoolLimit = 24,
 }
 
 impl Tag {
@@ -158,6 +164,8 @@ impl Tag {
             20 => Some(Tag::RewardParams),
             21 => Some(Tag::PerCertificateFees),
             22 => Some(Tag::FeesInTreasury),
+            23 => Some(Tag::RewardLimitByStake),
+            24 => Some(Tag::PoolLimit),
             _ => None,
         }
     }
@@ -187,6 +195,8 @@ impl<'a> From<&'a ConfigParam> for Tag {
             ConfigParam::RewardParams(_) => Tag::RewardParams,
             ConfigParam::PerCertificateFees(_) => Tag::PerCertificateFees,
             ConfigParam::FeesInTreasury(_) => Tag::FeesInTreasury,
+            ConfigParam::RewardLimitByStake(_) => Tag::RewardLimitByStake,
+            ConfigParam::PoolLimit(_) => Tag::PoolLimit,
         }
     }
 }
@@ -246,6 +256,10 @@ impl Readable for ConfigParam {
             Tag::FeesInTreasury => {
                 ConfigParamVariant::from_payload(bytes).map(ConfigParam::FeesInTreasury)
             }
+            Tag::RewardLimitByStake => {
+                ConfigParamVariant::from_payload(bytes).map(ConfigParam::RewardLimitByStake)
+            }
+            Tag::PoolLimit => ConfigParamVariant::from_payload(bytes).map(ConfigParam::PoolLimit),
         }
         .map_err(Into::into)
     }
@@ -276,6 +290,8 @@ impl property::Serialize for ConfigParam {
             ConfigParam::RewardParams(data) => data.to_payload(),
             ConfigParam::PerCertificateFees(data) => data.to_payload(),
             ConfigParam::FeesInTreasury(data) => data.to_payload(),
+            ConfigParam::RewardLimitByStake(data) => data.to_payload(),
+            ConfigParam::PoolLimit(data) => data.to_payload(),
         };
         let taglen = TagLen::new(tag, bytes.len()).ok_or_else(|| {
             io::Error::new(
@@ -318,6 +334,32 @@ impl ConfigParamVariant for TaxType {
         let tax_type = TaxType::read_frombuf(&mut rb)?;
         rb.expect_end()?;
         Ok(tax_type)
+    }
+}
+
+impl ConfigParamVariant for RewardLimitByStake {
+    fn to_payload(&self) -> Vec<u8> {
+        self.serialize_in(ByteBuilder::new()).finalize_as_vec()
+    }
+
+    fn from_payload(payload: &[u8]) -> Result<Self, Error> {
+        let mut rb = ReadBuf::from(payload);
+        let rewards_limit = RewardLimitByStake::read_frombuf(&mut rb)?;
+        rb.expect_end()?;
+        Ok(rewards_limit)
+    }
+}
+
+impl ConfigParamVariant for PoolLimit {
+    fn to_payload(&self) -> Vec<u8> {
+        self.serialize_in(ByteBuilder::new()).finalize_as_vec()
+    }
+
+    fn from_payload(payload: &[u8]) -> Result<Self, Error> {
+        let mut rb = ReadBuf::from(payload);
+        let pool_limit = PoolLimit::read_frombuf(&mut rb)?;
+        rb.expect_end()?;
+        Ok(pool_limit)
     }
 }
 
@@ -680,7 +722,7 @@ mod test {
 
     impl Arbitrary for ConfigParam {
         fn arbitrary<G: Gen>(g: &mut G) -> Self {
-            match u8::arbitrary(g) % 16 {
+            match u8::arbitrary(g) % 18 {
                 0 => ConfigParam::Block0Date(Arbitrary::arbitrary(g)),
                 1 => ConfigParam::Discrimination(Arbitrary::arbitrary(g)),
                 2 => ConfigParam::ConsensusVersion(Arbitrary::arbitrary(g)),
@@ -697,6 +739,8 @@ mod test {
                 13 => ConfigParam::RewardParams(Arbitrary::arbitrary(g)),
                 14 => ConfigParam::PerCertificateFees(Arbitrary::arbitrary(g)),
                 15 => ConfigParam::FeesInTreasury(Arbitrary::arbitrary(g)),
+                16 => ConfigParam::RewardLimitByStake(Arbitrary::arbitrary(g)),
+                17 => ConfigParam::PoolLimit(Arbitrary::arbitrary(g)),
                 _ => unreachable!(),
             }
         }
--- a/chain-impl-mockchain/src/ledger/ledger.rs
+++ b/chain-impl-mockchain/src/ledger/ledger.rs
@@ -12,6 +12,7 @@ use crate::fragment::{BlockContentHash, BlockContentSize, Contents, Fragment, Fr
 use crate::header::{BlockDate, ChainLength, Epoch, HeaderContentEvalContext, HeaderId};
 use crate::leadership::genesis::ActiveSlotsCoeffError;
 use crate::rewards;
+use crate::rewards::{PoolLimit, RewardLimitByStake};
 use crate::stake::{PercentStake, PoolError, PoolStakeInformation, PoolsState, StakeDistribution};
 use crate::transaction::*;
 use crate::treasury::Treasury;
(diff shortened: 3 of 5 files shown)


JaredCorduan:
> This looks like a result of a mismerge as it reset the whole tree to what was available 2 or 3 days ago

ah, I thought @kevinhammond 's changes were based off of where the submodule was when the main repo was at `v0.8.0-rc6`, but I guess that's not correct. I'll try to find the right common root.

vincenthz:
Briefly over the main changes, 2 early comments: I don't think it make sense to extend the existing param, I think adding a new param (tentative name: `RewardMinStakeThreshold`) would be much better, as this does override the whole mechanism, not specific parameter.

Similarly for npools, npools_threshold as it's not implemented, this really shouldn't be there, and when it is, it's likely making sense as a new modular config param.

kevinhammond:
Looks like Jared has dealt with the merge issues.  I'd had to reset the base having started on rc5, I think and didn't want to accidentally touch master

kevinhammond:
Feel free to choose alternative names and settings sections. rewards_limit can apply to both the halving and linear mechanisms.  At 100% it has no effect.

 The npools/pools_threshold mechanisms should be implemented (in rewards.rs and ledger.rs).

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -381,6 +389,12 @@ impl Ledger {
         if total_reward > Value::zero() {
             let total_blocks = leaders_log.total();
             let reward_unit = total_reward.split_in(total_blocks);
+            let mut pool_count = 0;
+
+            // Is there a neater way to do this?
+            for (pool_id, pool_blocks) in leaders_log.iter() {
```
vincenthz: `leaders_log.iter().len()`

chain-impl-mockchain/src/rewards.rs
```diff
@@ -276,13 +305,17 @@ mod tests {
                 }
                 NonZeroU32::new(number).unwrap()
             };
+            let npools_threshold = u32::arbitrary(g) + 1;
```
vincenthz: this can overflow when `u32::arbitrary() == u32::max_value()`

chain-impl-mockchain/src/rewards.rs
```diff
@@ -163,6 +174,24 @@ pub fn rewards_contribution_calculation(epoch: Epoch, params: &Parameters) -> Va
...
+pub fn rewards_contribution_calculation_scaled(
+    epoch: Epoch,
+    params: &Parameters,
+    total_delegated: Stake,
+) -> Value {
+    let r = rewards_contribution_calculation(epoch, params);
+
+    // limit the rewards if necessary
+    let rl = &params.rewards_limit;
+    const SCALE: u128 = 10 ^ 18;
```
vincenthz: That's a `xor` not a `power` function.

chain-impl-mockchain/src/rewards.rs
```diff
@@ -163,6 +174,24 @@ pub fn rewards_contribution_calculation(epoch: Epoch, params: &Parameters) -> Va
...
+
+    let limit_scaled = u64::from(total_delegated) as u128 * SCALE;
+    let limit = limit_scaled / rl.denominator.get() as u128;
+    cmp::min(Value((limit / SCALE) as u64), r)
+}
+
```
vincenthz: doesn't really make sense to me. shouldn't this be the ratio of delegated stake against the potential total stake ?

vincenthz:
Thanks @JaredCorduan for fixing many issues, but right now the major deal breaker is that it's modifying the reward parameter where as it should be a 3 new ConfigParams in config.rs:

* `RewardLimitByStake(Ratio)`
* `PoolLimit(u32, u32)`
* `FeeInTreasury(bool)`

chain-impl-mockchain/src/config.rs
```diff
@@ -328,25 +334,39 @@ impl ConfigParamVariant for RewardParams {
...
+                .u32(npools.get())
+                .u32(npools_threshold.get()),
             RewardParams::Halving {
                 constant,
                 ratio,
                 epoch_start,
                 epoch_rate,
+                rewards_limit,
+                npools,
+                npools_threshold,
```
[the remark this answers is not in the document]
kevinhammond: I will fix this.  I'm not seeing it in my copy of the code, so it may be the result of a merge error.  It would have passed the tests, of course, since we're using the same values.  I will change that, so we can spot any further error.

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -353,14 +355,20 @@ impl Ledger {
...
+            distribution.total_stake(),
+        );
 
         let mut total_reward = new_ledger.pots.draw_reward(expected_epoch_reward);
 
+        // Changed from true to false -- rewards *must* accrue to the treasury in the testnet.   KH
         // Move fees in the rewarding pots for distribution or depending on settings
         // to the treasury directly
-        if true {
+        if false {
```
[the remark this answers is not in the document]
kevinhammond: Not sure I understand this, since there is an if false, which is not doing the right thing (as agreed for the testnet).  We cannot move fees to rewards.  It is a vulnerability.

chain-impl-mockchain/src/ledger/ledger.rs
kevinhammond: Thanks!

chain-impl-mockchain/src/rewards.rs
kevinhammond: OK. I don't remember that code, but will check it.  There's no need for the threshold to be that large in practice.

chain-impl-mockchain/src/rewards.rs
kevinhammond: OK.  This seems to be consistent with the previous uses.  Are there some bugs there?  I will fix them if so.  I see you had spotted the epoch=0 error for delegators that was previously there.

chain-impl-mockchain/src/rewards.rs
kevinhammond: You don't need to know the total potential stake (which might anyway need a new fixed constant parameter to limit the system).  The rewards can be scaled against the delegated stake.

chain-impl-mockchain/src/ledger/ledger.rs
kevinhammond: len() doesn't seem to exist.  I have used count() instead, which is a bit cleaner than a loop.

chain-impl-mockchain/src/rewards.rs
kevinhammond: This was copied from the existing Rust code. I see you've now fixed the bugs that you had.  I have made this consistent.

chain-impl-mockchain/src/config.rs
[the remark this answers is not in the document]
vincenthz: Jared fixed this one already

chain-impl-mockchain/src/ledger/ledger.rs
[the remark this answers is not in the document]
vincenthz: this is a library, which is not about testnet or mainnet. we want this configurable and/or generics so that we can conduct tests/benchmarks of many possible scenario

chain-impl-mockchain/src/ledger/ledger.rs
vincenthz: my apologies for the mis-suggestion, I meant `count()` indeed ! glad you manage to find it.

chain-impl-mockchain/src/rewards.rs
vincenthz: indeed that would work too

kevinhammond: I've put in a test that shouldn't overflow.  We'll need to merge the code back later, but will wait until I have everything stable first.

kevinhammond:
I'm working on the config params, but have had to put some time into comms etc. (people are getting very excited about the testnet!).  Will carry on with them now - they are used in a few places, so I've had to study the control flow a bit.  We just need to get the system to a state where it will deliver predictable and consistent rewards that we can audit properly.  Alan seems to have found some things that could be down to inconsistent performance or something in the existing calculations - we'll work out how to iron these out.  Hopefully that won't be too hard.

vincenthz:
This comment has nothing to do with the task/issue at hand.

vincenthz:
Fees-in-treasury has been handled by #212

chain-impl-mockchain/src/rewards.rs
```diff
@@ -163,6 +174,30 @@ pub fn rewards_contribution_calculation(epoch: Epoch, params: &Parameters) -> Va
...
+    total_delegated: Stake,
+) -> Value {
+    let r = rewards_contribution_calculation(epoch, params);
+
+    // limit the rewards if necessary
+    let rl = &params.rewards_limit;
+    let res = match rl {
+        None => r,
+        Some(rlimit) => {
+            const SCALE: u128 = 10 ^ 18;
```
NicolasDP: this is still invalid

```suggestion
            const SCALE: u128 = 1_000_000_000_000_000_000;
```

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -353,14 +355,20 @@ impl Ledger {
 
         let epoch = new_ledger.date.epoch + 1;
 
-        let expected_epoch_reward =
-            rewards::rewards_contribution_calculation(epoch, &ledger_params.reward_params);
+        // changed to scale rewards calculation
+        let expected_epoch_reward = rewards::rewards_contribution_calculation_scaled(
+            epoch,
+            &ledger_params.reward_params,
+            // Pass total stake for all pools from distribution -- KH
```
NicolasDP: wha't s the meaning of `KH` ?

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -353,14 +355,20 @@ impl Ledger {
...
+            distribution.total_stake(),
+        );
 
         let mut total_reward = new_ledger.pots.draw_reward(expected_epoch_reward);
 
+        // Changed from true to false -- rewards *must* accrue to the treasury in the testnet.   KH
```
NicolasDP: what's the meaning of `KH` ?

chain-impl-mockchain/Cargo.toml
```diff
@@ -11,8 +11,8 @@ edition = "2018"
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-serde = { version = "^1.0", optional = true }
-serde_derive = { version = "^1.0", optional = true }
+serde = { version = "1.0", features = ["derive"] }
+serde_derive = { version = "^1.0" }
```
vincenthz: that shouldn't be modified

chain-impl-mockchain/src/rewards.rs
```diff
@@ -249,6 +329,8 @@ mod tests {
         );
     }
 
+    // would be good to add some additional reward calculation tests here, including scaling etc. KH
```
vincenthz: superfluous comment

chain-impl-mockchain/src/rewards.rs
```diff
@@ -267,6 +349,32 @@ mod tests {
...
+                denominator: NonZeroU32::new(denominator).unwrap(),
+            }
+        }
+    }
+
+    impl Arbitrary for PoolLimit {
+        fn arbitrary<G: Gen>(gen: &mut G) -> Self {
+            // ugly, but avoids generating unwanted 0.  KH
+            let npools = u32::arbitrary(gen) + 1;
+            let npools_threshold = u32::arbitrary(gen) + 1;
```
vincenthz: should be `(u32::arbitrary(gen) % REASONABLE_VALUE) + 1` to avoid the overflow

chain-impl-mockchain/src/rewards.rs
```diff
@@ -81,6 +84,64 @@ impl TaxType {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
```
vincenthz: we don't derive Serde stuff usually in chain-libs, favoring a stable implementation defined by the user of this library.

chain-impl-mockchain/src/rewards.rs
```diff
@@ -81,6 +84,64 @@ impl TaxType {
...
+
+    pub fn read_frombuf(rb: &mut ReadBuf) -> Result<Self, ReadError> {
+        let npools = rb.get_u32()?;
+        let npools_threshold = rb.get_u32()?;
+
+        Ok(PoolLimit {
+            npools: NonZeroU32::new(npools).unwrap(),
+            npools_threshold: NonZeroU32::new(npools_threshold).unwrap(),
+        })
+    }
```
vincenthz: This not particularly important, but just as FYI, this could be implemented by the Readable class also

chain-impl-mockchain/src/rewards.rs
```diff
@@ -163,6 +224,25 @@ pub fn rewards_contribution_calculation(epoch: Epoch, params: &Parameters) -> Va
     }
 }
 
+// Added new function to ensure rewards calculation complies with research results and commercial needs -- KH
```
vincenthz: this comment doesn't need to be there. chain-libs is agnostic to commercial needs, and will make sure to support every uses.

vincenthz:
Thanks @JaredCorduan this is looking much better already, this is very close

JaredCorduan:
> Thanks @JaredCorduan this is looking much better already, this is very close

you can thank @kevinhammond , he did all the heavy lifting :)

chain-impl-mockchain/src/rewards.rs
kevinhammond: We have fixed this to make it consistent with your other bug changes.

chain-impl-mockchain/src/ledger/ledger.rs
kevinhammond: Just an indicator to Jared and me that this is something I have changed.  Should be removed in the latest version.

chain-impl-mockchain/Cargo.toml
kevinhammond: I needed to do that to derive Serialize/Deserialize for some of the data types. If you know a better way, please do change that.

chain-impl-mockchain/src/rewards.rs
kevinhammond: A placeholder/indicator that some more work could be done.  We'll remove it, of course.

chain-impl-mockchain/src/rewards.rs
kevinhammond: OK.  Will change - was using the same code as in other places.

chain-impl-mockchain/src/rewards.rs
kevinhammond: This was the quick thing to do at 0400 :)  If it doesn't matter for release, we can fix it up later.

chain-impl-mockchain/src/rewards.rs
kevinhammond: OK.  Since we're close to release, we should probably not do this at this point

chain-impl-mockchain/src/rewards.rs
kevinhammond: OK.  It was just an explanation for why this section of code had been added.

PR #213, opened 2019-12-09
[Tests] Stake pool tax

dkijania:
- testing 3 parameters: fixed_tax, tax_ratio, max_limit

### DIFF
--- a/chain-impl-mockchain/src/testing/e2e/rewards/tax.rs
+++ b/chain-impl-mockchain/src/testing/e2e/rewards/tax.rs
@@ -1,0 +1,138 @@
+use crate::{
+    config::RewardParams,
+    rewards::Ratio,
+    testing::{
+        ledger::ConfigBuilder,
+        scenario::{prepare_scenario, stake_pool, template::StakePoolDefBuilder, wallet},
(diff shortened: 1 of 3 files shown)


chain-impl-mockchain/src/testing/e2e/rewards/reward_params.rs
```diff
@@ -0,0 +1 @@
+
```
vincenthz: empty file ?

chain-impl-mockchain/src/testing/e2e/rewards/tax.rs
```diff
@@ -0,0 +1,140 @@
...
+fn verify_distribute_rewards(
+    total_reward: u64,
+    stake_pool_builder: &mut StakePoolDefBuilder,
+    expected_stake_pool_reward: u64,
+) {
+    let (mut ledger, controller) = prepare_scenario()
+        .with_config(
+            ConfigBuilder::new(0)
+                .with_rewards(Value(total_reward + 1))
+                .with_treasury(Value(0))
```
vincenthz: better to use the `Value::zero()` constructor when possible

PR #218, opened 2019-12-10
Pool update

vincenthz:
add ability to update parameters of the pool


NicolasDP:
no merging until we have something ready in jormungandr for this too

PR #225, opened 2019-12-13
remove custom_error in favour of using thiserror

eugene-babichenko:
- Code becomes easier to navigate because now it uses the native enum syntax
- For the same reason it can be formatted with `rustfmt`
- More consistent error handling

Related to input-output-hk/jormungandr#1165


eugene-babichenko:
Note: the same was successfully merged in `jormungandr`

PR #229, opened 2019-12-20
[Tests] compilation fix for property-test-api feature

dkijania:


chain-impl-mockchain/src/transaction/witness.rs
```diff
@@ -217,7 +217,7 @@ impl Readable for Witness {
     }
 }
 
-#[cfg(test)]
+#[cfg(any(test, feature = "property-test-api"))]
 pub mod test {
```
mzabaluev: Why was this module public?
Can we have all unit test modules uniformly named `tests`?

chain-impl-mockchain/src/account.rs
```diff
@@ -57,7 +57,7 @@ impl std::fmt::Display for Identifier {
     }
 }
 
-#[cfg(test)]
+#[cfg(any(test, feature = "property-test-api"))]
```
mzabaluev: This and below is done so that the `Arbitrary` impls are defined for the property test API, right?
I think it's better to move the impls to a separate module, so that unit test code in `mod tests` everywhere does not get compiled into the crate when we only wanted the property test impls.

mzabaluev:
Note that there is `quickcheck` 0.9 now which incorporates the macro.
I don't know how much of heavy lifting it is to update to the new version, though.

chain-impl-mockchain/src/transaction/witness.rs
dkijania: ok

chain-impl-mockchain/src/account.rs
dkijania: I've done my best for majority of modules. However I haven't changed modules that are direct children on main . I can move them all into single file `testing.rs` like in `chain_crypto`, but there are tons of changes already in project. So i can do it in separate pr

PR #232, opened 2020-01-02
allow cloning the sqlite pool in new impl

eugene-babichenko:
Port of #228 for sqlite with in-memory index


chain-storage-sqlite/src/lib.rs
```diff
@@ -106,13 +109,14 @@ where
     }
 }
 
+#[derive(Clone)]
```
NicolasDP: This will require `B` to be `Clone too. is it what we want?

eugene-babichenko: I think we won't need cloning at all once #239 is integrated.

PR #233, opened 2020-01-02
Correct order of operations in leaderlog set_for

### LINKED ISSUE #224: LeadersParticipationRecord is increased even on error
There is an issue in code:

https://github.com/input-output-hk/chain-libs/blob/master/chain-impl-mockchain/src/block/leaderlog.rs#L41

It first increment total and then check if insert is correct which can leads to wrong counter values for participants record. See test:

```
 #[test]
    pub fn test_set_new_value_for_existing_stake_pool() {
        let stake_pool_id = new_stake_pool_id();

        let mut leaders_participation_record = create_log(vec![(stake_pool_id.clone(), 1)]);

        assert_eq!(
            InsertError::EntryExists,
            leaders_participation_record
                .set_for(stake_pool_id.clone(), 10)
                .err()
                .unwrap()
        );

        assert_eq!(
            1,
            get_count(&leaders_participation_record, &stake_pool_id).unwrap(),
            "wrong count for new stake pool"
        );

        verify_participants_count(&leaders_participation_record, 1);
        //correct expected value: 1
        verify_total(&leaders_participation_record, 11);
    }
```
in PR: https://github.com/input-output-hk/chain-libs/pull/223

eugene-babichenko:
Fixes #224

### DIFF
--- a/chain-impl-mockchain/src/block/leaderlog.rs
+++ b/chain-impl-mockchain/src/block/leaderlog.rs
@@ -39,8 +39,8 @@ impl LeadersParticipationRecord {
     /// if the value already exists, then it returns an insert error.
     /// This should only be used related to the iterator construction,
     pub fn set_for(&mut self, pool: PoolId, v: u32) -> Result<(), InsertError> {
-        self.total += v;
         self.log = self.log.insert(pool, v)?;
+        self.total += v;
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


dkijania:
@eugene-babichenko can you also fix the unit tests which relays on old expected values? They are marked, like this one: 

https://github.com/input-output-hk/chain-libs/blob/master/chain-impl-mockchain/src/block/leaderlog.rs#L149

PR #234, opened 2020-01-02
update dependencies

vincenthz:
* update ed25519-dalek to pre3
* drop old rand stuff for rand-core 0.5
* update quickcheck to 0.9


NicolasDP:
the module `chain-impl-mockchain` does not compile with `cargo build --all-features`

PR #238, opened 2020-01-10
benchmarks for sqlite storage

eugene-babichenko:


eugene-babichenko:
BTW those are results on my machine (change is in comparison to using in-memory index from the new storage module):

```
put_block               time:   [241.33 us 255.75 us 272.66 us]                       
                        change: [+99.724% +110.41% +121.45%]

get_block               time:   [14.755 us 14.776 us 14.802 us]                       
                        change: [+30.425% +32.269% +33.734%] 
```

chain-storage-sqlite-old/benches/storage.rs
```diff
@@ -0,0 +1,56 @@
+#[macro_use]
+extern crate criterion;
+extern crate tempfile;
```
NicolasDP: do you still need to include `extern crate` ? This project is not edition 2018?

chain-storage-sqlite-old/Cargo.toml
```diff
@@ -21,5 +21,11 @@ version = "0.16.0"
 features = ["bundled"]
 
 [dev-dependencies]
+criterion = "0.3.0"
```
NicolasDP: can these new dependencies be included only when `cargo bench` is launch? Just adding less deps that are not use when running the bench?

chain-storage-sqlite-old/benches/storage.rs
eugene-babichenko: Thanks for noting this, it doesn't really make sense here.

chain-storage-sqlite-old/Cargo.toml
eugene-babichenko: Hm, that's a bit tricky with the current setup. The working solution would be to:

- move `bench`-only dependencies
- mark them optional
- join them under a feature with the name like `with-bench`
- require this feature for `bench` builds

This *should* be fine, but requires activating features with `cargo` flags which don't really work with cargo workspaces right now.

eugene-babichenko: Apparently, running that command from within `chain-storage-sqlite-old` works just fine. The workspace is still recognized correctly and the top-level `target` directory is used. That's just a bit inconvenient.

vincenthz:
do you have the number for the in-memory `put_block` and `get_block` ? I have trouble parsing the change properly.

PR #239, opened 2020-01-10
require explicit sqlite connections creation

eugene-babichenko:
This is required for more efficient connection pool usage. Now users are
required to explicitly create database connections which are then reused
for database operations instead of getting them from the connection pool
for each operation.

Breaking change: SQLiteBlockStoreConnection implements BlockStore
instead of SQLiteBlockStore.

### DIFF
--- a/chain-storage-sqlite-old/src/lib.rs
+++ b/chain-storage-sqlite-old/src/lib.rs
@@ -19,6 +19,10 @@ pub struct SQLiteBlockStore {
     persistent_connection: Option<Connection>,
 }
 
+// persistent_connection does not implement Sync but is never actually used
+// which makes it safe to be shared
+unsafe impl Sync for SQLiteBlockStore {}
+
 pub struct SQLiteBlockStoreConnection<B>
 where
     B: Block,
--- a/chain-storage-sqlite-old/src/lib.rs
+++ b/chain-storage-sqlite-old/src/lib.rs
@@ -3,48 +3,46 @@ use chain_storage::{
     error::Error,
     store::{BackLink, BlockInfo, BlockStore},
 };
-use r2d2_sqlite::SqliteConnectionManager;
-use rusqlite::types::Value;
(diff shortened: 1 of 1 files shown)


chain-storage-sqlite-old/src/lib.rs
```diff
@@ -15,6 +14,14 @@ where
     dummy: std::marker::PhantomData<B>,
 }
 
+pub struct SQLiteBlockStoreConnection<B>
+where
+    B: Block,
```
NicolasDP: do we need this `Block` trait constraint?

eugene-babichenko: see https://github.com/input-output-hk/chain-libs/pull/240#discussion_r365939036

chain-storage-sqlite-old/src/lib.rs
```diff
@@ -3,16 +3,31 @@ use chain_storage::{
...
+use rusqlite::{types::Value, Connection, OpenFlags};
+use std::path::{Path, PathBuf};
+
+enum StoreType {
+    Memory,
+    File(PathBuf),
+}
 
-#[derive(Clone)]
 pub struct SQLiteBlockStore<B>
```
mzabaluev: I wonder if the generic parameter can be removed from this type, as only the connection and the `connect` method really require it. This could necessitate turbofishing on `connect`, but otherwise would declutter the code.

chain-storage-sqlite-old/src/lib.rs
eugene-babichenko: Well, now I see that we can actually remove the constraint from the structure and put constraints on the functions which actually use the `Block` trait.

mzabaluev: It may be better for consistency that a connection works with the same block type. Then again, the same can be said about `SQLiteBlockStore`...

PR #240, opened 2020-01-13
Prepare to remove the BlockStore trait

eugene-babichenko:
Duplicate the required functionality of `chain-storage` in `chain-storage-sqlite-old`. After merging this and integrating the in-memory index from `chain-storage-sqlite`, we can safely remove `chain-storage`.

### DIFF
--- a/chain-storage-sqlite-old/Cargo.toml
+++ b/chain-storage-sqlite-old/Cargo.toml
@@ -7,7 +7,6 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 chain-core = { path = "../chain-core" }
-chain-storage = { path = "../chain-storage" }
 r2d2 = "0.8"
 r2d2_sqlite = "0.8"
 rusqlite = { version = "0.16.0", features = ["bundled", "unlock_notify"] }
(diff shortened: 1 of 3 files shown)


chain-storage-sqlite-old/Cargo.toml
```diff
@@ -12,14 +12,13 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 chain-core = { path = "../chain-core" }
```
NicolasDP: + 10 if you can remove that dependency too

eugene-babichenko: I don't think it's really possible right now without introducing more breaking changes. With little modification I can stop using `Block` and `BlockId`, but the current implementation relies at least on `Serialize`/`Deserialize` traits from `chain-core`. I think it would be better to work on removing this dependency later in a separate task.

vincenthz: https://github.com/input-output-hk/chain-libs/pull/278

eugene-babichenko: I suggest to integrate it in a separate task. Replacing `Block` and `BlockId` traits involves a lot of additional refactoring on the `jormungandr` side.

PR #241, opened 2020-01-13
suppress static_secret_key warning

eugene-babichenko:


eugene-babichenko:
It is used in some `Arbitrary` implementation that is used only for tests. I guess that when a function in a library is not in its public interface and is not used by any public function it is considered unused. May be a better idea to mark such functions and implementations of traits like `Arbitrary` with `#[cfg(test)]`.

NicolasDP:
yes, `cfg(test)` would make more sense then

vincenthz:
was already on master (pushed directly), so just merging the dangling PR now

PR #245, opened 2020-01-24
fix sqlite DatabaseLocked

eugene-babichenko:
fix https://github.com/input-output-hk/jormungandr/issues/1638

### DIFF
--- a/chain-storage-sqlite-old/src/lib.rs
+++ b/chain-storage-sqlite-old/src/lib.rs
@@ -260,26 +260,35 @@ mod tests {
 
     const SIMULTANEOUS_READ_WRITE_ITERS: usize = 50;
 
+    #[test]
     fn put_get() {
-        let mut store = SQLiteBlockStore::<TestBlock>::memory();
+        let mut store =
+            SQLiteBlockStore::<TestBlock>::file("file:test_put_get?mode=memory&cache=shared");
         chain_storage::store::testing::test_put_get(&mut store);
     }
 
+    #[test]
     fn nth_ancestor() {
         let mut rng = OsRng;
-        let mut store = SQLiteBlockStore::<TestBlock>::memory();
+        let mut store =
+            SQLiteBlockStore::<TestBlock>::file("file:test_nth_ancestor?mode=memory&cache=shared");
         chain_storage::store::testing::test_nth_ancestor(&mut rng, &mut store);
     }
 
+    #[test]
     fn iterate_range() {
         let mut rng = OsRng;
-        let mut store = SQLiteBlockStore::<TestBlock>::memory();
+        let mut store =
+            SQLiteBlockStore::<TestBlock>::file("file:test_iterate_range?mode=memory&cache=shared");
         chain_storage::store::testing::test_iterate_range(&mut rng, &mut store);
     }
 
+    #[test]
     fn simultaneous_read_write() {
         let mut rng = OsRng;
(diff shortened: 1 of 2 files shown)


chain-storage-sqlite-old/src/lib.rs
```diff
@@ -69,14 +71,12 @@ where
...
-        /*
         connection
-            .execute("pragma synchronous = off", rusqlite::NO_PARAMS)
+            .execute_batch("pragma journal_mode = WAL")
             .unwrap();
-        */
 
         connection
-            .execute_batch("pragma journal_mode = WAL")
+            .execute_batch("pragma read_uncommitted = true")
```
mzabaluev: - Why does this need to be done twice?
- Can this result in inconsistent reads when there is an intervening write, or are our append-only records immune to this kind of problem?

chain-storage-sqlite-old/src/lib.rs
```diff
@@ -292,5 +334,6 @@ mod tests {
         put_get();
         nth_ancestor();
         iterate_range();
+        simultaneous_read_write()
```
mzabaluev: Why must these functions be called from single `#[test]` entry point, rather than each one annotated with `#[test]`?

chain-storage-sqlite-old/src/lib.rs
eugene-babichenko: 1. My bad, will remove the second one.
2. Yes, in theory something can go wrong. But since we don't alter or remove records, everything should be fine.

chain-storage-sqlite-old/src/lib.rs
eugene-babichenko: There are two things:

- cargo runs tests in multiple threads
- sqlite manages shared cache somewhere in the background and doesn't expose any context to a user, there is only one "global context"

Because of that concurrent tests are not isolated. There are three possible ways to deal with that:

- Limit the number of threads cargo uses to run tests. This is very inconvenient when running tests for the whole repo and needs additional documentation and attention from the user who wants to run tests.
- Put all tests under one function to prevent them from being run in parallel. This is what I did.
- Create test databases with URLs like `file:{db_name}?mode=memory&cache=shared` so that there are different in-memory databases. Actually, this may be a better solution.

chain-storage-sqlite-old/src/lib.rs
mzabaluev: Please leave a comment noting this assumption near the pragma query.

I guess some future trouble may be introduced with queries that select blocks in the current epoch, but with a where that does not address specific blocks.

Is there no way to use the [unlock_notify API](https://www.sqlite.org/unlock_notify.html)? That could be a subject for a future enhancement PR, though.

chain-storage-sqlite-old/src/lib.rs
mzabaluev: I see. We can continue like this, while documenting the reason in comments.
But to improve test granularity it's better to use separate databases.

chain-storage-sqlite-old/src/lib.rs
eugene-babichenko: Not sure about that. This may require adding changes to `rusqlite`.

eugene-babichenko: Well, everything was way more easier. `rusqlite` already has support for `unlock_notify` and it is working fine.

PR #248, opened 2020-01-29
add a short header description capability header.description()

vincenthz:
print in display mode 8 characters of hash, the height in hex, and the epoch.slot

### DIFF
--- a/chain-impl-mockchain/src/header/header.rs
+++ b/chain-impl-mockchain/src/header/header.rs
@@ -63,6 +63,36 @@ impl HeaderGenesisPraos {
     }
 }
 
+/// Header description
+#[derive(Clone)]
+pub struct HeaderDesc {
+    pub id: HeaderId,
+    pub date: BlockDate,
+    pub height: ChainLength,
+}
+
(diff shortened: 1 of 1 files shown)


chain-impl-mockchain/src/header/header.rs
```diff
@@ -63,6 +63,36 @@ impl HeaderGenesisPraos {
...
+    pub height: ChainLength,
+}
+
+impl fmt::Debug for HeaderDesc {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "{}-{:08x}-{}.{}",
+            self.id, self.height.0, self.date.epoch, self.date.slot_id,
+        )
```
mzabaluev: Why not derive `Debug` in the conventional way? It's `Display` that should work as the compact human-readable representation.

vincenthz: I think this could be seen as a primitive which make sense to have a more compact impl than the normal `Debug`. Also it's mostly "historical" is there as the previous debug implementation before committing was shorting the id to 16 characters

PR #249, opened 2020-01-29
enhance error report on leader proof verifications

rinor:
[x] separate error for VrfNonce empty
  [ ] enhance VrfNonce empty with additional data
  [ ] enhance InvalidLeaderProof with cause


vincenthz:
This is a good start, you just need to add some parameter to VrfInvalidProof like the stake threshold and the value that has been generated

PR #251, opened 2020-02-03
Add btree package

enzoc4:
# BTree package

Add BTree package. It's meant to have the primitives of a database index (or at least a key-value store).

This PR adds only the insertion (and search) implementation. A later PR would add the delete implementation, but It's left out from here because this is already a big amount of code (although a big part are tests/benchmark/example, but if it is too much I may try splitting this even more) and it needs a refactor/some documentation first.

This adds also a **design.md** with the ideas behind the current design, and it would be good to have feedback on that. 

A further PR should add more raw storage primitives. In particular, most of the focus here is on the index, but this adds an almost copy of the `Appender` in the `hdb-appender` branch (but with a few bugfixes) as a submodule.


vincenthz:
Compilation failed on windows due to some permission error: https://github.com/input-output-hk/chain-libs/pull/251/checks?check_run_id=423947367

enzoc4:
Fixed, it was some bad file handling for tests.

PR #281, opened 2020-02-11
network: protocol method ExchangeGossip

mzabaluev:
Add network protocol method `ExchangeGossip` to facilitate one-off gossip exchanges.
This can be used during bootstrap to get the initial gossip from a trusted peer and then disconnect, offloading the trusted peers for serving other nodes.

### DIFF
--- a/network-grpc/proto/node.proto
+++ b/network-grpc/proto/node.proto
@@ -111,6 +111,9 @@ service Node {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
 
+  // A one-off gossip exchange.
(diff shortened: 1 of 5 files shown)


network-grpc/proto/node.proto
```diff
@@ -111,6 +111,9 @@ service Node {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
 
+  // A one-off gossip exchange.
+  rpc ExchangeGossip(Gossip) returns (Gossip);
```
NicolasDP: Hmm no this was not the direction we wanted to go. Here we don't even know about gossips but about other peers. @vincenthz was looking into this too I believe.

PR #283, opened 2020-02-13
implement get_block_by_chain_height

eugene-babichenko:


vincenthz:
The PR contains a commit from another PR, I don't think it matter to merge it since git will handle it, but at the same time probably a good idea to remove before merging

eugene-babichenko:
oops, seems like it was based on some tag instead of master

PR #284, opened 2020-02-17
Move heap allocations to mmap slices

enzoc4:
# Use mmaped region instead of heap allocations

## Motivation

Before this PR, the storage is using mmap just for reading and writing from/to heap allocated pages. This means almost the only benefit of using a memory mapped file is the os cache.

## Summary

The actual important changes are the ones in *storage.rs* and *pages.rs*, but changing the location of the ownership of the data required more changes in the higher level transaction module. 

- Avoid copying from the mmap and just use indexes to take slices of page size.
- Separate the *write* from the *extend* operation so we can synchronize in order to avoid dangling pointers when unmapping and remapping.
- Make the the insert transaction algorithm track id's instead of box allocations.

The previous `Page` abstraction (that was basically a `Box`) is replaced for a `PageHandle`, that contains a slice to the memory mapped region. The main issue with this approach, is that extending the mmaped region requires an `unmap` operation first, so we must take care of not allowing reads nor writes to proceed when this happens. At this point, a writer may need to extend the storage, but won't be able to proceed as long as a reader is holding a read lock, this should anyway be rare as we always allocate more space than we need (and there is only one writer at a time, so that part is not problematic).

### Disclaimer
There is a bit of noise in the diffs because I split the the version module into two, in general, there are no changes for the tree algorithms, but I couldn't keep everything localized to the storage (particularly the copy on write part needed to be changed, some error propagation, first page initialization)

### DIFF
--- a/btree/src/storage.rs
+++ b/btree/src/storage.rs
@@ -1,25 +1,15 @@
 use memmap::MmapMut;
-use std::convert::TryFrom;
+use std::cell::UnsafeCell;
+
 use std::convert::TryInto;
 use std::fs::File;
 use std::io;
 use std::mem::ManuallyDrop;
-
-pub trait Storage<'a> {
-    type Output;
-
-    fn get(&'a self, location: u64, count: u64) -> Result<Self::Output, io::Error>
-    where
-        Self::Output: AsRef<[u8]> + 'a;
-
-    fn put(&mut self, location: u64, bytes: impl AsRef<[u8]>) -> Result<(), io::Error>;
-
-    fn sync(&self) -> Result<(), io::Error>;
-}
+use std::sync::atomic::{AtomicU64, Ordering};
 
 pub struct MmapStorage {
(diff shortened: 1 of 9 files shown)


btree/src/storage.rs
```diff
@@ -35,62 +25,74 @@ impl MmapStorage {
...
     }
 
-    fn sync(&self) -> Result<(), io::Error> {
-        self.mmap.flush()
+    pub fn sync(&self) -> Result<(), io::Error> {
+        // there is nothing really unsafe here, we need the block only because of unsafe cell (at least nothing that is not already present in the memmap api)
+        unsafe { &*self.mmap.get() }.flush()
+    }
+
+    fn next_power_of_two(n: u64) -> u64 {
```
vincenthz: This is just:

https://doc.rust-lang.org/1.0.0/core/primitive.u64.html#method.next_power_of_two

PR #290, opened 2020-02-19
Bump rusqlite version

### LINKED ISSUE #288: Update rusqlite to 0.21, consolidate chain-storage-sqlite crates
The `rusqlite` version we use is quite old and has some problems for Windows builds: @danielSanchezQ reported it prevents fixing https://github.com/input-output-hk/jormungandr/issues/1699, while my standalone experiment with 0.21 and "bundled" worked well with `crt-static`.

Update the `rusqlite` dependency to the latest version (keep "bundled" on or optionally-default for the time being). There is a use of `r2d2_sqlite` backend in `chain-storage-sqlite`, which is ironically the more outdated crate, so that needs to be dropped and `chain-storage-sqlite-old` renamed as the only surviving storage implementation.

eugene-babichenko:
partially resolve #288

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,6 @@ members = [
     "chain-time",
     "chain-crypto",
     "chain-storage",
-    "chain-storage-sqlite",
     "chain-storage-sqlite-old",
     "chain-impl-mockchain",
     "cardano-legacy-address",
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -7,7 +7,6 @@ members = [
     "chain-time",
     "chain-crypto",
     "chain-storage",
-    "chain-storage-sqlite",
     "chain-storage-sqlite-old",
```
mzabaluev: Renaming `-old` needs more work?

PR #292, opened 2020-02-21
Mmaped flatfile

enzoc4:
Replace the seek and write/read approach used for storing the contents of the store for a similar abstraction based on a memory mapped file (before it was only used for the index/tree pages). This also moves the synchronization of the file to the appender, which means that now `BTreeStore` can be shared without external synchronization (because `insert` takes `&self` and not `&mut self`). Although maybe this is not a good thing?.

* Note *: This still needs a few more tests, particularly because most tree tests only test the index, so this kind of gets left out.

### DIFF
--- a/btree/src/flatfile.rs
+++ b/btree/src/flatfile.rs
@@ -1,5 +1,7 @@
 use crate::storage::MmapStorage;
 use byteorder::{ByteOrder, LittleEndian};
+use parking_lot::lock_api;
+use parking_lot::{RwLock, RwLockReadGuard, RwLockUpgradableReadGuard, RwLockWriteGuard};
 use std::io::{self, Error, ErrorKind, Read, Seek, SeekFrom, Write};
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::{fs, path};
@@ -119,7 +121,7 @@ impl From<u64> for Pos {
 }
 
 pub struct MmapedAppendOnlyFile {
-    storage: MmapStorage,
+    storage: RwLock<MmapStorage>,
     next_pos: AtomicU64,
 }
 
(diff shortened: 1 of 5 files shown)


btree/src/flatfile.rs
```diff
@@ -34,71 +41,103 @@ impl Appender {
...
+
+        let region_len = szbuf.len() as u64 + buf.len() as u64;
+
+        let mmaped_region = unsafe {
+            match storage.get_mut(next_pos, region_len) {
+                Ok(slice) => slice,
+                Err(including) => {
+                    {
+                        let mut new_guard = RwLockUpgradableReadGuard::upgrade(storage);
+                        new_guard.extend(including)?;
```
enzoc4: I've been thinking that may be it's just easier to just map a huge area from the start (basically enough to cover `MAX_POS_OFFSET`) and never resize. Was that the idea behind that constraint @vincenthz ? Although the only real advantage is that we could return pointers to the mapped are without the need to invalidate pointers/wait for readers to end before extending.

vincenthz: you can only do that if you prereserve your file on disk, and it's probably unreasonable to pre-reserve a huge chunk (a terabyte), but it's not unreasonable to work on extend of around 128mb ... i.e. you can guarantee to mmap in a stable way 128mb at a time.

btree/src/flatfile.rs
```diff
@@ -34,71 +41,103 @@ impl Appender {
...
 
     /// Get the blob stored at position @pos
-    pub fn get_at(&mut self, pos: Pos) -> Result<Box<[u8]>, io::Error> {
-        self.rhandle.seek(SeekFrom::Start(pos.0))?;
+    pub fn get_at(&self, pos: Pos) -> Result<Box<[u8]>, io::Error> {
+        // TODO: this will panic if position if out of range, we may want to handle that?
+        let storage = self.storage.read();
+        let szbuf = unsafe { storage.get(pos.into(), 4) };
+
+        let len = LittleEndian::read_u32(&szbuf);
```
vincenthz: std now have `u32::from_le_bytes()` with a `try_into().unwrap()` to copy from slice to array

btree/src/flatfile.rs
enzoc4: isn't 128mb too conservative? right now it's growing exponentially, because I'm using powers of two, but I thought about setting a threshold and expand linearly from there. What I'm mostly wondering though, there is actually no way of working directly on the mmaped area, because each time you expand you invalidate pointers, so cloning is the only option, right?

vincenthz: my idea of working on this is a pagetable of X size. the pagetable contains where a range of X size is mapped by holding its mmaped pointer (or null if not currently mapped).

Each time the file is grown, the file is extended by X through an `fallocate` or `ftruncate` and the pagetable is extended by 1.

enzoc4: Oh, glad I asked then. So your idea was not to create a big contiguous mmap covering the whole file, but have something more granular and have something like a translation unit? I guess then you can map pages on demand, the only problem is deciding when to unmap them (if we actually do).

I'm not sure why didn't I consider doing something like that, although it is a bit more complex than what's done now.

As an aside note, that would probably make deleting things more easier.

PR #293, opened 2020-02-25
allow to set storage busy timeout

eugene-babichenko:


chain-storage-sqlite-old/src/lib.rs
```diff
@@ -181,7 +190,7 @@ where
 
         let tx = self
             .inner
-            .transaction()
+            .transaction_with_behavior(TransactionBehavior::Immediate)
```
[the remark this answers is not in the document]
eugene-babichenko: Not only that. `Immediate` is required for a transaction to immediately acquire the write lock. With the default `Deferred` behavior the lock is updated from read to write in the middle of a transaction. This prevents sqlite from using `busy_handler` and `SQLITE_BUSY` will be thrown regardless of the busy timeout. But if the lock if for write from the very beginning it works just fine.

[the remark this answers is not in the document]
mzabaluev: Exactly, if the lock is deferred, this leaves the possibility of a conflict with a concurrent write, but this way it's all atomic and conflict-free, so only an abnormally slow write transaction could make other operations time out.

chain-storage-sqlite-old/src/lib.rs
```diff
@@ -460,18 +450,72 @@ where
...
-        block_hash: &B::Id,
-        distance: u64,
-    ) -> Result<BlockInfo<B::Id>, Error> {
-        for_path_to_nth_ancestor_internal::<B, _>(connection, block_hash, distance, |_| {})
+fn get_block_info_internal<B>(
+    connection: &Connection,
+    block_hash: &B::Id,
+) -> Result<BlockInfo<B::Id>, Error>
+where
+    B: Block,
```
mzabaluev: Try making these generic over the hash type, then you probably won't need turbofishing at the use sites.

mzabaluev: Try making these generic over the hash type instead of an `impl Block`, then you probably won't need turbofishing at the use sites.

mzabaluev: Try making these generic over the hash type instead of an `impl Block`, then you probably won't need turbofishing at the use sites.

PR #299, opened 2020-03-10
Use the customized container image in CircleCI

mzabaluev:
Make use of our [tailored build images](https://github.com/input-output-hk/rust-ci-docker) in CircleCI builds.

Running `cargo audit` without extra setup is now possible, but temporarily disabled pending #294 that will remove the known vulnerability in `prost`.


NicolasDP:
can we have the audit job back with this?

mzabaluev:
@NicolasDP I have rebased this and restored the audit job.

mzabaluev:
This is too old, we can run the remaining jobs with the off-the-shelf Rust container images, and we are close to ditching CircleCI altogether for its weird failures.

PR #301, opened 2020-03-16
Add leaf and internal delete

enzoc4:
# Start integrating delete algorithm

This PR primarily adds both `InternalNode` and `LeafNode` delete and rebalance algorithms. They are not usable by themselves, because this doesn't add the top level delete algorithm, but as this was quite long itself I thought it was better to split it.

There are a few things to improve still, for example, the test code is a bit verbose, there is probably some way of making it more succinct. Currently `rebalance` doesn't actually do the rebalancing, instead it determines which is the required rebalancing technique and returns a new object which takes the additional parameters. This is done in order to just clone/shadow the siblings when they are needed.

There is also an additional change in this PR, which is adding internal mutability to the transaction object (the one that keeps account of current cloned/not cloned pages). The reason for this is that the rebalancing algorithm may require mutating three nodes at a time, and this is the simplest way, I could split this if preferred.

### DIFF
--- a/btree/benches/benchmark.rs
+++ b/btree/benches/benchmark.rs
@@ -30,7 +30,7 @@ fn single_key_insertion(c: &mut Criterion) {
     let key_size = std::mem::size_of::<U64Key>();
     let page_size = 4096;
 
-    let mut tree: BTreeStore<U64Key> =
+    let tree: BTreeStore<U64Key> =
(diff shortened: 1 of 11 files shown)


vincenthz:
I'll look at this once this is rebased to the latest

enzoc4:
it's rebased now, I'm not quite sure how readable it is without actually knowing how the rebalance algorithm works. Let me know if it is too hard to read, probably some things can be improved (I'm thinking the tests, but some other things also)

PR #303, opened 2020-03-18
explicitly use an immediate transaction in put_tag

eugene-babichenko:
This is required for proper synchronization with put_block.
Otherwise put_tag may fail with SQLITE_BUSY error.

Fix a part of https://github.com/input-output-hk/jormungandr/issues/1934


eugene-babichenko:
currently rebased onto the tag used in the latest jormungandr master

PR #304, opened 2020-03-18
Add delete algorithm

enzoc4:
As #301 adds the delete and rebalance algorithms for the node types, this adds the actual functions to do the delete. This includes traversing the tree to find the leaf where the key to delete is found, clone the whole path, and traverse back (by using a stack/backtrack) calling rebalance (and cloning/shadowing siblings) as necessary. 

Caveats:
- There is still a bit of code duplication between the newly added `DeleteBacktrack` and the previous `InsertBacktrack`, as they act similarly but not quite.
- The chaining of borrows in `TransactionManager` -> `WriteTransaction` -> `DeleteBacktrack` -> `DeleteNextElement` made lifetime annotations a bit verbose, it's not really complex, but there may be some way to write that better (each one borrows from the previous one and exposes the necessary methods at each stage, this simplifies borrowing, but I'm not sure how idiomatic it is)

### DIFF
--- a/btree/benches/benchmark.rs
+++ b/btree/benches/benchmark.rs
@@ -24,6 +24,9 @@ impl<'a> Storeable<'a> for U64Key {
     fn read(buf: &'a [u8]) -> Result<Self::Output, Self::Error> {
         Ok(U64Key(LittleEndian::read_u64(buf)))
     }
+    fn as_output(self) -> Self {
+        self
+    }
 }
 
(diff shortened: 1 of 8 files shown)


enzoc4:
Forgot to mention, this only removes entries from the tree, although that has the effect of deleting the entry, the space is never reclaimed, as removing from the append only file would require more thought

PR #306, opened 2020-03-21
[Tests] stake distribution test and leadership test

dkijania:


chain-impl-mockchain/src/leadership/genesis/mod.rs
```diff
@@ -662,4 +663,34 @@ mod tests {
...
+        let stake_pool = StakePoolBuilder::new().build();
+        ledger
+            .delegation()
+            .register_stake_pool(stake_pool.info())
+            .expect("cannot register stake pool");
+        let selection = LeadershipData::new(date.epoch, &ledger);
+        let rng = rand_core::OsRng;
+        let sk = &SecretKey::generate(rng);
+        let header = HeaderBuilderNew::new(BlockVersion::Ed25519Signed, &Contents::empty())
+            .set_parent(&HeaderId::zero_hash(), ledger.chain_length().next())
```
vincenthz: `.increase()` instead of `.next()` so that `property::ChainLength` is not needed.

chain-impl-mockchain/src/stake/stake.rs
```diff
@@ -1,8 +1,9 @@
 use crate::value::Value;
 use std::ops::{Add, AddAssign};
 
-#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub struct Stake(u64);
+#[cfg_attr(feature = "generic-serialization", derive(serde_derive::Serialize))]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Stake(pub u64);
```
vincenthz: why public, why as_ref<u64> and why serializable ?

PR #307, opened 2020-03-23
Ledger recovery

danielSanchezQ:
Added all necessary methods so we can save the ledger and bring it back:

- Implemented serializing and deserializing methods for needed types
- Implemented tests for each method
- Implemented an owning version of the ledger::iter::Entry type
- Implemented `Serialize`/`Deserialize` trait for Ledger (deserialize uses from_iter trait)
- Implemented deserialize for some types that already had a serialize method
- Implemented `Arbitrary` for some types that could use `quickcheck`


danielSanchezQ:
Please use squash when merging this PR

PR #313, opened 2020-03-30
split contiguous mmaped file in pages

enzoc4:
Based on https://github.com/input-output-hk/chain-libs/pull/292#discussion_r394167482

# About
Extend mmaped region by pages instead of unmapping and remapping whole file. Using a hashmap to translate absolute positions to (page, offset)

## Index (tree nodes/pages)

Now write transactions never need to wait for read transactions to end. This was needed previously when extending, as the read transactions had pointers that couldn't be invalidated. This means now readers and writers can't block each other.

## Appender
 This allows returning slices to the maped area, removing the need for cloning on reads.
### Possible problem
 As blobs are of arbitrary size, unlike tree nodes, it's possible they may not fit in the remaining space of a (the current) page, in that case they are written in the next page and the skipped space remains unused. This may be problematic is the page size is too small with respect to the blob sizes, but it seems to be the simpler solution.

### DIFF
--- a/btree/src/storage.rs
+++ b/btree/src/storage.rs
@@ -10,81 +10,34 @@ use std::mem::ManuallyDrop;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Mutex;
 
+pub const DEFAULT_PAGE_SIZE: u64 = (1 << 20) * 128; // 128mb
+
 pub struct MmapStorage {
     pages: ManuallyDrop<PageTable>,
     file_len: AtomicU64,
     allocated_size: AtomicU64,
     file: *mut File,
+    page_size: u64,
 }
 
 type PageId = u64;
-const PAGE_SIZE: u64 = (1 << 20) * 128; // 128mb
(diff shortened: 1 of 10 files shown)


btree/src/storage.rs
```diff
@@ -1,50 +1,96 @@
...
     file: *mut File,
+    page_size: u64,
+}
+
+type PageId = u64;
+
+struct PageTable {
+    // XXX: A vector may be a decent choice too
+    // It may be possible to reduce locking here
+    lookup: RwLock<HashMap<PageId, Page>>,
```
enzoc4, line 30: It may be possible to use [sled's pagetable](https://crates.io/crates/pagetable) too. Which sees to be (at first glance) a radix-tree linux style. I'm not sure of depending on `sled_sync` though. But it may be worth looking into.

vincenthz: down the line probably fine, it doesn't seems too sled specific, but for now also fine without it

PR #315, opened 2020-04-02
Add btree update operation

enzoc4:
Branched from #304 mostly because to avoid a messy rebase later.

## Add update value operation for the btree

Although tihs was already possible with a delete followed by an update, the implementation is really simple and avoids unnecessary rebalances.


NicolasDP:
@enzoc4 can you rebase this PR when you have time?

enzoc4:
> @enzoc4 can you rebase this PR when you have time?

done @NicolasDP

PR #338, opened 2020-04-21
[Tests] Small utils for changes in libs

dkijania:
Updates StakePoolBuilder and Wallet structs


NicolasDP:
The serial number has been removed from the exposed certificate.

dkijania:
@NicolasDP but `serial` it's still in `PoolRegistration` struct so it's still a valid part of certificate.  I need to take it into consideration while working on pool update tests. Jcli still sets it to 0 when new stake registration creation. There are still some valid tests which issue stake Pool update changing only `serial`. I can update tests and remove it, but then they will fail

NicolasDP:
it is still in there but it is not allow to set it to the users, It is kept for compatibility with the ITN. but it is to be remove once we can drop the support for the ITN.

NicolasDP:
just put `0` all the time for the serial

PR #339, opened 2020-04-21
Added all-features flag for all build targets

dkijania:


dkijania:
Due to issues with enabling particular features from workspace root (https://github.com/rust-lang/cargo/issues/5015) i've added new explicit build for chain-impl-mockchain

NicolasDP:
can we make it a github action instead?

.circleci/config.yml
```diff
@@ -103,6 +103,29 @@ jobs:
           environment:
             RUST_BACKTRACE: 1
           command: cargo test --verbose --frozen
+    
+  test_mockchain_all_features:
+    docker:
+      - image: rustlang/rust:nightly
```
mzabaluev, line 109: Are there unstable features so that you must use nightly?

.circleci/config.yml
```diff
@@ -103,6 +103,29 @@ jobs:
           environment:
             RUST_BACKTRACE: 1
           command: cargo test --verbose --frozen
+    
+  test_mockchain_all_features:
```
mzabaluev, line 107: This may become useful for testing other crates with non-default features, so I'd name the job `test_all_features` with intent to add more per-crate build/test steps as needed later.

.circleci/config.yml
dkijania: nope :) removing

PR #347, opened 2020-04-23
Ledger recovery vote plan support

danielSanchezQ:


chain-impl-mockchain/src/certificate/vote_plan.rs
```diff
@@ -83,6 +83,10 @@ impl VoteOptions {
             num_choices: num_choices & Self::NUM_CHOICES_MAX,
         }
     }
+
+    pub fn choices(&self) -> u8 {
```
NicolasDP, line 87: `into_byte()` ?

NicolasDP: this is going to be extended (or not) in the future, so I think it would be better to be straight forward about it

chain-impl-mockchain/src/ledger/recovery.rs
```diff
@@ -50,12 +50,12 @@ use crate::fragment::{ConfigParams, FragmentId};
 use crate::header::{ChainLength, HeaderId};
 use crate::key::{serialize_public_key, BftLeaderId};
 use crate::ledger::{Globals, Ledger, LedgerStaticParameters};
-use crate::legacy;
 use crate::multisig::{DeclElement, Declaration};
 use crate::stake::{PoolLastRewards, PoolState};
 use crate::transaction::Output;
 use crate::update::{UpdateProposal, UpdateProposalId, UpdateProposalState, UpdateVoterId};
 use crate::value::Value;
+use crate::{certificate, legacy};
```
NicolasDP, line 58: unused import
```suggestion
use crate::legacy;
```

chain-impl-mockchain/src/certificate/vote_plan.rs
danielSanchezQ: I'll use `as_byte` since `into` methods are usually supposed to consume the value.

PR #350, opened 2020-04-24
chain-storage is not needed in the chain-impl-mockchain

NicolasDP:
remove a dependency from chain-impl-mockchain: `chain-storage` is not needed, only for a test.

### DIFF
--- a/chain-impl-mockchain/src/multiverse.rs
+++ b/chain-impl-mockchain/src/multiverse.rs
@@ -8,7 +8,6 @@
 
 use crate::chaintypes::{ChainLength, HeaderId};
 use crate::ledger::Ledger;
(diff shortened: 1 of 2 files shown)


chain-impl-mockchain/src/multiverse.rs
```diff
@@ -188,10 +187,11 @@ impl Multiverse<Ledger> {
     /// Get the chain state at block 'k' from memory if present;
     /// otherwise reconstruct it by reading blocks from storage and
     /// applying them to the nearest ancestor state that we do have.
+    #[cfg(test)]
```
eugene-babichenko, line 190: Can we move it from impl to `mod test`? It doesn't touch private interface.

PR #368, opened 2020-05-05
API to encode public address in subscription requests

mzabaluev:
Provide a client side way to set the public peer address in a metadata
header of subscription requests. This functionality was lost in the tokio 0.2 rework (#294).

Previously, node IDs were encoded in a different metadata header.

### DIFF
--- a/chain-network/src/grpc/convert.rs
+++ b/chain-network/src/grpc/convert.rs
@@ -93,7 +93,7 @@ pub(super) fn decode_peer(meta: &MetadataMap) -> Result<Peer, Error> {
 }
 
 pub(super) fn encode_peer(peer: &Peer, meta: &mut MetadataMap) {
-    let addr = format!("{}", peer.addr());
+    let addr = peer.addr().to_string();
     meta.insert(PEER_ADDRESS_HEADER, MetadataValue::from_str(&addr).unwrap());
 }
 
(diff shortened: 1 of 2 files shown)


chain-network/src/grpc/convert.rs
```diff
@@ -92,6 +92,11 @@ pub(super) fn decode_peer(meta: &MetadataMap) -> Result<Peer, Error> {
     Ok(address.into())
 }
 
+pub(super) fn encode_peer(peer: &Peer, meta: &mut MetadataMap) {
+    let addr = format!("{}", peer.addr());
```
NicolasDP, line 96:
```suggestion
    let addr = peer.addr().to_string();
```

chain-network/src/grpc/convert.rs
```diff
@@ -92,6 +92,11 @@ pub(super) fn decode_peer(meta: &MetadataMap) -> Result<Peer, Error> {
...
+    meta.insert(PEER_ADDRESS_HEADER, MetadataValue::from_str(&addr).unwrap());
```
NicolasDP, line 97: I would try not to have an `unwrap` here as the `addr().to_string()` may not be always ascii. Just in case, can we have a check on this?

mzabaluev: I expect the string representation of the IP address to be always ASCII. If it's not, the text header will not work and we'd have to change the metadata key anyway.

PR #369, opened 2020-05-06
made vote finish exclusive

dkijania:
Test 'vote::manager::tests::vote_plan_manager_can_vote' failed for case:

```
VotePlan { 
vote_start: BlockDate { epoch: 6, slot_id: 13 }, 
vote_end: BlockDate { epoch: 55, slot_id: 48 }, 
committee_end: BlockDate { epoch: 86, slot_id: 89 }, proposals: Proposals {....}
}
```

`BlockDate { epoch: 55, slot_id: 48 })`

expected result was that can_vote should return false, but it returns true.

Applied fix in commit for this corner case.


NicolasDP:
Good spotting this.

PR #371, opened 2020-05-06
Fall back to transport address as the peer address for server-side subscriptions

mzabaluev:
If the `peer-address` metadata value in server-side subscription requests does not provide the public address, use the remote peer address from the request.


mzabaluev:
Replaced by #374

PR #372, opened 2020-05-06
remove GH Actions reduntant runs

eugene-babichenko:
Tests are run on every push so there is no need to trigger them again
when opening a pull request.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,8 @@
-# copy of https://github.com/actions-rs/meta/blob/master/recipes/quickstart.md
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
 
 name: Continuous integration
 


mzabaluev:
OK if it's really redundant. I recall Travis checked out a pre-merged tree to build for pull request so the result might be different there.

eugene-babichenko:
@mzabaluev hm, seems you're right. I checked `actions/checkout@v1`, here is an example:

`git -c http.extraheader="AUTHORIZATION: basic ***" fetch --tags --prune --progress --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/pull/371/merge:refs/remotes/pull/371/merge`.

In that case the question is whether there is a need to run the `push` trigger for **every** branch even if it is not a part of a pull request?

mzabaluev:
If `pull_request` rebuilds on (force-)pushes after the initial PR has been created, then `push` for bare branches is redundant.

eugene-babichenko:
@mzabaluev I was experimenting in this pull request https://github.com/input-output-hk/jorup/pull/16. `pull_request` is triggered every time there is a push to this pull request.

PR #373, opened 2020-05-07
chain-core remove of property-test-api

danielSanchezQ:
The idea is to remove the usage of the property-test-api and extract it into a crate that should be used for the only purpose of testing. This will avoid some intoxication of external crates when importing chain-core.
This is the first step in refactoring the whole chain-libs to avoid the above explained.


Cargo.toml
```diff
@@ -8,6 +8,7 @@ members = [
     "chain-crypto",
     "chain-network",
     "chain-storage",
+    "chain-test-utils",
```
danielSanchezQ, line 11: Any other name feels better?

PR #375, opened 2020-05-08
Fix deps for missing new chain-test-utils

danielSanchezQ:
Until completely removal of the `feature-test-api` the new crate should be included also in non-dev deps.


mzabaluev:
Can't it be optional on 'property-test-api', but also added as a plain dependency to `dev-dependencies`?

danielSanchezQ:
@mzabaluev , I'm not sure how to do that. Just for clarify, you mean adding `chain-test-utils` as a dependency for `property-test-api`?

mzabaluev:
I've fixed it, seems to work with jormungandr.

PR #376, opened 2020-05-11
Remove test-api-feature from chain-time

danielSanchezQ:
Another module with support in the new `chain-test-utils`.


chain-test-utils/src/time/mod.rs
```diff
@@ -0,0 +1,136 @@
+use chain_time::{Epoch, Slot, TimeEra};
+use smoke::{Generator, R};
+
+// Generate an `Epoch` from a generator function
+fn _generate_epoch<GenF>(mut gen: GenF) -> Epoch
```
mzabaluev, line 5: Maybe more conventional `generate_epoch_with`?

chain-test-utils/src/time/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+where
+    GenF: FnMut() -> u32,
+{
+    Epoch(gen())
+}
+
+// Generate an `Slot` from a generator function
+fn _generate_slot<GenF>(mut gen: GenF) -> Slot
```
mzabaluev, line 13: `generate_slot_with`, as above

chain-test-utils/src/time/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+where
+    GenF: FnMut() -> u64,
+{
+    gen().into()
+}
+
+// Generate an `Epoch` given a `smoke::R` (random generator)
```
mzabaluev, line 20: Are we now transitioning to `smoke` from `quickcheck`, which also has a framework for arbitrary, but reproducible stuff? Does it add value to use both?

I wish I was able to see what `smoke` can do from the documentation, but there's scarcely any.

chain-test-utils/src/utils/mod.rs
```diff
@@ -0,0 +1,7 @@
+pub mod seed;
+pub use smoke::R;
+
+#[allow(non_snake_case)] // Allow because R is a type name, the function is snake case anyway
+pub fn new_R_from_random_seed() -> R {
```
mzabaluev, line 5: Maybe come up with a more idiomatic name instead? Or just use the two functions combined, which seem hardly much longer to write than this one function name?

chain-test-utils/src/time/mod.rs
danielSanchezQ: Well, the thing with `quickcheck::Arbitrary` is that since it is a `trait` it should be implemented in the same crate as the target type. So, the idea here is to have a "centralize" crate with the utilities to generate our inner types. That way that crate can be simply used in the `tests` without relying on features. We can still use `quickcheck::Arbitrary` for the self crate tests since it is useful, but rely on `chain-test-utils` for whatever is not in that crate.

About smoke, it just has some trait for generating types and abstracts some of the most basic random numbers generators.

chain-test-utils/src/utils/mod.rs
danielSanchezQ: Well, it felt like I was gonna be instantiating a generator many times, and maybe we would need to change how it is done later so I wanted to have the simplest possible method to get it.
How does `new_random_generator` sounds?

mzabaluev: Good enough.

chain-test-utils/src/time/mod.rs
mzabaluev: > So, the idea here is to have a "centralize" crate with the utilities to generate our inner types.

That might prove a bit more difficult to maintain in separation from the crates defining the types.

> That way that crate can be simply used in the tests without relying on features.

What was the original problem we are trying to solve here? The "property-test-api" feature added the testing trait impls and possibly other test utilities for testing uses, and it was to be used in approximately this way:

```toml
# For non-test binaries.
# If the whole target is meant for testing, add "property-test-api" here as well
[dependencies]
chain-core = { path = "chain-deps/chain-core" }

# For cfg(test), integration tests under tests/, etc.
[dev-dependencies]
chain-core = { path = "chain-deps/chain-core", features = "property-test-api" }
```

vincenthz: @mzabaluev the whole point of a generator based approach, is that it allow to define those generator elsewhere, since they are effectively just function/closures. A generator is just an "infinite" iterator of random value where the random engine is deterministically generated from a given seed, hence assuring reproducibility from a given `Seed`.

danielSanchezQ: I think I'm not explaining it properly, sorry.
With the `property-test-api` we are exporting the `quickcheck` dep. We are using this feature in the jormungandr tests (can't point out where exactly now). Which apparently is a misuse. 

I see your point though. We could keep the feature for testing just the `chain-deps`. (But we may have the temptation in the future to use it just because it may be easier 🍔 )

mzabaluev: So there are two things:
- we want a more flexible API than `arbitrary` to construct testing data;
- there is a concern about test facilities embedded into the library crates and conditionally enabled, as being cumbersome to use, or easy to get wrong and pull in unnecessary dependencies.

Would the latter be a matter of convention that has been followed only loosely, which may be what is causing problems?

Another benefit of testing APIs built into crates is that you don't have to break encapsulation to provide the testing facilities: a public generator defined in a conditionally compiled `testing` submodule can freely use the type's internals to generate values, while we don't want to provide equivalent freedom to construct values in public API.

danielSanchezQ: > Another benefit of testing APIs built into crates is that you don't have to break encapsulation to provide the testing facilities: a public generator defined in a conditionally compiled testing submodule can freely use the type's internals to generate values, while we don't want to provide equivalent freedom to construct values in public API.

Yes, this is something that will happen for sure.

Well, we can keep both though. And careful use them where they are needed.

danielSanchezQ:
I'm letting the `feature-test-api` as commented (bad branch name now). We can keep the testing utilities and see how to move in the future.

chain-test-utils/src/time/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// Generate an `TimeEra` given a `smoke::R` (random generator)
+pub fn generate_time_era(r: &mut R) -> TimeEra {
+    TimeEra::new(generate_slot(r), generate_epoch(r), r.num())
+}
+
+// `TimeEra` configuration, encapsulates the building boundaries for the inner data
+#[derive(Clone)]
+pub struct TimeEraGenCfg {
```
mzabaluev, line 47: Maybe a bit less abbreviated: `TimeEraGenConfig`

chain-test-utils/src/time/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+impl TimeEraGenerator {
+    pub fn new(config: Option<TimeEraGenCfg>) -> Self {
+        Self { config }
+    }
+
+    pub fn with_config(&mut self, config: TimeEraGenCfg) {
+        self.config = Some(config);
+    }
```
mzabaluev, line 77: These look a bit redundant: `new` could use the default configuration, so just be the same as the `Default` impl.

chain-test-utils/src/time/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    pub fn clear_config(&mut self) {
```
mzabaluev, line 79: Do we really need the mutability?

chain-test-utils/src/time/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+// `TimeEra` configuration, encapsulates the building boundaries for the inner data
+#[derive(Clone)]
+pub struct TimeEraGenCfg {
+    pub slot_range: (u64, u64),
+    pub epoch_range: (u32, u32),
+    pub slots_per_epoch_range: (u32, u32),
+}
+
+// Generate an `TimeEra` given a `smoke::R` (random generator) and a `TimeEraGenCfg` range limit tuple
+pub fn generate_time_era_with_config(r: &mut R, config: &TimeEraGenCfg) -> TimeEra {
```
mzabaluev, line 54: This (as a public API) also looks redundant to the composition of `TimeEraGenerator::with_config` and the `gen` method.

chain-test-utils/src/time/mod.rs
mzabaluev: The `TimeEraGenCfg` struct looks like it could provide a builder-style API, but maybe it's too much polish for the testing interface.

chain-test-utils/src/time/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+
+// Generate an `Epoch` given a `smoke::R` (random generator) and a `(u32, u32)` range limit tuple
+pub fn generate_epoch_with_range(r: &mut R, range: (u32, u32)) -> Epoch {
+    generate_epoch_with(|| r.num_range(range.0, range.1))
+}
+
+// Generate an `Slot` given a `smoke::R` (random generator)
+pub fn generate_slot(r: &mut R) -> Slot {
+    generate_slot_with(|| r.num())
+}
```
mzabaluev, line 33: Can this generate nonsensical slot values?
I'm trying to understand the purpose of non-bounded random generator methods here.

chain-test-utils/src/time/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+        r.num_range(
+            config.slots_per_epoch_range.0,
+            config.slots_per_epoch_range.1,
+        ),
+    )
+}
+
+// Generator wrapper for TimeEra generator methods
+pub struct TimeEraGenerator {
+    config: Option<TimeEraGenCfg>,
```
mzabaluev, line 67: I'd rather have a non-optional config and set range bounds to sensible defaults for a `TimeEra`.

chain-test-utils/src/time/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+impl TimeEraGenerator {
+    pub fn new(config: Option<TimeEraGenCfg>) -> Self {
+        Self { config }
+    }
```
NicolasDP, line 73: what's the point to have a `new` with an optional of something and then a `with_config` to set the config?

use a `Config` and a `ConfigBuilder` if you need it. Use `Default` too. This will make your types more ergonomic.

chain-test-utils/src/time/mod.rs
danielSanchezQ: Uhm, for modifying the struct I think we need.

chain-test-utils/src/time/mod.rs
danielSanchezQ: Well, I was thinking that we may want to change the way the generator generates values mid-way on some computation and still keep the reproducible step from the same seed. I see that `with_config` was an unfortunate name for this. It should be a `set_config`.

chain-test-utils/src/time/mod.rs
danielSanchezQ: Well, my idea was that some times you will need a one shot time build and some times you will need a bunch of generated items. It feels overkill to build a generator just to get one object when you can just call a method.

chain-test-utils/src/time/mod.rs
danielSanchezQ: Yes, it actually just makes sense if you want to test both acceptable values and not.

chain-test-utils/src/time/mod.rs
danielSanchezQ: Where can I find some documentation about what would be some range bounds for a `TimeEra`?

mzabaluev: Good question. @vincenthz ?

I would like the elementary newtype values to be domain-valid by construction, unless their domain really covers the entire underlying integer type.

danielSanchezQ:
@NicolasDP , I expanded (and fixed) the documentation. But I'm not sure if it was in the way you intended. I think I did not understand properly what you meant about the types interaction.

danielSanchezQ:
I'm closing this for now. It will be resumed in the future.

PR #381, opened 2020-05-14
GitHub CI performance improvements

mzabaluev:
Cache crates.io registry and the downloaded dependencies.
No more parallel jobs, to reuse the cache and save on the CPU time budget.


.github/workflows/main.yml
```diff
@@ -1,65 +1,62 @@
...
+        with:
+          command: check
+
+      - name: Run test suite
+        uses: actions-rs/cargo@v1
+        continue-on-error: false
+        with:
+          command: test
+
+      - name: Check clippy lints
```
danielSanchezQ, line 59: I think that by using jobs if the first one fails the others will not even be run. So maybe we can move the lints to the top, or after `cargo fmt` so it is not blocked by the build/test

mzabaluev: Maybe, but now it only dumps warnings into GitHub logs. When we are ready to deny clippy warnings, we can move it to the top after the `fmt` check.

danielSanchezQ: Ah yes its true, it just dump warnings.

mzabaluev:
Testing of cache performance is frustrated by flakiness in the cache service: https://github.com/actions/cache/issues/267.
Perhaps the introduction of caching should be postponed until these problems are resolved.

mzabaluev:
The cache has been behaving well lately and I have made a few more improvements, so trying to resurrect this PR if GitHub lets me.

mzabaluev:
No luck, filed as #383

PR #382, opened 2020-05-15
Crypto generators

danielSanchezQ:
Created generators for:
* vrf
* ed25519
* blake2b256

Implemented utils generators:
* Unique items generator

Implemented insecure random number generator wrapper (just for testing).


chain-test-utils/src/generators/utils/constant_generator.rs
```diff
@@ -0,0 +1,18 @@
+use crate::generators::utils::{Generator, R};
+
+#[derive(Clone)]
+pub struct ConstantGenerator<T>(T);
```
danielSanchezQ, line 4: For some reason, the one in `smoke` wasn't public.

NicolasDP: make it public, push a PR

danielSanchezQ: My bad, actually it was public but starting lowercase 😅

Cargo.toml
```diff
@@ -8,6 +8,7 @@ members = [
     "chain-crypto",
     "chain-network",
     "chain-storage",
+    "chain-generators",
     "chain-test-utils",
```
mzabaluev, line 12: Should we try to keep all test utilities in a single crate? Or is `chain-test-utils` meant to be a core library onto which other generator crates with their varying dependencies are based? Trying to capture the intent here.

chain-generators/src/crypto/ed25519.rs
```diff
@@ -0,0 +1,48 @@
+use chain_crypto;
```
mzabaluev, line 1: Remove the lint

chain-generators/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "chain-generators"
+version = "0.1.0"
+authors = ["danielsanchezq <daniel.sanchez@iohk.io>"]
+edition = "2018"
+license = "MIT OR Apache-2.0"
+
```
mzabaluev, line 7: It would be nice to leave a description for the purpose of the crate.

chain-test-utils/Cargo.toml
```diff
@@ -10,4 +10,6 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 chain-core = { path = "../chain-core" }
 quickcheck = "0.9.2"
-quickcheck_macros = "0.9.1"
\ No newline at end of file
+quickcheck_macros = "0.9.1"
+smoke = "0.1.0"
+rand = "0.7.3"
```
mzabaluev, line 15: Let's put a newline here so that it stops bugging the diff :)

Cargo.toml
danielSanchezQ: What I had in mind was that in `utils` we should have whatever that deal with the chain libs testing part but that is common for them. What we have already, for example, the utilities for checking the serialization bijection. Also, I added the `NonSecureRng` in the utils because even if it is part of the generators stuff I think it is a general concept enough to be separated from the main ones. Also, some generators that I have in mind for the future (for example a generator that keeps track of the generated values to discard the duplicated ones) I was planning to put them here. By keeping them separated we may also import just the utils ones for implementing the same for some jormungandr types if we do not really need any chain type.
I don't say it is the best option though, and I see the value on keeping everything together (for example with a utils submodule inside the `chain-generators` crate).
What do @input-output-hk/jormungandr team think?

mzabaluev: I think it's OK to have a library of common test utilities and another providing the generator framework. Just keep these responsibilities clearly separated.

danielSanchezQ:
Closing this for now. We are taking other dev paths for this.

PR #384, opened 2020-05-18
CircleCI improvements

mzabaluev:
Improve cache affinity in the vein of #383.

Switch to config version 2.1, which allows to remove much of the repetition.


mzabaluev:
These changes shave up to a minute off the typical good day build latencies seen recently in master.
So we can keep CircleCI as an early warning, completing its few most important test configurations in 1/3 of the time it takes the GitHub workflow runner to plod through its test matrix (which is going to get expanded in #383).

PR #385, opened 2020-05-19
chain-storage: sled migration + permanent store

eugene-babichenko:


eugene-babichenko:
`cargo bench --features with-bench --bench sled -- --noplot` to run a benchmark, replace `sled` with `storage` if you want to benchmark the current SQLite implementation.

eugene-babichenko:
**TBD**

- [ ] Implementation of `chain_storage::Block` for `chain_impl_mockchain::Block` visible to external users
- [ ] Integration with jormungabdr

NicolasDP:
somehow the test are taking too long and hanging in forever

eugene-babichenko:
**TODO** break down the `info` tree into separate trees - one per `BlockInfo` field and see how it affects the performance. The idea is to avoid serialization/deserialization of `BlockInfo`.

chain-storage/src/lib.rs
```diff
@@ -1069,11 +798,145 @@ pub mod tests {
...
+            genesis_block.chain_length,
+        );
+        store
+            .put_block(&genesis_block.serialize_as_vec(), genesis_block_info)
+            .unwrap();
+
+        let mut blocks = vec![];
+
+        for _i in 0..N_BLOCKS {
+            let block = genesis_block.make_child(None);
```
NicolasDP, line 918: Add a test with forks in too

eugene-babichenko: This code create 5 different branches actually. Could you please clarify what do you mean?

chain-impl-mockchain/src/multiverse.rs
```diff
@@ -258,7 +260,13 @@ mod test {
         */
```
NicolasDP, line 260: Do we still need this dead code?

chain-storage/src/lib.rs
```diff
@@ -1,604 +1,424 @@
...
-                block_info_row_id as isize,
-            )
-            .unwrap();
+                info.insert(block_info.id(), block_info.serialize().as_slice())?;
 
-        tx.commit()
-            .map_err(|err| Error::BackendError(Box::new(err)))?;
+                Ok(Ok(()))
+            },
+        );
```
NicolasDP, line 184: This is a big lambda and could be a free function. It would also allow you to write individual tests eventually.

chain-storage/src/lib.rs
```diff
@@ -1,604 +1,424 @@
...
+            // that.
+            match result {
+                Ok(Ok(maybe_next_tip)) => {
+                    maybe_current_tip = maybe_next_tip;
+                }
+                Ok(Err(err)) => return Err(err),
+                Err(TransactionError::Storage(err)) => {
+                    return Err(Error::BackendError(Box::new(err)))
+                }
+                Err(TransactionError::Abort(())) => unreachable!(),
```
NicolasDP, line 403: can't we do the pruning in one transaction instead?

chain-storage/src/lib.rs
```diff
@@ -607,173 +427,118 @@ where
...
+                let mut block_info_reader: &[u8] = &block_info_bin;
+                BlockInfo::deserialize(&mut block_info_reader)
+            })?;
     }
+
+    Ok(current)
 }
 
 #[cfg(any(test, feature = "with-bench"))]
 pub mod test_utils {
```
NicolasDP, line 541: move this in a different file too so this file becomes smaller and easier to read/review

chain-storage/src/lib.rs
```diff
@@ -1,604 +1,424 @@
...
-    /// get_nth_ancestor(), and calls put_block_internal() to do the
-    /// actual write.
-    pub fn put_block(&mut self, block: &B) -> Result<(), Error> {
-        let block_hash = block.id();
-
-        let mut index = self.index.write().unwrap();
-
-        let tx = self
+    /// Write a block to the store. The parent of the block must exist (unless
+    /// it's the zero hash).
```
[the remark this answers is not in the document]
eugene-babichenko: Had the same thought considering we want to have a separate persistent storage in the future.

chain-storage/src/lib.rs
eugene-babichenko: Technically not a big problem. But if a branch is too big this transaction will lock the database for a relatively long time (need to measure that though).

eugene-babichenko:
> 
> 
> **TODO** break down the `info` tree into separate trees - one per `BlockInfo` field and see how it affects the performance. The idea is to avoid serialization/deserialization of `BlockInfo`.

Dropping this, significantly slower than using serialized `BlockInfo` (around 20-25%).

chain-storage/src/block_info.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Clone)]
+pub struct BlockInfo {
+    id: Box<[u8]>,
+    parent_id: Box<[u8]>,
+    chain_length: u32,
+    ref_count: u32,
+}
+
+impl BlockInfo {
+    pub fn new(id: Vec<u8>, parent_id: Vec<u8>, chain_length: u32) -> Self {
```
mzabaluev, line 12: The `Vec` parameters could be generalized to something like `impl Into<Box<[u8]>>`.

chain-storage/benches/storage.rs
```diff
@@ -1,8 +1,7 @@
 use criterion::{criterion_group, criterion_main, BatchSize, Criterion};
 use rand_core::{OsRng, RngCore};
 
-use chain_core::property::Block as _;
-use chain_storage::{test_utils::Block, BlockStoreBuilder};
+use chain_storage::{test_utils::Block, BlockInfo, BlockStore};
```
mzabaluev, line 4: Is `test_utils` part of the production library, or should it be cfg-gated for tests?

eugene-babichenko: Not a part of production. It is enabled for `cfg(test)` and `with-bench` feature.

mzabaluev: Ah right, this is in benches.

chain-impl-mockchain/src/multiverse.rs
```diff
@@ -208,8 +208,8 @@ mod test {
     };
 
     use chain_addr::Discrimination;
-    use chain_core::property::Block as _;
-    use chain_storage::BlockStoreBuilder;
+    use chain_core::property::{self, Block as _, BlockId as _, Deserialize, Serialize};
```
NicolasDP, line 211: can we remove this deps?

eugene-babichenko: Yes, looks like it is only used in tests. Will do.

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,832 @@
...
+//! Since blockchain can be branching extensively, this library provides the
+//! mechanism to have multiple blockchain branches. However, doing so for all
+//! block screates a lot of overhead for maintaining volatile storage. At the
+//! same time, branches usually die out pretty fast.
+//!
+//! Given that, the storage is separated into two parts:
+//!
+//! * Volatile storage should be used for the part of a block chain, where a
+//!   developer may need access to different branches.
+//! * Permanent storage for the part of block chain, that is guaranteed not to
```
NicolasDP, line 27:
```suggestion
//! * Permanent storage for the part of the blockchain that is guaranteed not to
```

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,832 @@
...
+//! # Volatile + permanent storage model
+//!
+//! Since blockchain can be branching extensively, this library provides the
+//! mechanism to have multiple blockchain branches. However, doing so for all
+//! block screates a lot of overhead for maintaining volatile storage. At the
+//! same time, branches usually die out pretty fast.
+//!
+//! Given that, the storage is separated into two parts:
+//!
+//! * Volatile storage should be used for the part of a block chain, where a
```
NicolasDP, line 25:
```suggestion
//! * Volatile storage should be used for the part of a blockchain where a
```

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,832 @@
...
+//! * Permanent storage for the part of block chain, that is guaranteed not to
+//!   change anymore.
+//!
+//! ## Moving blocks to the permanent store.
+//!
+//! If you have this block structure and call
+//! `store.flush_to_permanent_store(block 2 id)`, then `Block 1` and `Block 2`
+//! will be moved to the permanent storage. If you call
+//! `store.flush_to_permanent_store(block 3 id)`, `Block 3` will also be moved
+//! to the permanent store, but `Block 3'` will still exist. Note that if you
```
NicolasDP, line 36:
```suggestion
//! to the permanent store, but `Block 3'` will still exist in the volatile store. Note that if you
```

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,832 @@
...
+//! call `store.get_blocks_by_chain_length(3)` only `Block 3` (which is in the
+//! permanent store) will be returned; and you cannot call
+//! `store.flush_to_permanent_store(block 3' id)` now.
```
NicolasDP, line 39: Since Block 3 is not in the permanent storage, the fact that the _prime_ branch (`Block 3'` and `Block 4'`) are forking from a block that is already being dismissed should flush it out of the volatile storage: i.e. as soon a branch points to a block in the permanent storage that is below the last block of the permanent storage needs to be removed/pruned.

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,832 @@
...
+//! |              |       |             |
+//! +--------------+       +-------------+
+//! |              |
+//! | block1       |
+//! |              |
+//! +--------------+       +-------------+
+//! |              |       |             |
+//! | len(block2)  |       | block2 pos  |
+//! |              |       |             |
+//! +--------------+       +-------------+
```
NicolasDP, line 122: You don't need to store the `len(block*)` in the store. It is known from the `block no. index`. For example the length of the `block N` is the `(block N+1 pos)    -   (block N pos)`

eugene-babichenko: Yes, I plan to remove that.

chain-storage/src/lib.rs
eugene-babichenko: Not sure if we want to do that automatically in the storage module. The reason for this is that we will need to scan all branches to check if they begin in the permanent storage. Instead of this we can just leave removing branches up to the application logic and this is most likely that the application will actually do that.

chain-storage/src/block_info.rs
```diff
@@ -0,0 +1,88 @@
...
+    parent_id: Value,
+    chain_length: u32,
+    /// This field is used internally by the volatile storage only. Its purpose
+    /// is to store the number of blocks that maintain this block as a parent +
+    /// the number of tags for this block. A block CANNOT be removed from the
+    /// volatile storage if the reference counter is greater than 1. For blocks
+    /// from the permanent storage this value is always equal to 1 and MUST NOT
+    /// be used.
+    /// NOTE: "removing a block" relates only to removing an abanded branch
+    /// entirely and does not apply to moving a block to the permanent storage.
```
mzabaluev, line 19: Regular comments should be used for documenting internal fields.

chain-storage/src/iterator.rs
```diff
@@ -0,0 +1,147 @@
...
+        };
+        assert!(
+            to_info.chain_length() + 1 >= distance,
+            "expected distance {} <= chain length {}",
+            distance,
+            to_info.chain_length() + 1
+        );
+        let from_length = to_info.chain_length() + 1 - distance;
+
+        let state = if let Some(_) = permanent_store.get_block_by_chain_length(from_length) {
```
mzabaluev, line 48: Clippy complains here; could be made a `match` instead.

chain-storage/src/iterator.rs
```diff
@@ -0,0 +1,147 @@
...
+        stop_at_length: u32,
+    },
+    Volatile {
+        ids: Vec<Value>,
+    },
+}
+
+impl StorageIterator {
+    /// Create a new iterator. `from` and `to` values MUST have a path between
+    /// them and this needs to be checked before the iterator is started.
```
mzabaluev, line 25: How is this MUST enforced?
Please detail if it is a panic, by using a conventional `# Panics` section.

chain-storage/src/iterator.rs
```diff
@@ -0,0 +1,147 @@
...
+                        Some(Value::permanent(item))
+                    }
+                    None => {
+                        self.state = IteratorState::Volatile {
+                            ids: gather_blocks_ids(
+                                self.to.clone(),
+                                &self.block_info,
+                                *current_length,
+                            )
+                            .ok()?,
```
mzabaluev, line 94: As I said previously, exposing the iterator API on fallible operations is fraught with problems.
Here, the iterator will stop on failure, and the consumer can only distinguish the failure case from end-of-iterator case by looking into the block metadata to check if the blocks received are what was requested. Then there will be code that will not do that and cause mysterious havoc when some rare error occurs.

I think it's important to avoid offering an error-ignorant iterator as the easiest available API, despite the inconvenience. I'd rather make this iterator return a `Result` and deal with the inconvenience, possibly assisted with utilities like [fallible-iterator](https://docs.rs/fallible-iterator/0.2.0/fallible_iterator/).

mzabaluev (changes requested):

chain-storage/src/iterator.rs
eugene-babichenko: Yes, sounds like a better approach. I thought you were against having `Item = Result<T, E>` so I didn't do that. Probably, just misunderstanding.

chain-storage/src/iterator.rs
eugene-babichenko: Well, this actually needs to be updated, this is an outdated piece of documentation. There is no `from` supplied to `Iterator::new`.

chain-impl-mockchain/src/multiverse.rs
```diff
@@ -352,30 +342,28 @@ mod test {
...
         let mut ids = vec![];
         for i in 1..10001 {
             date = date.next(&era);
             let block = build_bft_block(&parent, date, state.chain_length.increase(), &leader);
             state = apply_block(&state, &block);
             assert_eq!(state.chain_length().0, i);
-            assert_eq!(state.date, block.date());
-            store.put_block(&block).unwrap();
-            _ref = Some(multiverse.add(block.id(), state.clone()));
+            assert_eq!(state.date, block.header.block_date());
```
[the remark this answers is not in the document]
mzabaluev: Not a hard requirement this one.

chain-storage/src/value.rs
```diff
@@ -0,0 +1,74 @@
...
+use std::hash::{Hash, Hasher};
+
+#[derive(Debug, Clone)]
+enum ValueImpl {
+    Volatile(IVec),
+    Owned(Box<[u8]>),
+    Permanent(SharedMmap),
+}
+
+/// Wrapper for data held by the database. This wrapper holds structs return by
```
mzabaluev, line 12: spellcheck: returned

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,809 @@
...
-        for _ in 0..10 {
-            let mut parent_block = pick_from_vector(rng, &blocks).clone();
-            let r = 1 + (rng.next_u32() % 9999);
-            for _ in 0..r {
-                let block = parent_block.make_child(None);
-                store.put_block(&block).unwrap();
-                parent_block = block.clone();
-                blocks.push(block);
+#[inline]
+#[allow(clippy::too_many_arguments)]
```
mzabaluev, line 713: Well, maybe clippy is trying to tell you something there :wink:

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,809 @@
...
+            return Ok(true);
         }
+
+        info.get(block_id)
+            .map(|maybe_block| maybe_block.is_some())
+            .map_err(Into::into)
     }
 
-    // Determine whether block 'ancestor' is an ancestor of block 'descendent'
+    /// Determine whether block 'ancestor' is an ancestor of block 'descendent'
```
mzabaluev, line 441: As this is public API, please fix spelling to 'descendant'.

BTW, the doc does not refer to correct argument names and the convention is to have them in backticks.

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,809 @@
...
+                        info,
+                        chain_length_to_block_ids,
+                        tips,
+                        &self.permanent,
+                        current_tip,
+                        self.root_id.as_ref(),
+                        self.id_length,
+                    )? {
+                        Ok(res) => res,
+                        Err(err) => return Ok(Err(err)),
```
mzabaluev, line 404: To make this clear: this lets all previous removals in the transaction be committed, even though this one fails?
Then the method returns `Err` anyway?
Would it be better to roll back the whole transaction?

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,809 @@
...
+fn remove_tip_impl(
+    blocks: &TransactionalTree,
+    info: &TransactionalTree,
+    chain_length_to_block_ids: &TransactionalTree,
+    tips: &TransactionalTree,
+    permanent_store: &PermanentStore,
+    block_id: &[u8],
+    root_id: &[u8],
+    id_size: usize,
+) -> Result<Result<RemoveTipResult, Error>, ConflictableTransactionError<()>> {
```
mzabaluev, line 723: Maybe, for everyone's sanity, make it a one-level `Result`, with an internal enum for the `Ok` value to carry outcomes considered non-fatal for the transaction by the logic of the `remove_tip` method.

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,809 @@
...
 
-#[derive(Clone, Debug)]
-pub struct BackLink<Id: BlockId> {
-    /// The distance to this ancestor.
-    pub distance: u64,
-    /// The hash of the ancestor.
-    pub block_hash: Id,
+enum RemoveTipResult {
+    NextTip(Vec<u8>),
+    HitPermanentStore(Vec<u8>),
```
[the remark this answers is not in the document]
mzabaluev: OK, I've learned that block IDs are just bags of bytes just like every other values for storage.

Something like this would be more self-explanatory:

```rust
enum RemoveTipResult {
    NextTip { id: Vec<u8> },
    HitPermanentStore { id: Vec<u8> },
    Done,
}
```

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,809 @@
...
+//! ```
+//!
+//! # Storage directory layout
+//!
+//! ```ignore
+//! store
+//! ├── permanent       - permanent storage directory
+//! │   └── flatfile    - storage file that can be transferred over the network
+//! └── volatile        - volatile storage
+//! ```
```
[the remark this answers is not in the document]
eugene-babichenko: `apt install tree` or `pacman -S tree` if we are talking about fancy trees :)
http://asciiflow.com/ for everything else.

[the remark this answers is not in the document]
eugene-babichenko: Probably should use DOT for graphs and convert it to ASCII :D

chain-storage/src/lib.rs
eugene-babichenko: Apparently, this can be rewritten to `Result<RemoveTipResult, ConflictableTransactionError<Error>>` in the latest version of `sled` and I will do so.

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,787 @@
...
-                    Value::Blob(block_hash.serialize_as_vec().unwrap()),
-                ]),
-            None => {
-                if index.get_block(block_hash).is_none() {
-                    return Err(Error::BlockNotFound);
-                }
+        if let RemoveTipResult::HitPermanentStore(id) = result {
+            let block_info = self
+                .get_block_info(&id)
+                .expect("parent block in permanent store not found");
```
mzabaluev, line 427: Perhaps a store consistency error case would be good to use here, too?

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,787 @@
...
-
-    impl chain_core::property::ChainLength for ChainLength {
-        fn next(&self) -> Self {
-            Self(self.0 + 1)
-        }
+        true
+    };
+
+    if parent_in_volatile_store {
+        let parent_block_info_bin = info.get(block_info.parent_id())?.unwrap();
```
mzabaluev, line 655: Another data inconsistency panic that we need to decide what to do with.

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,787 @@
...
-        let tx = self
-            .inner
-            .transaction()
-            .map_err(|err| Error::BackendError(Box::new(err)))?;
+        let blocks = self.volatile.open_tree(tree::BLOCKS)?;
+        let chain_length_to_block_ids = self.volatile.open_tree(tree::CHAIN_LENGTH_INDEX)?;
 
-        Self::get_block_internal(&tx, &index, block_hash)
-    }
+        let chain_length_index_prefix = chain_length.to_le_bytes();
```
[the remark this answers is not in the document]
eugene-babichenko: Makes sense from this point of view! Actually, I just wanted to use the same endiannes as the vast majority of hardware this will be ran on.

chain-impl-mockchain/src/multiverse.rs
```diff
@@ -244,21 +242,13 @@ mod test {
                 break state_ref;
             }
 
-            let cur_block_info = store.get_block_info(&cur_hash).unwrap();
-            blocks_to_apply.push(cur_hash);
-            cur_hash = cur_block_info.parent_id();
+            let cur_block = store.get(&cur_hash).unwrap();
+            blocks_to_apply.push(cur_hash.clone());
+            cur_hash = Hash::deserialize(cur_block.header.block_parent_hash().as_ref()).unwrap();
```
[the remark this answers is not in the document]
eugene-babichenko: I doubt this is really needed in the testing code.

chain-storage/src/lib.rs
eugene-babichenko: It definitely does, but having the entire body of this function in the closure look even worse :)

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,827 @@
...
-            })
-            .unwrap();
-
-            assert_eq!(
-                ancestor_info.chain_length + distance,
-                block.chain_length().0
-            );
+    let block_info_bin = info.get(block_id)?.ok_or(ConsistencyFailure::BlockInfo)?;
+    let mut block_info_reader: &[u8] = &block_info_bin;
+    let block_info = BlockInfo::deserialize(&mut block_info_reader, id_size, block_id.to_vec())?;
```
mzabaluev, line 768: Could do with

```rust
    let block_info = BlockInfo::deserialize(&block_info_bin[..], id_size, block_id.to_vec())?;
```

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,827 @@
...
+#[allow(clippy::too_many_arguments)]
+fn remove_tip_impl(
+    blocks: &TransactionalTree,
+    info: &TransactionalTree,
+    chain_length_to_block_ids: &TransactionalTree,
+    tips: &TransactionalTree,
+    permanent_store_index: &TransactionalTree,
+    block_id: &[u8],
+    root_id: &[u8],
+    id_size: usize,
```
mzabaluev, line 756: Maybe instead of dragging these around everywhere, they should be in a `StoreOptions` struct.

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,827 @@
...
-        let blocks_1 = blocks.clone();
+    let parent_block_info_bin = info
+        .get(block_info.parent_id())?
+        .ok_or(ConsistencyFailure::MissingParentBlock)?;
+    let mut parent_block_info_reader: &[u8] = &parent_block_info_bin;
+    let mut parent_block_info = BlockInfo::deserialize(
+        &mut parent_block_info_reader,
+        id_size,
+        block_info.parent_id().clone(),
+    )?;
```
mzabaluev, line 805: Consider a helper function for getting block info from the sled store, this is becoming repetitive.

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,827 @@
...
+                Error::BlockNotFound => Ok(None),
+                e => Err(e),
+            })?
+        {
+            distance += 1;
+            if parent_block_info.id().as_ref() == ancestor_id {
+                return Ok(Some(distance));
+            }
+            current_block_info = parent_block_info;
+        }
```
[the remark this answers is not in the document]
eugene-babichenko: Oh, that makes sense. I don't think backlinks were ever used in a way they provide any performance benefits.

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,827 @@
...
+mod tests;
+mod value;
 
-#[derive(Clone, Debug)]
-pub struct BackLink<Id: BlockId> {
-    /// The distance to this ancestor.
-    pub distance: u64,
-    /// The hash of the ancestor.
-    pub block_hash: Id,
+use permanent_store::PermanentStore;
```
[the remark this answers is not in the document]
eugene-babichenko: Do you mean trying to separate it, or just move everything to a module and leave `lib.rs` only with the documentation and re-exports?

chain-storage/src/lib.rs
```diff
@@ -1,1079 +1,827 @@
...
-    fn block_exists_internal(
-        connection: &Connection,
-        index: &ChainStorageIndex<B>,
-        block_hash: &B::Id,
-    ) -> Result<bool, Error> {
-        match Self::get_block_info_internal(connection, index, block_hash) {
-            Ok(_) => Ok(true),
-            Err(Error::BlockNotFound) => Ok(false),
-            Err(err) => Err(err),
+        if self.permanent.contains_key(block_id)? {
```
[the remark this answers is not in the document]
eugene-babichenko: Good point.

mzabaluev (approved):

chain-storage/src/lib.rs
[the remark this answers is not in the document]
eugene-babichenko: Actually, this is used only in this library now and not by jormungandr. Makes sense to just remove it entirely, write a backwards iterator when needed and write and replace `get_nth_ancestor` with a method that is better optimized.

Aside from just walking all the way back to the n-th ancestor we can do the following:

1. Assume that the storage is consitent.
2. Get blocks at length `ancestor_len = descendant.block_length - distance`
3. If there is only one block at `ancestor_len` it is an ancestor of the given block.

This should be a rather common scenario if we do garbage collection.

chain-storage/src/lib.rs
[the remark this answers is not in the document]
eugene-babichenko: > I think it's negligible compared to the index access operations

And you are right.

chain-storage/src/lib.rs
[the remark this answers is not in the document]
mzabaluev: Right, this fast path check will succeed most of the time.

chain-storage/src/lib.rs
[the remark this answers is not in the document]
mzabaluev: Only the latter; ideally `lib.rs` and `mod.rs` should be catalogues of their constituent modules plus re-exports.

mzabaluev (approved):

mzabaluev (approved):

PR #386, opened 2020-05-19
add bft leader id getter

enzoc4:
I realized/remembered that the jormungandr explorer has an `unimplemented!` for the bft leader id tracking per block, because there is no way of getting the leader id (I think). [code here](https://github.com/input-output-hk/jormungandr/blob/e661d520896d4cceacd14a30affdf1025157ec32/jormungandr/src/explorer/indexing.rs#L226)

Nobody seems to be using it in bft mode anyway, but it's just a few lines of code.

Feel free to close this if there is actually a way to get the id that I missed.


NicolasDP:
Good catch, this is going to be necessary for the VIT

PR #387, opened 2020-05-26
Fix and improve caching in GitHub CI

mzabaluev:
There should be no weird failures to unlink git pack files
in the cargo registry index on Windows.

Use two caches side by side, one for the registry index and
another for dependency crates.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,6 +31,7 @@ jobs:
             git ls-remote --heads https://github.com/rust-lang/crates.io-index.git master |
             cut -f 1
           )
+          echo "$commit"
           echo "::set-output name=head::$commit"
 
       - name: Cache cargo registry index
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,6 +10,8 @@ jobs:
   update_deps:
     name: Update dependencies
     runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


eugene-babichenko:
Could we also store `Cargo.lock` in cache instead of artifacts storage? And also use an output of `update_deps` in downstream jobs instead of `hashFiles('Cargo.lock')`.

mzabaluev:
> Could we also store `Cargo.lock` in cache instead of artifacts storage?

I'd rather not for multiple reasons: We don't want the same `Cargo.lock` to persist between builds. `actions/cache@v1` does not accept multiple paths, so it'd have to be a separate cache just for `Cargo.lock` which seems like waste to me. There is `v2` now, but it's still too buggy. The artefact actions are lightning fast, while cache is slower and has latency gremlins.

> And also use an output of `update_deps` in downstream jobs instead of `hashFiles('Cargo.lock')`.

That's a good idea in light of the Windows line ending silliness. I have quickly adapted this from an end executable build workflow where `Cargo.lock` is checked in; I think it can be further reworked for the library.

mzabaluev:
> And also use an output of `update_deps` in downstream jobs instead of `hashFiles('Cargo.lock')`.

In the end, it would be unnecessary complication to the workflow as we pass the `Cargo.lock` between jobs anyway and its line endings are not fudged by artefact download on Windows.

PR #388, opened 2020-05-26
Add timeout to GitHub CI test jobs

mzabaluev:
Saw 6 hour workflow runs lately due to a hangup in tests.


mzabaluev:
I've got the whole workflow to run 15 minutes with my "optimizations" in #387 :grin:
Let's be generous.

PR #389, opened 2020-05-26
remove all unused imports...

NicolasDP:
remove all remaining unused imports in chain-libs

### DIFF
--- a/chain-impl-mockchain/src/config.rs
+++ b/chain-impl-mockchain/src/config.rs
@@ -765,7 +765,9 @@ impl TagLen {
 #[cfg(any(test, feature = "property-test-api"))]
 mod test {
     use super::*;
-    use quickcheck::{Arbitrary, Gen, TestResult};
+    #[cfg(test)]
+    use quickcheck::TestResult;
+    use quickcheck::{Arbitrary, Gen};
     use strum::IntoEnumIterator;
 
     quickcheck! {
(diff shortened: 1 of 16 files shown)


chain-impl-mockchain/src/config.rs
```diff
@@ -765,7 +765,9 @@ impl TagLen {
 #[cfg(any(test, feature = "property-test-api"))]
 mod test {
     use super::*;
-    use quickcheck::{Arbitrary, Gen, TestResult};
+    #[cfg(test)]
+    use quickcheck::TestResult;
+    use quickcheck::{Arbitrary, Gen};
```
mzabaluev, line 770: It would be better if we kept the API parts that are enabled with `property-test-api` feature to a separate module, which I tried to insist is called `testing` (BTW, the conventional name for unit test modules is `tests`, not `test`, but I'm probably being too OCD about this). But I can't see how this can even export any API since the module is private.

It might help to build with `--features property-test-api` and resolve any warnings about unused things also there.

chain-impl-mockchain/src/date.rs
```diff
@@ -121,7 +121,8 @@ impl str::FromStr for BlockDate {
 mod tests {
```
mzabaluev, line 121: Same here, this probably does not need the `cfg(feature = ...)` alternative at all. And everywhere else where the mod is private.

chain-impl-mockchain/src/config.rs
NicolasDP: I am putting my faith in the chain-generator for better API on the long run. This is problematic because of how the chain-libs expose property-test-api in order to help the scenario based testing.

PR #393, opened 2020-05-30
enable some vote related functionality

rinor:
- [x] Fragment::VoteCast

Should we enable also this one or not yet?
- [ ] Fragment::VotePlan

---

Leaving disabled for now:
- [ ] Fragment::UpdateProposal
- [ ] Fragment::UpdateVote


chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -884,10 +881,6 @@ impl Ledger {
                 // operating with jormungandr nodes (ITN)
                 // see to have a setting to prevent anyone from submitting
                 // a vote when not allowed
```
NicolasDP, line 883: remove the comments too

PR #400, opened 2020-06-10
Encode legacy node IDs as bincode

mzabaluev:
:man_facepalming:

Also add "legacy" feature gate under which `bincode` comes as a dependency.

### DIFF
--- a/chain-network/Cargo.toml
+++ b/chain-network/Cargo.toml
@@ -7,6 +7,7 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 async-trait = "0.1"
+bincode = { version = "1.2", optional = true }
 futures = "0.3"
 pin-project = "0.4"
 prost = "0.6"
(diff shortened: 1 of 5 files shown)


chain-network/Cargo.toml
```diff
@@ -7,6 +7,7 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 async-trait = "0.1"
+bincode = { version = "1.2", optional = true }
```
NicolasDP, line 10: oh no... can't we just add the missing encoding format by hand here?

mzabaluev: Why, I'd rather write in another dependency than figure it out :)
What's the framing, anyway? Big endian marshalling?

mzabaluev: It literally takes more time for me to dig through `bincode` sources, because the format is not documented.

mzabaluev: And then there's [this](https://github.com/servo/bincode/blob/439b12dbcba903b7681879f96f932dc20b0a5115/src/config/int.rs#L182). I'd rather delegate it to the dependency.

PR #407, opened 2020-06-12
GitHub CI: switch to actions/cache@v2

mzabaluev:
There's a bug in v1 that keeps the registry index cache in an old state that's started breaking the builds.

There is still a bug in v2 as well, that results in separate cache values made by Windows jobs due to issues with parameter normalization, but this will self-correct once the action bug is fixed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,22 +35,22 @@ jobs:
           echo "::set-output name=head::$commit"
 
       - name: Cache cargo registry index
-        uses: actions/cache@v1
+        uses: actions/cache@v2
         with:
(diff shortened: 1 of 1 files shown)


mzabaluev:
It was not the index state (though v1 does bug out and gives us an old cache value), it's a regression in nightly cargo: https://github.com/rust-lang/cargo/issues/8351

mzabaluev:
The cache bug is still there though :disappointed:

```
Cache restored from key: cargo-index-v2-
```

mzabaluev:
> v1 does bug out and gives us an old cache value

That was not true either, it was a typo on my part. Well, now that I have ported the thing anyway...

PR #417, opened 2020-06-16
implement Display for vote Weight

eugene-babichenko:


chain-impl-mockchain/src/vote/tally.rs
```diff
@@ -129,3 +130,9 @@ impl From<Weight> for u64 {
         w.0
     }
 }
+
+impl fmt::Display for Weight {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0)
```
NicolasDP, line 136: In that case I think the usage is to defer the printing logic to the `u64` type:

```suggestion
        self.0.fmt(f)
```

PR #444, opened 2020-08-28
add secret key to_inner

enzoc4:
Context: For the QR code generation for the voting app, we need to generate private keys and be able to write them. But currently the only way I find of getting the key out is as bech32. I'm not sure if there are any security concerns with doing it this way?

### DIFF
--- a/chain-crypto/src/key.rs
+++ b/chain-crypto/src/key.rs
@@ -167,6 +167,9 @@ impl<A: AsymmetricKey> SecretKey<A> {
     pub fn from_binary(data: &[u8]) -> Result<Self, SecretKeyError> {
         Ok(SecretKey(<A as AsymmetricKey>::secret_from_binary(data)?))
     }
+    pub fn inner(self) -> A::Secret {
+        self.0
+    }
 }
 
 impl<A: AsymmetricPublicKey> PublicKey<A> {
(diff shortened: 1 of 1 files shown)


chain-crypto/src/key.rs
```diff
@@ -167,6 +167,9 @@ impl<A: AsymmetricKey> SecretKey<A> {
     pub fn from_binary(data: &[u8]) -> Result<Self, SecretKeyError> {
         Ok(SecretKey(<A as AsymmetricKey>::secret_from_binary(data)?))
     }
+    pub fn security_leak(self) -> A::Secret {
+        self.0
+    }
```
NicolasDP, line 172: Missing a comment on the function to explain what this function does: leak the secret key, giving back the ownership of the secret key. There may not be an appropriate destructor for it anymore.

mzabaluev: I'd name it `leak_secret` and document the effect. Otherwise I see no problem with adding the method.

enzoc4: I don't see a `Drop` implementation for `SecretKey` anyway, though.

chain-crypto/src/key.rs
```diff
@@ -167,6 +167,12 @@ impl<A: AsymmetricKey> SecretKey<A> {
     pub fn from_binary(data: &[u8]) -> Result<Self, SecretKeyError> {
         Ok(SecretKey(<A as AsymmetricKey>::secret_from_binary(data)?))
     }
+    // get ownership of the secret key, allowing to get its byte representation.
```
mzabaluev, line 170: Could be a doc comment, but then again the whole module API is currently undocumented.

mzabaluev (approved):

PR #445, opened 2020-09-07
Add node ID authentication to protocol handshake

mzabaluev:
Modify the `Handshake` request and response to add server node ID authentication.
Add a `ClientAuth` RPC method to authenticate the client's node ID using the server-side nonce returned in the `Handshake` response.

### DIFF
--- a/chain-network/src/data/p2p.rs
+++ b/chain-network/src/data/p2p.rs
@@ -98,3 +98,9 @@ impl AuthenticatedNodeId {
         &self.signature
     }
 }
+
+impl From<AuthenticatedNodeId> for NodeId {
+    fn from(auth: AuthenticatedNodeId) -> Self {
+        auth.id
+    }
+}
--- a/chain-network/src/data/p2p.rs
+++ b/chain-network/src/data/p2p.rs
@@ -52,14 +52,14 @@ impl NodeId {
         &self.0
     }
 
-    pub fn signed(self, signature: &[u8]) -> Result<SignedNodeId, Error> {
+    pub fn authenticated(self, signature: &[u8]) -> Result<AuthenticatedNodeId, Error> {
         if signature.len() != NODE_SIGNATURE_LEN {
             return Err(Error::new(
                 Code::InvalidArgument,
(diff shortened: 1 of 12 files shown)


mzabaluev:
Authentication could be implemented inside `chain-network`, but I was hesitant to add a dependency of chain-crypto or directly on crypto algorithm crates. On the other hand, the sizes of the public key and the signature are hardcoded to match those used in Ed25519, which is what the jormungandr part of the change uses to authenticate.

Perhaps the auth field format should be made self-descriptive a-la multikey.

chain-network/src/data/p2p.rs
```diff
@@ -27,3 +30,77 @@ impl fmt::Display for Peer {
...
+}
+
+impl NodeId {
+    #[inline]
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.0
+    }
+
+    pub fn authenticated(self, signature: &[u8]) -> Result<AuthenticatedNodeId, Error> {
+        if signature.len() != NODE_SIGNATURE_LEN {
```
eugene-babichenko, line 56: The `VerificationAlgorithm` trait has `SIGNATURE_SIZE` within it. Can we reuse is somehow?

mzabaluev: That would bring a dependency on `chain-crypto` and all its dependencies into chain-network, as mentioned in https://github.com/input-output-hk/chain-libs/pull/445#issuecomment-698795009.

This, and also exporting verification API on `NodeId`, is perhaps better, to make the functionality self-contained.

eugene-babichenko: If so, then why not include it in the `Node` trait?

mzabaluev: That would entangle the `Node` trait into the signature of `NodeId`. Maybe `NodeId` should encapsulate a `Box<[u8]>` instead, and the same for the signature field, to keep things flexible and allow the application to choose an algorithm freely. It's not performance critical and the API uses allocated values elsewhere, anyway.

mzabaluev:
Gone for a dependency on `chain-crypto` and batteries-included authentication API in `chain-network`.

PR #449, opened 2020-09-16
Encvote fixings

danielSanchezQ:
The `ed25519_dalek::Digest` was causing some problems. The first solution was to use `SHA256`. But Im not sure if this is a problem. Tests run smooth...


.circleci/config.yml
```diff
@@ -89,7 +89,7 @@ commands:
         default: ''
       cargo_behavior:
         type: string
-        default: --locked --offline --verbose
```
mzabaluev, line 92: Drop this for now, we should fix the CI separately and @vincenthz said he has something for this.

chain-vote/src/lib.rs
```diff
@@ -121,6 +123,7 @@ impl Tally {
     }
 }
 
+#[allow(clippy::ptr_arg)]
```
mzabaluev, line 126: The last argument could be &[TallyDecryptShare], clippy is right, no need to pass a `Vec` by reference.

mzabaluev (commented):
Added some comments, but the PR structure is a bit difficult to review.
Is Vincent (and maybe others) still working on the encvote PR, so this should be incorporated there first?

chain-vote/src/lib.rs
danielSanchezQ: I did not feel confident in changing the API. But It makes sense.

danielSanchezQ:
This was fixed already in the main branch. Closing.

PR #453, opened 2020-09-22
Multiverse GC improvements + permanent store flush fix

eugene-babichenko:
- Make `Multiverse::gc` available for any `State`.
- Allow to configure it with the desired depth (to be used with epoch stability depth).

### DIFF
--- a/chain-impl-mockchain/src/multiverse.rs
+++ b/chain-impl-mockchain/src/multiverse.rs
@@ -117,14 +117,6 @@ impl<State> Multiverse<State> {
             .insert(k, GcEntry::Retained(state.clone()));
         Ref::new(k, state)
     }
-}
-
-impl Multiverse<Ledger> {
-    /// Add a state to the multiverse. Return a `Ref` object that
-    /// pins the state into memory.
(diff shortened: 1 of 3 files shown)


mzabaluev (approved):

mzabaluev (approved):
LGTM. The lint failures have been introduced to chain-vote that's already in master, #454 should fix at least some.

chain-storage/src/value.rs
```diff
@@ -72,3 +72,13 @@ impl From<Vec<u8>> for Value {
         Self::owned(value.into_boxed_slice())
     }
 }
+
+impl fmt::Display for Value {
```
mzabaluev, line 76: I'd rather have a custom `Debug` rather than use `Display` for this.
A `Value` could be as large as a block, right?

eugene-babichenko: Maybe. The idea was to use it display hex block ids.

mzabaluev: Perhaps it's better to dump specific type instances after parsing?

eugene-babichenko: Yeah probably you're right.

PR #454, opened 2020-09-24
Envote tools integration

danielSanchezQ:
Exposed some private objects. 
Added some `from/to_bytes` wrappers.


chain-vote/src/committee.rs
```diff
@@ -119,6 +120,18 @@ impl MemberState {
...
 }
 
+impl MemberSecretKey {
+    pub fn to_bytes(&self) -> [u8; 32] {
+        self.0.sk.to_bytes()
+    }
+}
+
+impl MemberPublicKey {
+    pub fn to_bytes(&self) -> Vec<u8> {
```
mzabaluev, line 130: It's generally better to provide a non-allocating `as_bytes()` accessor that returns a slice reference:

```rust
    pub fn as_bytes(&self) -> &[u8] {
        // ...
    }
```

Making an owned value is trivially composed: `key.as_bytes().to_vec()`

For an optimized moving conversion, an `impl From<MemberPublicKey> for Vec<u8>` could be added.

chain-vote/src/gang/p256k1.rs
```diff
@@ -57,12 +57,12 @@ impl GroupElement {
         }
     }
 
-    pub fn sum<'a, I>(mut i: I) -> Self
+    pub fn sum<'a, I>(i: I) -> Self
     where
         I: Iterator<Item = &'a Self>,
```
mzabaluev, line 62: An `IntoIterator` bound would make the API more flexible, but this change does not touch this, so could be done later.

chain-vote/src/gang/p256k1.rs
```diff
@@ -57,12 +57,12 @@ impl GroupElement {
...
     {
         let mut sum = GroupElement::zero();
-        while let Some(v) = i.next() {
+        for v in i {
             sum = &sum + v;
```
mzabaluev, line 66: Another sideline note: the `Add` trait impl might take the LHS by value to be less allocatey and more convenient to use. See it's implemented in `String`.

chain-vote/src/committee.rs
danielSanchezQ: The thing is that the underlying methods already return an owned array.

mzabaluev: Maybe it's a subject for a wider refactoring then. In `chain-network` and elsewhere we have `as_bytes`, make sense to keep the APIs similar.

chain-vote/src/unit_vector.rs
```diff
@@ -28,13 +28,17 @@ impl UnitVector {
...
     pub fn iter(&self) -> UnitVectorIter {
-        UnitVectorIter(0, self.clone())
+        UnitVectorIter(0, *self)
     }
 
     pub fn len(&self) -> usize {
         self.size
     }
 
+    pub fn is_empty(&self) -> bool {
```
vincenthz, line 38: This is why I'm pathologically against "satisfying the linter" as .. `is_empty` cannot ever be useful in the case of unit_vector, as the size will always be `> 0` as enforced in `new()`.

danielSanchezQ: Makes sense though. I'll just disable the warning and add this explanation.

PR #455, opened 2020-09-25
[Test] added block/header ser/deser tests

dkijania:


chain-impl-mockchain/src/header/header.rs
```diff
@@ -66,7 +66,7 @@ impl HeaderGenesisPraos {
 }
 
 /// Header description
-#[derive(Clone)]
+#[derive(Clone, Eq, PartialEq)]
```
mzabaluev, line 69: I think, if it were needed for production, `PartialEq` can be implemented in a faster by just comparing the IDs.
For test asserts, perhaps a helper function in the test module would be better.

dkijania: right, thanks

PR #456, opened 2020-09-25
Encrypted Voting

vincenthz:
Add encrypted voting capability where the committee needs to participate to open the tally


danielSanchezQ:
LGTM. I can deal with the remaining lints later

PR #457, opened 2020-09-25
Private vote payload

danielSanchezQ:


chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -158,6 +165,49 @@ impl ProposalManager {
...
+                        }
+                    }
+                }
+            }
+        }
+
+        // TODO: What to check here?
+        // if self.check(stake.assigned(), governance, &results) {
+        //     f(&self.action)
+        // }
```
danielSanchezQ, line 201: I don't really know what should I do here

chain-impl-mockchain/src/testing/builders/tx_cert_builder.rs
```diff
@@ -211,6 +211,10 @@ pub fn tally_sign(
 
             TallyProof::Public { id, signature }
         }
+        PayloadType::Private => {
+            //TODO: how to sign this?
+            unimplemented!("Need to implement signature for private votes")
+        }
```
danielSanchezQ, line 217: We need to discuss what should the flow of signing this be.

chain-impl-mockchain/src/vote/payload.rs
```diff
@@ -96,6 +138,9 @@ mod tests {
         fn arbitrary<G: Gen>(g: &mut G) -> Self {
             match PayloadType::arbitrary(g) {
                 PayloadType::Public => Payload::public(Choice::arbitrary(g)),
+                PayloadType::Private => {
+                    unimplemented!("Arbitrary is not implemented for private votes")
+                }
```
danielSanchezQ, line 143: I don't even think it can be implemented for now because all of them should be signed with the same key or it will be futile since they won't work together.

danielSanchezQ:
I will need to rebase again once Vincents PR #456  is merged.

chain-impl-mockchain/src/vote/manager.rs
enzoc4: So, as far as I understand, this is used for the on-chain "automatic" tally, which I'm not sure if we need for the moment. But, the way I see it, I don't think having this function makes sense, as it can't be executed by non-committee nodes. I imagine that for automatic on-chain actions with private votes, the committee members would need to publish a new fragment type with the `TallyDecryptShare` in order for all the nodes to validate the tally, and take governance actions.
(If this is the case I think it would be better to have two different types, btw)

chain-impl-mockchain/src/vote/payload.rs
```diff
@@ -18,11 +20,26 @@ use typed_bytes::ByteBuilder;
...
 pub enum PayloadType {
     Public = 1,
+    Private = 2,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Payload {
     Public { choice: Choice },
+    Private { encrypted_vote: EncryptedVote },
```
enzoc4, line 29: Question: don't we need the `ProofOfCorrectVote` here too? Or how do we distribute those?

danielSanchezQ: Im not sure. I think the proof is for the voter to check the encrypted onchain vote. @rinor what do you think?

chain-vote/src/unit_vector.rs
```diff
@@ -45,11 +49,8 @@ impl UnitVector {
                 "out of bounds: unit vector {} accessing index {}",
                 self.size, j
             );
-        } else if j == self.ith {
-            true
-        } else {
-            false
         }
+        j == self.ith
```
eugene-babichenko, line 53: ha ha nice one

chain-impl-mockchain/src/vote/payload.rs
eugene-babichenko: How is it used in tests exactly? If tests work _only_ with vectors of those values, then we could implement `Arbitrary for Vec<Payload>` instead and use the same vote encryption key for all payloads.

chain-impl-mockchain/src/vote/payload.rs
```diff
@@ -49,6 +67,15 @@ impl Payload {
 
         match self {
             Self::Public { choice } => bb.u8(payload_type as u8).u8(choice.as_byte()),
+            Self::Private { encrypted_vote } => {
+                let size = encrypted_vote.len();
+                let mut bb = bb.u64(size as u64);
+                for ct in encrypted_vote.iter() {
```
eugene-babichenko, line 73: Is there a reason for having multiple votes in a single payload?

chain-impl-mockchain/src/vote/payload.rs
eugene-babichenko: IMO we should have it here so that everybody could validate the vote, not only the node that included this vote into a block.

chain-impl-mockchain/src/testing/builders/tx_cert_builder.rs
eugene-babichenko: Should be signed with the sender key. What is the problem here actually?

chain-impl-mockchain/src/vote/payload.rs
danielSanchezQ: No, it is not that there are multiple votes in the payload. But the encrypted vote is itself a list of some components.

chain-impl-mockchain/src/testing/builders/tx_cert_builder.rs
rinor: @danielSanchezQ no matter what we decide regarding the committee keys, the signature of certificates will always require ed25519 keys, so in this case the flow is the same, just the returned TallyProof is private.

chain-impl-mockchain/src/vote/payload.rs
danielSanchezQ: Thats a good idea. How would we control the size of the vec in that case? Could we implement it for some `[Payload; T]`?
What do you think?

eugene-babichenko: That's a bit more complicated. AFAIK Rust will require you to implement it for all lengths you want, like `[Payload; 3]`, `[Payload; 5]`, since const generics are not there yet.

chain-impl-mockchain/src/vote/payload.rs
eugene-babichenko: Ah, understood, I remembered that. Then it is better to make this less confusing, like having a named field called `encrypted_vote_parts` or something like that.

chain-impl-mockchain/src/vote/payload.rs
eugene-babichenko: You could always write a macro for that though.

danielSanchezQ: So the thing is that an arbitrary size of votes may be overkilling IMO

chain-impl-mockchain/src/vote/payload.rs
rinor: Furthermore the `Proof` can be used from the node receiving the VoteCast transaction to check if indeed the correct `ElectionPublicKey` (EncryptingVoteKey) was used and reject the transaction if not.

chain-impl-mockchain/src/vote/payload.rs
eugene-babichenko: Strictly speaking, I don't think think that it is an overkill. It's just the language preventing us from expressing this in some nice way.

chain-impl-mockchain/src/vote/payload.rs
enzoc4: I don't know the details, but I though you needed the proof to validate that someone didn't vote for two options, or that the vector is actually unitary. That seems like something that each participant in the network should be able to validate.

chain-impl-mockchain/src/vote/payload.rs
danielSanchezQ: Well, the type itself mask that already I think.

chain-impl-mockchain/src/vote/payload.rs
danielSanchezQ: Yeah, but I mean. I don't feel comfortable not controlling what is the size of the sample.

chain-impl-mockchain/src/vote/payload.rs
danielSanchezQ: Yes, we should do that indeed. Where should I include that check? Can @eugene-babichenko or @enzoc4 point into some snippet please?

chain-impl-mockchain/src/vote/payload.rs
enzoc4: I don't know how is this used, but for a single instance, you could define a constant/static array of encryption keys, and sample from those. Maybe with a non-uniform distribution to make it probable to get many payloads signed with the same one. 
(On a side note, I always find easier to work with `proptest`, as you can just pass parameters).

chain-impl-mockchain/src/vote/payload.rs
eugene-babichenko: Probably manager is the right place https://github.com/input-output-hk/chain-libs/blob/master/chain-impl-mockchain/src/vote/manager.rs#L106

chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -109,6 +113,18 @@ impl ProposalManager {
...
+        match &payload {
+            Payload::Public { .. } => {}
+            Payload::Private {
+                encrypted_vote,
+                proof,
+            } => {
+                // TODO: verify private vote. Need committee public member keys
+                // if !chain_vote::verify_vote(, encrypted_vote, proof) {
+                //     return Err(VoteError::VoteVerificationError)
+                // }
```
danielSanchezQ, line 125: @eugene-babichenko Any idea how would we have the committee member public key here?

chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -410,6 +479,24 @@ impl VotePlanManager {
...
+                    self.plan.committee_member_public_keys(),
+                );
+                if !chain_vote::verify_vote(&pk, encrypted_vote, proof) {
+                    Err(VoteError::VoteVerificationError)
+                } else {
+                    Ok(())
+                }
+            }
+        } {
+            Err(e)
```
danielSanchezQ, line 499: Is it ok to check it here?

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -1140,6 +1155,58 @@ impl Ledger {
...
+        //             if self.governance.parameters.logs_register(action).is_err() {
+        //                 unimplemented!("the action was already recorded for this epoch")
+        //             } else {
+        //                 // nothing
+        //             }
+        //         }
+        //     }
+        // }
+
+        Ok(self)
```
danielSanchezQ, line 1207: I don't really know what to do here if not just verifying it. @eugene-babichenko , @rinor

eugene-babichenko: You need to build the encrypted tally state and store it (I guess somewhere in the proposal manager).

Weights should be retrieved just like we do that for public tally.

https://github.com/input-output-hk/chain-libs/blob/master/chain-vote/src/lib.rs#L97

danielSanchezQ: @eugene-babichenko , where are we retrieving the weights for the public tally?

danielSanchezQ: How about the actions? Just to confirm but we really do not need them for the private tally right? @rinor

eugene-babichenko: Yeah, actions should be triggered on VoteTally certificate. Probably we should call this one VoteEncryptedTally to avoid confusion.

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -1140,6 +1170,51 @@ impl Ledger {
...
+    ) -> Result<Self, Error> {
+        if sig.verify(tally.tally_type(), bad) == Verification::Failed {
+            return Err(Error::VoteTallyProofFailed);
+        }
+
+        let _stake = StakeControl::new_with(&self.accounts, &self.utxos);
+
+        let mut _f = |_action: &VoteAction| ();
+
+        // TODO: Decrypt shares here
```
danielSanchezQ, line 1213: I'm missing how and where should we decrypt the shares.

chain-impl-mockchain/src/testing/builders/tx_cert_builder.rs
```diff
@@ -200,17 +225,58 @@ pub fn tally_sign(
...
+    let auth_data = builder.get_auth_data();
+    let signature = SingleAccountBindingSignature::new(&auth_data, |d| key.sign_slice(&d.0));
+
+    match payload_type {
+        PayloadType::Public => unreachable!("Encrypted vote tally payload should never be public"),
+        PayloadType::Private => TallyProof::Private {
+            id,
+            signature,
+            // TODO: how to get the shares here?
+            shares: vec![],
```
danielSanchezQ, line 278: So, how to have the shares within the private payload?

chain-impl-mockchain/src/ledger/ledger.rs
eugene-babichenko: We should not really decrypt them. Just feed them and the tally state (IIRC it is the same for all users) to the `result` function. https://github.com/input-output-hk/chain-libs/blob/master/chain-vote/src/lib.rs#L129

chain-impl-mockchain/src/testing/builders/tx_cert_builder.rs
eugene-babichenko: I think this function should just accept them as bytes. Is there a problem with doing that?

chain-impl-mockchain/src/certificate/encrypted_vote_tally.rs
```diff
@@ -0,0 +1,149 @@
...
+#[derive(Debug, Clone)]
+pub struct EncryptedVoteTallyProof {
+    pub id: CommitteeId,
+    pub signature: SingleAccountBindingSignature,
+}
+
+#[derive(Debug, Eq, PartialEq, Hash, Clone)]
+pub struct EncryptedVoteTally {
+    id: VotePlanId,
+    payload: VoteTallyPayload,
```
enzoc4, line 24: do we really need this here? isn't it a given that the vote should be private? (`VoteTallyPayload` doesn't seem to have any more information).

chain-impl-mockchain/src/certificate/private_vote_tally.rs
```diff
@@ -0,0 +1,130 @@
...
+    vote::{PayloadType, TryFromIntError},
+};
+use chain_core::{
+    mempack::{ReadBuf, ReadError, Readable},
+    property,
+};
+use typed_bytes::{ByteArray, ByteBuilder};
+
+#[derive(Debug, Eq, PartialEq, Hash, Clone)]
+pub struct PrivateVoteTally {
```
enzoc4, line 14: I think this could've been just `VoteTally` where `VoteTallyPayload` is
```rust
enum VoteTallyPayload {
    Public,
    Private { shares: Vec<Vec<chain_vote::TallyDecryptShare>> },
}
```
Although maybe I lost some part of the discussion. But again, what's the point of `VoteTallyPayload` if it can only be `Private`?

chain-impl-mockchain/src/certificate/encrypted_vote_tally.rs
eugene-babichenko: No we don't.

mzabaluev:
Superseded by #466.

PR #461, opened 2020-10-06
add ElectionPublicKey from_raw

enzoc4:
We need this to pass `ElectionPublicKey` through ffi in the wallet-libs/plugin. (`gargamel::PublicKey` has a from_bytes method)

The reason we need this is that (last time I checked) for casting a private vote, we would fetch the encryption key for the given proposal (from the vit servicing station), and not each one of the individual member public keys.

I think the internal representation is explicitly hidden for a reason, though, so maybe there is a better way to do this.


chain-vote/src/committee.rs
```diff
@@ -29,6 +29,11 @@ impl ElectionPublicKey {
     pub fn as_raw(&self) -> &PublicKey {
         &self.0
     }
+
+    #[doc(hidden)]
+    pub fn from_raw(pk: PublicKey) -> ElectionPublicKey {
```
mzabaluev, line 34: Why does it need to be hidden and use scary "raw" naming?
Is it dangerous to designate any public key as an `ElectionPublicKey` by conversion?

danielSanchezQ: I think it would be better to use a plain `from_bytes` method and not use the `publickey` directly as a parameter:

```rust
    pub fn from_bytes(buf : &[u8]) -> Option<Self> {
        PublicKey::from_bytes(buf).map(ElectionPublicKey)
    }
```

mzabaluev: That'll do. Maybe even a descriptive error to return instead of `None` on failure.

enzoc4: >  Is it dangerous to designate any public key as an ElectionPublicKey by conversion?

Well, only if by mistake you encrypt your vote with only one of the keys... But the same thing can be achieved with using only one key in `from_participants`.

But yeah, `as_raw` was a bad name, tbh I just "copied" the `as_raw` method and I hadn't noticed the other `impl` block at the end of the file.

I'll go with `from_bytes`, I agree it's better. I went through the public key just because that's what was done in `jcli` for the `MemberCommunicationPublicKey`, but it's probably not the same.

mzabaluev (approved):

PR #465, opened 2020-10-09
Chain vote serialization

eugene-babichenko:


chain-vote/src/committee.rs
```diff
@@ -124,6 +124,11 @@ impl MemberSecretKey {
     pub fn to_bytes(&self) -> [u8; 32] {
         self.0.sk.to_bytes()
     }
+
+    pub fn from_bytes(bytes: &[u8; 32]) -> Option<Self> {
```
mzabaluev, line 128: I think it's more convenient to take a dynamically-sized slice, like in the other places.

And all these places badly need error reporting. `Option` is a wrong way to indicate bad input data.

chain-vote/src/gargamel.rs
```diff
@@ -67,16 +67,16 @@ impl Ciphertext {
         }
     }
 
-    pub fn to_bytes(&self) -> Vec<u8> {
-        let mut r = self.e1.to_bytes().to_vec();
-        r.extend_from_slice(&self.e2.to_bytes());
+    pub fn to_bytes(&self) -> [u8; 130] {
```
mzabaluev, line 70: This may be too much to return on stack. I'd rather kept it a `Vec`. This also gets rid of a magic size in the API.

chain-vote/src/lib.rs
```diff
@@ -124,6 +124,70 @@ impl Tally {
...
+    }
+
+    pub fn from_bytes(bytes: &[u8]) -> Option<Self> {
+        group_elements_from_bytes(bytes).map(|r2s| Self { r2s })
+    }
+}
+
+fn group_elements_to_bytes(elements: &[gang::GroupElement]) -> Vec<u8> {
+    use std::io::Write;
+    let mut bytes: Vec<u8> = Vec::with_capacity(65 * elements.len());
```
mzabaluev, line 173: I don't like this magic number everywhere... Is it fixed in the algorithm and can never change later?

eugene-babichenko: Yeah, size of `GroupElement` is fixed size. I think we can have an associated constant for it.

PR #466, opened 2020-10-13
Private voting

eugene-babichenko:
Includes changes from (I cannot push there) https://github.com/input-output-hk/chain-libs/pull/457


rinor:
don't we need `committee_member_public_keys` also here
https://github.com/input-output-hk/chain-libs/blob/c97d49669f49379b89c7c84467a897809b33fc06/chain-impl-mockchain/src/vote/status.rs#L10-L17

chain-impl-mockchain/src/certificate/vote_tally.rs
```diff
@@ -22,6 +23,9 @@ pub struct VoteTally {
 #[derive(Debug, Eq, PartialEq, Hash, Clone)]
 pub enum VoteTallyPayload {
     Public,
+    Private {
+        shares: Box<[Box<[TallyDecryptShare]>]>,
```
mzabaluev, line 27: Should there be a more descriptive newtype for the first level of boxing?

chain-impl-mockchain/src/certificate/vote_tally.rs
```diff
@@ -56,9 +66,28 @@ impl VoteTally {
         self.payload.payload_type()
     }
 
+    pub fn decryption_shares(&self) -> Option<Box<[Box<[TallyDecryptShare]>]>> {
```
mzabaluev, line 69: I think there should be a newtype like said above, this return type signature is painful to read and it's unclear what the array dimensions mean.

mzabaluev: Something like this:

```rust
pub struct VoteDecryptedShares {
    inner: Box<[Box<[TallyDecryptShare]>],
}

impl VoteDecryptedShares {
    pub fn proposal_shares(&self, proposal_id: u8) -> &[TallyDecryptShare] {
        &self.inner[proposal_id as usize]
    }
}
```

chain-impl-mockchain/src/certificate/vote_plan.rs
```diff
@@ -427,12 +443,22 @@ impl Readable for VotePlan {
             proposals.proposals.push(proposal);
         }
 
+        let member_keys_len = buf.get_u64()?;
+        let mut committee_public_keys = Vec::new();
+        for _ in 0..member_keys_len {
+            let key_len = buf.get_u64()?;
+            let key_buf = buf.get_slice(key_len as usize)?;
+            // Unwrap should be ok here, since we did serialize it ourselves
+            committee_public_keys.push(MemberPublicKey::from_bytes(key_buf).unwrap());
```
mzabaluev, line 452: This is a validating reader from storage format, and if the length is wrong it will blow up. I don't think it's OK.

chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -340,12 +452,20 @@ impl VotePlanManager {
             })
             .collect();
 
+        let committee_public_keys = self
+            .plan()
+            .committee_public_keys()
+            .iter()
+            .cloned()
+            .collect();
```
mzabaluev, line 460: Clippy is right, `to_vec` is better.

chain-impl-mockchain/src/vote/tally.rs
```diff
@@ -21,15 +21,38 @@ pub struct TallyResult {
...
 #[derive(Debug, Error, PartialEq, Eq, Clone)]
 pub enum TallyError {
     #[error("Invalid option choice")]
     InvalidChoice { options: Options, choice: Choice },
+    #[error("Invalid privacy")]
+    InvalidPrivacy,
+    #[error("Choice number in private vote was too high. Maximum of 255 is allowed.")]
+    InvalidPrivateChoiceSize {
+        #[from]
+        source: std::num::TryFromIntError,
```
mzabaluev, line 50: nitpick: this is too baroque, an integer member would be lighter and you could use it in the display message.

chain-impl-mockchain/src/certificate/vote_tally.rs
```diff
@@ -56,9 +69,27 @@ impl VoteTally {
         self.payload.payload_type()
     }
 
+    pub fn decryption_shares(&self) -> Option<&TallyDecryptShares> {
```
mzabaluev, line 72: Bikeshed: is it "decrypted shares", "decrypt-shares", or "decryption shares"? We should settle on the most meaningful naming variant and use it consistently for greppability.

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -1140,6 +1157,32 @@ impl Ledger {
         Ok(self)
     }
 
+    pub fn apply_encrypted_vote_tally<'a>(
+        mut self,
+        tally: &certificate::EncryptedVoteTally,
+        bad: &TransactionBindingAuthData<'a>,
```
mzabaluev, line 1163: Why are these double pointers passed everywhere? The structure is already a thin wrapper around a reference.

chain-impl-mockchain/src/ledger/recovery.rs
```diff
@@ -936,6 +936,33 @@ fn unpack_payload_type<R: std::io::BufRead>(
...
 }
 
+fn pack_committee_public_keys<W: std::io::Write>(
+    keys: &[chain_vote::MemberPublicKey],
+    codec: &mut Codec<W>,
+) -> Result<(), std::io::Error> {
+    codec.put_u64(keys.len() as u64)?;
+    for k in keys {
+        let buf = k.to_bytes();
+        codec.put_u64(buf.len() as u64)?;
```
mzabaluev, line 946: Here as well, we can safely assume that key sizes are smaller than 4 GB.

chain-impl-mockchain/src/ledger/ledger.rs
mzabaluev: This is not necessarily to be addressed immediately, just a style remark.

chain-impl-mockchain/src/certificate/vote_plan.rs
```diff
@@ -279,6 +288,12 @@ impl VotePlan {
     }
 
     pub fn serialize_in(&self, bb: ByteBuilder<Self>) -> ByteBuilder<Self> {
+        let mut member_keys_buf: ByteBuilder<u8> = ByteBuilder::new();
+        member_keys_buf = member_keys_buf.u64(self.committee_public_keys.len() as u64);
+        for key in &self.committee_public_keys {
+            let buf = key.to_bytes();
+            member_keys_buf = member_keys_buf.u64(buf.len() as u64).bytes(&buf);
```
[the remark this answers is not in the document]
eugene-babichenko: I'm pretty sure that keys have some constant size.

chain-vote/src/shvzk.rs
```diff
@@ -32,30 +34,125 @@ impl ABCD {
...
-#[derive(Clone)]
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
 pub struct Proof {
     ibas: Vec<IBA>,
     ds: Vec<Ciphertext>,
     zwvs: Vec<ZWV>,
     r: Scalar,
 }
 
+impl IBA {
```
eugene-babichenko, line 61: All serialization from here should be reviewed carefully. For many `chain-vote` structures it is possible to have constant length buffers.

chain-impl-mockchain/src/ledger/recovery.rs
```diff
@@ -936,6 +936,31 @@ fn unpack_payload_type<R: std::io::BufRead>(
...
+
+fn unpack_committee_public_keys<R: std::io::BufRead>(
+    codec: &mut Codec<R>,
+) -> Result<Vec<chain_vote::MemberPublicKey>, std::io::Error> {
+    let size = codec.get_u8()?;
+    let mut result = Vec::new();
+    for _ in 0..size {
+        let buf = codec.get_bytes(chain_vote::MEMBER_PUBLIC_KEY_BYTES_LEN)?;
+        // It should be safe to unpack here since this is only used with our custom packing too.
+        result.push(chain_vote::MemberPublicKey::from_bytes(&buf).unwrap());
```
mzabaluev, line 959: This will blow up if the first byte in `buf` is not 0x4. Should we trust the ledger on-disk to be correct?

chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -158,6 +171,88 @@ impl ProposalManager {
         })
     }
 
+    #[must_use = "Compute the PrivateTally in a new ProposalManager, does not modify self"]
+    pub fn private_tally(&self, stake: &StakeControl) -> Result<Self, VoteError> {
```
mzabaluev, line 175: Would it make sense to move `self` into this and other similar methods below?
Is the extant `ProposalValue` instance of any use, so that all the cloning is necessary?

chain-impl-mockchain/src/vote/payload.rs
```diff
@@ -87,15 +136,43 @@ mod tests {
...
+            use rand_core::SeedableRng;
+
             match PayloadType::arbitrary(g) {
                 PayloadType::Public => Payload::public(Choice::arbitrary(g)),
+                PayloadType::Private => {
+                    let mut seed = [0u8; 32];
+                    g.fill_bytes(&mut seed);
+                    let mut gen = rand_chacha::ChaCha20Rng::from_seed(seed);
+                    let mc = MemberCommunicationKey::new(&mut gen);
+                    let threshold = 1;
```
mzabaluev, line 162: Is this fixed in all practical uses, or could there be more variation?

chain-impl-mockchain/src/vote/payload.rs
```diff
@@ -87,15 +136,43 @@ mod tests {
...
+                    let h = CRS::random(&mut gen);
+                    let m = MemberState::new(&mut gen, threshold, &h, &[mc.to_public()], 0);
+                    let participants = vec![m.public_key()];
+                    let ek = EncryptingVoteKey::from_participants(&participants);
+                    let vote_options = 3;
```
mzabaluev, line 167: Same question as above.

chain-vote/src/lib.rs
```diff
@@ -250,7 +289,10 @@ pub fn result(
             }
         }
     }
-    TallyResult { votes }
+    TallyResult {
+        votes,
+        options: (0..options as u8),
```
danielSanchezQ, line 294: @mzabaluev I'm not really happy about this conversion (it comes as an `usize`). Any suggestion?

mzabaluev: `TallyState` tabulates per proposal, so it's appropriate. May want to add an assertion, though.

mzabaluev: I have changed the options range type to `usize` for consistency throughout chain-vote.
The limits on the range of choices should be applied in mockchain.

chain-vote/src/gang/p256k1.rs
```diff
@@ -92,7 +92,7 @@ impl GroupElement {
         let gen = GroupElement::generator();
         let mut r = &gen * Scalar::one();
 
-        for _ in 1..table_size + 1 {
+        for _ in 0..table_size {
```
danielSanchezQ, line 95: O__O

mzabaluev (approved):

PR #469, opened 2020-10-14
Expose more vote keys conversions

enzoc4:
See: https://github.com/input-output-hk/jormungandr/pull/2603

Add a few more `from` methods.

PD: I'm not sure if this is a good time to start introducing an `Error` type for all this conversions. It may be better to stabilize things a bit more?

### DIFF
--- a/chain-vote/src/committee.rs
+++ b/chain-vote/src/committee.rs
@@ -135,6 +135,10 @@ impl MemberPublicKey {
     pub fn to_bytes(&self) -> Vec<u8> {
         self.0.to_bytes()
     }
+
+    pub fn from_public_key(pk: PublicKey) -> Self {
+        MemberPublicKey(pk)
+    }
 }
 
 impl MemberCommunicationKey {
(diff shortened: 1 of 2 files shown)


chain-vote/src/committee.rs
```diff
@@ -135,6 +135,10 @@ impl MemberPublicKey {
     pub fn to_bytes(&self) -> Vec<u8> {
         self.0.to_bytes()
     }
+
+    pub fn from_public_key(pk: PublicKey) -> Self {
```
mzabaluev, line 139: This could be just a `From<PublicKey>` impl, so that you could do `pk.into()`.
Why is `MemberPublicKey` a newtype, anyway?

chain-vote/src/gargamel.rs
```diff
@@ -43,6 +43,12 @@ impl SecretKey {
         let sk = Scalar::random(rng);
         Self { sk }
     }
+
+    pub fn from_bytes(bytes: &[u8]) -> Option<Self> {
+        use std::convert::TryInto as _;
+        let bytes: [u8; 32] = bytes.try_into().ok()?;
+        Scalar::from_bytes(&bytes).map(|sk| Self { sk })
```
mzabaluev, line 50: You can use `Scalar::from_slice` so that you don't need an intermediate array.

mzabaluev (changes requested):

mzabaluev:
> PD: I'm not sure if this is a good time to start introducing an Error type for all this conversions. It may be better to stabilize things a bit more?

I agree. We should do it everywhere consistently.

mzabaluev (approved):

PR #480, opened 2020-11-02
Expose serialization API for ProofOfCorrectVote

mzabaluev:
No need to implement it another time in jormungandr.
Also make the serialization of proof self-contained by including the length.
As the number of options in mockchain can never exceed 255, control for this
and use an `u8` for the length field.

### DIFF
--- a/chain-impl-mockchain/src/vote/payload.rs
+++ b/chain-impl-mockchain/src/vote/payload.rs
@@ -1,11 +1,11 @@
 use crate::vote::Choice;
 use chain_core::mempack::{ReadBuf, ReadError};
 use chain_vote::shvzk;
-use chain_vote::{Ciphertext, EncryptedVote, ProofOfCorrectVote, Scalar};
+use chain_vote::{Ciphertext, Scalar};
 use std::convert::{TryFrom, TryInto as _};
 use std::hash::Hash;
 use thiserror::Error;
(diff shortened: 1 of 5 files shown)


chain-impl-mockchain/src/vote/payload.rs
```diff
@@ -140,6 +113,67 @@ impl Payload {
...
+            .fold(self.0.ds(), |bb, d| bb.bytes(&d.to_bytes()))
+            .fold(self.0.zwvs(), |bb, zwv| bb.bytes(&zwv.to_bytes()))
+            .bytes(&self.0.r().to_bytes())
+    }
+
+    pub fn serialize(&self) -> ByteArray<Self> {
+        self.serialize_in(ByteBuilder::new()).finalize()
+    }
+
+    pub(crate) fn read<'a>(buf: &mut ReadBuf<'a>) -> Result<Self, ReadError> {
```
eugene-babichenko, line 143: Why `pub(crate)` here?

eugene-babichenko (approved):

chain-impl-mockchain/src/vote/payload.rs
mzabaluev: There is no need for this method outside of the crate yet. Can be made public when we need it.

PR #481, opened 2020-11-04
Implemented encrypt_vote wrapper

danielSanchezQ:


chain-impl-mockchain/src/vote/privacy.rs
```diff
@@ -0,0 +1,13 @@
+use crate::vote::ProofOfCorrectVote;
+use chain_vote::{EncryptedVote, EncryptingVoteKey, Vote};
+use rand_core::{CryptoRng, RngCore};
+
+#[allow(dead_code)]
+pub fn encrypt_vote<R: RngCore + CryptoRng>(
+    rng: &mut R,
+    public_key: &EncryptingVoteKey,
+    vote: Vote,
+) -> (EncryptedVote, ProofOfCorrectVote) {
```
mzabaluev, line 10: May want to `pub use chain_vote::EncryptedVote` in mod `vote` so that all types used here can be imported from mockchain.

mzabaluev (approved):

Issue #479, opened 2020-11-02
Fix warnings


mzabaluev:
Our CI considers clippy issues a failure, but lets compiler warnings be. This does not make sense and has bearing on code quality.
We need to go through all warnings, fix them, and then set all warnings to deny in CI builds.

mzabaluev:
Done in #483

PR #483, opened 2020-11-06
Clippy warnings rush

danielSanchezQ:
Part of #479

### DIFF
--- a/chain-impl-mockchain/src/testing/arbitrary/address.rs
+++ b/chain-impl-mockchain/src/testing/arbitrary/address.rs
@@ -95,31 +95,24 @@ impl ArbitraryAddressDataValueVec {
         self.0
             .iter()
             .cloned()
-            .filter(|x| match x.address_data.kind() {
-                Kind::Single { .. } => true,
-                _ => false,
-            })
+            .filter(|x| matches!(x.address_data.kind(), Kind::Single { .. }))
             .collect()
     }
+    #[allow(clippy::match_like_matches_macro)]
     pub fn accounts(&self) -> Vec<AddressDataValue> {
         self.0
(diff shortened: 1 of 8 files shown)


chain-impl-mockchain/src/header/test.rs
```diff
@@ -25,7 +25,7 @@ impl Arbitrary for BlockVersion {
 
 impl Arbitrary for AnyBlockVersion {
     fn arbitrary<G: Gen>(g: &mut G) -> Self {
-        match g.next_u32() % 1 {
+        match g.next_u32() % 2 {
             0 => AnyBlockVersion::Supported(BlockVersion::from_u16(u16::arbitrary(g) % 3).unwrap()),
             1 => AnyBlockVersion::Unsupported(u16::arbitrary(g)),
             _ => unreachable!(),
```
danielSanchezQ, line 31: This was more a bug than a warning though.

chain-impl-mockchain/src/testing/arbitrary/address.rs
```diff
@@ -95,31 +95,24 @@ impl ArbitraryAddressDataValueVec {
...
             .iter()
             .cloned()
-            .filter(|x| match x.address_data.kind() {
-                Kind::Single { .. } => true,
-                _ => false,
-            })
+            .filter(|x| matches!(x.address_data.kind(), Kind::Single { .. }))
             .collect()
     }
+    #[allow(clippy::match_like_matches_macro)]
```
mzabaluev: These are redundant because you have changed the expressions to use `matches!`. I have removed these overrides in another commit.

mzabaluev (approved):

PR #485, opened 2020-11-12
use SmallRng instead of ChaCha for generating test keys

eugene-babichenko:
Use a faster non-cryptographic generators for generating `Arbitrary` keys.

### DIFF
--- a/chain-crypto/src/testing.rs
+++ b/chain-crypto/src/testing.rs
@@ -2,8 +2,8 @@ use super::*;
 use crate::digest;
 
 use quickcheck::{Arbitrary, Gen};
-use rand_chacha::ChaChaRng;
-use rand_core::SeedableRng;
+use rand_core::{SeedableRng, RngCore, CryptoRng, Error as RngError};
+use rand::rngs::SmallRng;
 
 /// an Arbitrary friendly cryptographic generator
 ///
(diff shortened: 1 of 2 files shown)


eugene-babichenko:
`test_release` passes in less than 2 minutes now so this is an obvious improvement.

I wonder if we actually care about reproducible cryptographic keys? If not, we can just go with thread-local RNGs seeded from the system entropy, but this will require a significant refactoring. I tried to do that and there was no easy way to incorporate this change.

chain-crypto/src/testing.rs
```diff
@@ -15,6 +15,36 @@ use rand_core::SeedableRng;
 #[derive(Clone, Debug)]
 pub struct TestCryptoGen(pub u64);
 
+pub struct TestCryptoRng(SmallRng);
```
mzabaluev, line 18: Please document that this is a fake crypto RNG that does not really provide cryptograhic safety.

mzabaluev (approved):

mzabaluev:
> I wonder if we actually care about reproducible cryptographic keys?

If a test fails, we'd like it to be reproducible by reusing the seed. This may be not how it works right now, but in principle all generators used in tests should be seedable. By default the seed should be initialized from entropy and printed. This may need to be done for every test case.

chain-crypto/src/testing.rs
eugene-babichenko: Done.

PR #488, opened 2020-11-17
chain-network: add PullBlocks endpoint to request a range of blocks

eugene-babichenko:


chain-network/proto/node.proto
```diff
@@ -104,6 +104,14 @@ message PullBlocksToTipRequest {
...
 }
 
+// Request message for method PullBlocks
+message PullBlocksRequest {
+  // The identifiers of a starting block.
+  bytes from = 1;
+  // The identifier of the end block
+  bytes to = 2;
+}
+
```
mzabaluev, line 114: This is confusingly different from `PullHeadersRequest` and `PullBlocksToTipRequest`: the `from` field in those requests is repeated, meaning to give a list of checkpoints.

mzabaluev (commented):
When we designed the pull protocol, it looked like this: the chain of block headers is retrieved with `PullHeaders` and pre-validated first, then `GetBlocks` requests are sent with lists of already known block IDs, batched and possibly distributed among recently active peers. Back then, it was faster to fetch blocks from storage by IDs. Is it different now?

eugene-babichenko:
Regarding the performance part.

Given the current implementation of finding ancestor in the storage it is different. Most of the time determining if `from` is an ancestor of `to` can be done in a rather fast way with only two requests to storage. Moving forward, the implementation of the storage iterator is such that it makes use of the permanent storage where most of the blocks should be stored. When using the permanent storage, we just start from the given block length and move through all blocks in a linear fashion. We don't even need to know their IDs in advance in case of permanent storage, so for most ranges there will be one less step to start iterating.

I started implementing this request with a performance concern in mind too. When doing parallel block pulls, we will need to request large chunks of blocks from different nodes (say, 512 blocks from each known node at a time) and I thought that these requests will get too big: 16 KiB (not counting the Protobuf overhead) per request if we use 512 blocks per batch.

Anyway, I can try doing a benchmark to compare the two and to verify my assumptions.

chain-network/proto/node.proto
eugene-babichenko: Will `from_id` work here?

eugene-babichenko:
Added benchmarks, they look promising https://github.com/input-output-hk/chain-libs/pull/489

chain-network/proto/node.proto
mzabaluev: I'd like to keep all similar requests with homogeneous parameter fields.
That said, a non-repeated field in protobuf can be converted to repeated [in a backward compatible way][protobuf-updating] (the doc talks about optional fields, which means any field in proto3). So maybe we can keep this `from` bound as a non-repeating field for now. It will even work correctly (but wastefully) if, in a future update, a client sends a repeating `from` field with the most remote checkpoint as the last member, but the server replaces all repeating values with the last one it receives.

Another option is to just make the field repeating now and implement the method the same way as we do `PullHeaders`. We just won't use multiple `from` checkpoints in our node, but the protocol will look consistent.

[protobuf-updating]: https://developers.google.com/protocol-buffers/docs/proto#updating

mzabaluev:
To think more of this, we can deprecate `PullBlocksToTip` due to server-side race/contention issues due to the tip updating while the request is being streamed (https://github.com/input-output-hk/jormungandr/issues/1530), and tell both simple and advanced protocol clients to bootstrap using `PullBlocks` with a list or a single known starting point, respectively.

eugene-babichenko:
That was the initial idea: use this protocol regardless of the number of peers available. Did not about the issue though.

chain-network/proto/node.proto
eugene-babichenko: I am leaning towards the first option. From the practical point of view (request a range of blocks) having a repeated field does not make any sense. To make it less confusing we could just rename the method to `PullBlocksRange` 🤷

mzabaluev (approved):
Looks good. Ignore the lint failures, they've been fixed.

PR #495, opened 2020-12-02
change vote plan error message in ledger

eugene-babichenko:
Resolves https://github.com/input-output-hk/jormungandr/issues/2769

### DIFF
--- a/imhamt/src/hamt.rs
+++ b/imhamt/src/hamt.rs
@@ -5,6 +5,8 @@ use super::node::{
 };
 pub use super::operation::{InsertError, RemoveError, ReplaceError, UpdateError};
 use std::borrow::Borrow;
+use std::error::Error;
+use std::fmt::Debug;
 use std::iter::FromIterator;
 use std::marker::PhantomData;
 use std::mem::swap;
@@ -130,7 +132,7 @@ impl<H: Hasher + Default, K: Eq + Hash + Clone, V: Clone> Hamt<H, K, V> {
     pub fn update<F, U>(&self, k: &K, f: F) -> Result<Self, UpdateError<U>>
     where
         F: FnOnce(&V) -> Result<Option<V>, U>,
-        U: std::error::Error + std::fmt::Debug + 'static,
+        U: Error + Debug + 'static,
     {
         let h = HashedKey::compute(self.hasher, &k);
         let newroot = update_rec(&self.root, h, 0, k, f)?;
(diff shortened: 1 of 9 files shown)


imhamt/src/hamt.rs
```diff
@@ -130,6 +130,7 @@ impl<H: Hasher + Default, K: Eq + Hash + Clone, V: Clone> Hamt<H, K, V> {
     pub fn update<F, U>(&self, k: &K, f: F) -> Result<Self, UpdateError<U>>
     where
         F: FnOnce(&V) -> Result<Option<V>, U>,
+        U: std::error::Error + std::fmt::Debug + 'static,
```
mzabaluev, line 133: Could add use statements at the top rather than absolute paths everywhere.

mzabaluev: Why are these extra bounds needed here? They don't look essential to these methods.

imhamt/src/lib.rs
```diff
@@ -139,7 +140,7 @@ mod tests {
     }
 
     #[allow(clippy::trivially_copy_pass_by_ref)]
-    fn next_u32(x: &u32) -> Result<Option<u32>, ()> {
+    fn next_u32(x: &u32) -> Result<Option<u32>, Infallible> {
```
mzabaluev, line 143: clippy is right by the way, it's pointless to pass an integer by reference.

imhamt/src/operation.rs
```diff
@@ -1,21 +1,32 @@
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+use thiserror::Error;
+
+#[derive(Debug, Error, Clone, Copy, PartialEq, Eq)]
 pub enum InsertError {
+    #[error("entry with the provided key already exists")]
```
mzabaluev, line 5: This is what was missing, I don't think the bounds on generic operations are needed though.

imhamt/src/hamt.rs
eugene-babichenko: `U` is the `#[source]` of an error variant inside of `UpdateError`. To be the `#[source]` an error must be at least `Error + 'static` (see [`Error::source`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source)). Since `UpdateError` is `Debug`, `U` must also be `Debug`.

imhamt/src/lib.rs
eugene-babichenko: Yeah, it's right, but this function is used _only_ for `Hamt::update` and it passes the current value by reference.

imhamt/src/hamt.rs
mzabaluev: Ah, I see.

imhamt/src/lib.rs
mzabaluev: I think we should have been less liberal with throwing those clippy overrides around, even if they only relate to internal code. Nothing major here, though.

mzabaluev (approved):

PR #503, opened 2021-01-05
Bump rand stack

danielSanchezQ:
Update rand, rand_core, and rand_chacha.
We must wait for https://github.com/dalek-cryptography/curve25519-dalek/pull/338
and update that dependency also.


eugene-babichenko:
this is replaced by #532

PR #504, opened 2021-01-06
fix the panic when trying to vote for a non-existent private proposal

eugene-babichenko:


eugene-babichenko:
Clippy errors are unrelated to this pull request. Looks like they were introduced with the toolchain update.

dkijania (approved):

PR #505, opened 2021-01-07
Removed elided lifetimes

danielSanchezQ:
And fixed some clippy warnings.

### DIFF
--- a/chain-impl-mockchain/src/certificate/pool.rs
+++ b/chain-impl-mockchain/src/certificate/pool.rs
@@ -57,6 +57,8 @@ pub struct PoolPermissions(u64);
 pub type ManagementThreshold = u8;
 
 const MANAGEMENT_THRESHOLD_BITMASK: u64 = 0b11_1111; // only support 32, reserved one for later extension if needed
+
+#[allow(clippy::unusual_byte_groupings)]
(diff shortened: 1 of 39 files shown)


chain-impl-mockchain/src/certificate/pool.rs
```diff
@@ -57,6 +57,8 @@ pub struct PoolPermissions(u64);
 pub type ManagementThreshold = u8;
 
 const MANAGEMENT_THRESHOLD_BITMASK: u64 = 0b11_1111; // only support 32, reserved one for later extension if needed
+
+#[allow(clippy::unusual_byte_groupings)]
 const ALL_USED_BITMASK: u64 =
     0b00000000_00000000_00000000_00000000_00000000_00000000_00000000_00111111;
```
mzabaluev, line 63: I wonder what's unusual about this. It may be better to just remove the redundant zeroes and leave the 8 LSBs.

mzabaluev (approved):

chain-impl-mockchain/src/certificate/pool.rs
danielSanchezQ: Grouping by 4 is the Clippy style.  I feel is ok that way, completely explicit about what is being used.

PR #512, opened 2021-01-21
Update tonic and prost for tokio1 support

danielSanchezQ:


mzabaluev (commented):
Simplest API version upgrade in recent memory.

PR #516, opened 2021-01-26
use baby-step giant-step to speed up tallying

Zeegomo:
Use baby-step giant-step algorithm (https://en.wikipedia.org/wiki/Baby-step_giant-step) to speed up the tallying, memory and time O(sqrt(MAX_VOTES)). With this change we can tally a fund2-like vote plan in 3s.

To utilize this efficiently we should add a command in jcli to do the tallying for all the proposals in a vote plan.

### DIFF
--- a/chain-vote/src/gang/babystep.rs
+++ b/chain-vote/src/gang/babystep.rs
@@ -4,26 +4,32 @@ use super::p256k1::*;
 use super::zerocaf::*;
 use lazy_static::*;
 use std::collections::HashMap;
+use std::convert::TryFrom;
 
+// make steps asymmetric, in order to better use caching of baby steps.
+// balance of 2 means that baby steps are 2 time more than sqrt(MAX_VOTES_BSGS)
+const BALANCE: u64 = 2;
 // change this to adjust precomputed range, given this number it is possible to
 // find mupltiples up to SQRT_STEP_SIZE ^ 2 - 1
 lazy_static! {
-    pub(crate) static ref SQRT_STEP_SIZE: u64 = {
+    static ref SQRT_STEP_SIZE: u64 = {
         let max_votes = option_env!("MAX_VOTES_BSGS")
-            .unwrap_or("9999999999") // 1e10 -1
+            .unwrap_or("31999999999") // 1e10 -1
             .parse::<u64>()
             .expect("Invalid MAX_VOTES_BSGS: expected a integer");
         ((max_votes as f64).sqrt() + 1.0) as u64
     };
+    static ref BABY_STEPS_SIZE: u64 = *SQRT_STEP_SIZE * BALANCE;
 }
 
 lazy_static! {
-    static ref BS: HashMap<GroupElement, u64> = {
+    static ref BS: HashMap<[u8; 32], u64> = {
         let mut bs = HashMap::new();
         let gen = GroupElement::generator();
         let mut e = GroupElement::zero();
-        for i in 0..*SQRT_STEP_SIZE {
-            bs.insert(e.clone(), i);
+        // use negation trick and store only the x coordinate
+        for i in 0..*BABY_STEPS_SIZE / 2 {
+            bs.insert(<[u8; 32]>::try_from(&e.to_bytes()[1..33]).unwrap(), i);
             e = &e + &gen;
         }
         bs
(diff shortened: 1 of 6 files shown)


chain-vote/src/gang/babystep.rs
```diff
@@ -0,0 +1,107 @@
...
+// change this to adjust precomputed range, given this number it is possible to
+// find mupltiples up to SQRT_STEP_SIZE ^ 2 - 1
+lazy_static! {
+    static ref SQRT_STEP_SIZE: u64 = {
+        let max_votes = option_env!("MAX_VOTES_BSGS")
+            .unwrap_or("31999999999") // 1e10 -1
+            .parse::<u64>()
+            .expect("Invalid MAX_VOTES_BSGS: expected a integer");
+        ((max_votes as f64).sqrt() + 1.0) as u64
+    };
```
mzabaluev, line 22: This is a lot of hidden parameterization for a library.

I wonder if we could make this table an explicit object in the public API of chain-vote, that you generate at a place and with parameters of your choosing and pass to the `result` function (which should be renamed `tally_result` to emphasize that it does a lot of work and is not just some trivial data getter)?

mzabaluev: > with parameters of your choosing

Something like `PrivateTallyTable::new(total_voting_power)` which would get the equivalent of your `MAX_VOTES_BSGS` environment variable.

mzabaluev (changes requested):

chain-vote/src/gang/babystep.rs
danielSanchezQ: Yes, maybe a builder for it (builder pattern or a function). Making the magic numbers some documented constants in the babystep API.

Zeegomo: Agreed, it was simpler when I started building it but then I noticed we need more options and it definitely grew out of its place.

Zeegomo:
I will rebase this once we have a validation function for the online tally check

chain-vote/src/gang/babystep.rs
```diff
@@ -0,0 +1,110 @@
...
+        let gen = GroupElement::generator();
+        let mut e = GroupElement::zero();
+        // use negation trick and store only the x coordinate
+        for i in 0..=baby_step_size / 2 {
+            bs.insert(<[u8; 32]>::try_from(&e.to_bytes()[1..33]).unwrap(), i);
+            e = &e + &gen;
+        }
+        Self {
+            table: bs,
+            baby_step_size: baby_step_size,
```
mzabaluev, line 45: We expect changes to be lint-free because the CI will never stop complaining about this.

chain-vote/src/gang/babystep.rs
```diff
@@ -0,0 +1,110 @@
...
+    pub fn generate_with_balance(max_voting_power: u64, balance: u64) -> Self {
+        let sqrt_step_size = (max_voting_power as f64).sqrt().ceil() as u64;
+        let baby_step_size = sqrt_step_size * balance;
+        let mut bs = HashMap::new();
+        let gen = GroupElement::generator();
+        let mut e = GroupElement::zero();
+        // use negation trick and store only the x coordinate
+        for i in 0..=baby_step_size / 2 {
+            bs.insert(<[u8; 32]>::try_from(&e.to_bytes()[1..33]).unwrap(), i);
+            e = &e + &gen;
```
mzabaluev, line 41: Hmm, no `AddAssign` impl for `GroupElement`?

chain-vote/src/gang/babystep.rs
```diff
@@ -0,0 +1,110 @@
...
+) -> Vec<Option<u64>> {
+    let baby_step_size = table.baby_step_size;
+    let giant_step = &table.giant_step;
+    let table = &table.table;
+    points
+        .into_par_iter()
+        .map(|mut p| {
+            let mut a = 0;
+            loop {
+                if let Some(x) = table.get(&p.to_bytes()[1..33]) {
```
mzabaluev, line 65: This uses secret knowledge about the internal representation of `GroupElement`.
I wonder if we should add some accessors to the coordinates, but eccocide only provides these for affine points.
Cc @vincenthz

chain-vote/src/gang/babystep.rs
```diff
@@ -0,0 +1,110 @@
...
+use std::collections::HashMap;
+use std::convert::TryFrom;
+
+// make steps asymmetric, in order to better use caching of baby steps.
+// balance of 2 means that baby steps are 2 time more than sqrt(MAX_VOTES_BSGS)
+const DEFAULT_BALANCE: u64 = 2;
+
+/// Holds precomputed baby steps for the baby-stap giant-step algorithm
+/// for solving discrete log on ECC and tally private votes
+pub struct PrivateTallyTable {
```
mzabaluev, line 12: Hmm, maybe the internal name in this module should not refer to a private tally, because the rest of `gang` is just group theory math.

chain-vote/src/lib.rs
```diff
@@ -219,11 +219,11 @@ fn group_elements_from_bytes(bytes: &[u8]) -> Option<Vec<gang::GroupElement>> {
     Some(elements)
 }
 
-pub fn result(
```
mzabaluev, line 222: Please also rename the usage in `chain-impl-mockchain`.

mzabaluev (changes requested):

chain-vote/src/lib.rs
Zeegomo: Is it needed though if we plan to replace that with the validation function on the chain? I was waiting for that change before merging this

chain-vote/src/gang/babystep.rs
Zeegomo: We need coordinates of affine points here, maybe something like `pub(super) to_affine_coordinates()`?

mzabaluev (approved):

PR #518, opened 2021-01-28
Add verification of computed private tally; make tallying fallible

### LINKED ISSUE #515: Implement validation function in chain-vote
For on-ledger validation of a modified private vote tally transaction with results computed offline and decryption shares provided as proof, a more lightweight algorithm can be provided.

mzabaluev:
Add a `verify` function to enable quick online verification of a tally that is computed offline and submitted to the ledger along with the decrypt shares certified by the committee.

Resolve the problem of dealing with `None` holes in `TallyResult` by not having holes, but returning an error. The assumption is that any cases where such a result can occur raise from either a programming error with the total vote count, or bad input data.

Resolves #515.

### DIFF
--- a/chain-impl-mockchain/src/vote/manager.rs
+++ b/chain-impl-mockchain/src/vote/manager.rs
@@ -438,7 +438,7 @@ impl ProposalManagers {
         }
     }
 
-    pub fn private_tally_start(&self, stake: &StakeControl) -> Result<Self, VoteError> {
+    pub fn start_private_tally(&self, stake: &StakeControl) -> Result<Self, VoteError> {
         let mut proposals = Vec::with_capacity(self.0.len());
         for proposal in self.0.iter() {
             proposals.push(proposal.private_tally(stake)?);
@@ -447,7 +447,7 @@ impl ProposalManagers {
         Ok(Self(proposals))
     }
 
-    pub fn private_tally_finalize<F>(
+    pub fn finalize_private_tally<F>(
         &self,
         shares: &TallyDecryptShares,
         governance: &Governance,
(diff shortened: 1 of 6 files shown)


mzabaluev:
This is a first naive attempt and needs a test.

chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -237,31 +237,17 @@ impl ProposalManager {
     where
         F: FnMut(&VoteAction),
     {
-        use std::convert::TryInto;
         let tally = self.tally.as_ref().ok_or(TallyError::NoEncryptedTally)?;
         let (encrypted_tally, total_stake) = tally.private_encrypted()?;
         let state = encrypted_tally.state();
         let table_size = total_stake.0 as usize / self.options.choice_range().len();
-        let private_result = chain_vote::result(total_stake.0, table_size, &state, shares);
+        let private_result = chain_vote::tally_result(total_stake.0, table_size, &state, shares)
```
Zeegomo, line 244: The `verify` function will be used here when the necessary info will be in the transaction, right?

mzabaluev: Yes, not changing that in this PR.

Zeegomo (approved):

Zeegomo (approved):
Looks good to me, I wonder if we could enforce a `Tally` to be always valid by turning `tally` and `verify` into constructors for the struct

mzabaluev:
> if we could enforce a Tally to be always valid by turning tally and verify into constructors for the struct

The online verification use case implies getting the tally from serialization and verifying against the tallying input stuff, so I don't think it can be made correct by construction.

PR #521, opened 2021-02-02
[Tests] tally benches

dkijania:
- Generate a number of votes and apply them to the ledger
- Benchmark the creation of the encrypted tally by running it multiple times without saving the result
- Save the result and benchmark the decryption step and applying a decrypted result in the same way

### DIFF
--- a/chain-impl-mockchain/benches/tally.rs
+++ b/chain-impl-mockchain/benches/tally.rs
@@ -4,8 +4,8 @@ use chain_impl_mockchain::{
     fee::LinearFee,
     header::BlockDate,
     testing::{
-        build_tally_decrypt_share,
         data::CommitteeMembersManager,
+        decrypt_tally,
         ledger::ConfigBuilder,
         scenario::{prepare_scenario, proposal, vote_plan, wallet},
         VoteTestGen,
@@ -141,16 +141,18 @@ fn tally_benchmark(c: &mut Criterion) {
         })
         .unwrap();
 
-    c.bench_function("build_tally_decrypt_share", |b| {
+    c.bench_function("decrypt_tally", |b| {
         b.iter(|| {
-            build_tally_decrypt_share(vote_plan_status, &members);
+            decrypt_tally(vote_plan_status, &members);
         })
     });
 
-    let shares = build_tally_decrypt_share(vote_plan_status, &members);
+    let shares = decrypt_tally(vote_plan_status, &members);
 
-    let vote_tally = VoteTally::new_private(vote_plan.to_id(), shares);
-    let fragment = controller.fragment_factory().vote_tally(&alice, vote_tally);
+    let decrypted_tally = VoteTally::new_private(vote_plan.to_id(), shares);
+    let fragment = controller
+        .fragment_factory()
+        .vote_tally(&alice, decrypted_tally);
 
     c.bench_function("vote_tally", |b| {
         b.iter(|| {
--- a/chain-impl-mockchain/benches/tally.rs
+++ b/chain-impl-mockchain/benches/tally.rs
@@ -21,7 +21,6 @@ use rand_core::SeedableRng;
 const ALICE: &str = "Alice";
 const STAKE_POOL: &str = "stake_pool";
 const VOTE_PLAN: &str = "fund1";
-const VOTERS_COUNT: usize = 250;
 
 fn voters_aliases(count: usize) -> Vec<String> {
     let mut counter = 0;
@@ -33,7 +32,7 @@ fn voters_aliases(count: usize) -> Vec<String> {
     .collect()
 }
 
-fn tally_benchmark(c: &mut Criterion) {
+fn tally_benchmark(voters_count: usize, voting_power_per_voter: u64, c: &mut Criterion) {
     const MEMBERS_NO: usize = 3;
     const THRESHOLD: usize = 2;
     let favorable = Choice::new(1);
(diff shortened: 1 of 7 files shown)


chain-impl-mockchain/benches/tally.rs
```diff
@@ -0,0 +1,168 @@
...
+        .vote_encrypted_tally(&alice, encrypted_tally);
+
+    let parameters = ledger.parameters.clone();
+    let date = ledger.date();
+
+    c.bench_function("vote_encrypted_tally", |b| {
+        b.iter(|| {
+            ledger
+                .ledger
+                .apply_fragment(&parameters, &fragment, date)
```
Zeegomo, line 127: It's not super clear at first sight why this does not save the result for me, so maybe add a comment?

chain-impl-mockchain/benches/tally.rs
```diff
@@ -0,0 +1,168 @@
...
+        .owns(STAKE_POOL)
+        .committee_member();
+    wallets.push(&mut alice_wallet_builder);
+
+    let voters_aliases = voters_aliases(VOTERS_COUNT);
+    let mut wallet_builders: Vec<WalletTemplateBuilder> =
+        voters_aliases.iter().map(|alias| wallet(alias)).collect();
+
+    for wallet_builder in wallet_builders.iter_mut() {
+        wallet_builder.with(10);
```
eugene-babichenko, line 55: IMO we should increase stakes to be more realistic. I understand that this is going to slow down the tallying process as it is now, but this should not be the case after offline tallying is merged.

eugene-babichenko (commented):
I think we should postpone this PR a little bit and rebase it when #250 is merged. The tallying process is going to be changed a bit so does this benchmark. Otherwise, this is a good job.

Also, I think this is the time to introduce the repository-wide documentation on running benchmarks, since it is a bit annoying to remember that you should add `--feature with-bench` every single time.

eugene-babichenko:
The build on the latest nightly compiler is affected by a mistake in `cryptoxide`, I filed a PR to fixed that. https://github.com/typed-io/cryptoxide/pull/15

Zeegomo:
> I think we should postpone this PR a little bit and rebase it when #250 is merged. 

you meant #520?

eugene-babichenko:
Yes.

chain-impl-mockchain/benches/tally.rs
```diff
@@ -0,0 +1,193 @@
...
+    });
+
+    ledger.apply_fragment(&fragment, ledger.date()).unwrap();
+}
+
+fn tally_benchmark_128_voters_1000_ada(c: &mut Criterion) {
+    tally_benchmark(128, 1000, c);
+}
+
+fn tally_benchmark_200_voters_1000_ada(c: &mut Criterion) {
```
Zeegomo, line 174: This seems quite similar to the test case above it

Zeegomo (approved):

chain-impl-mockchain/benches/tally.rs
eugene-babichenko: The point was to test with different total voting power. For the previous implementation the voting power made a huge difference.

chain-impl-mockchain/benches/tally.rs
```diff
@@ -0,0 +1,193 @@
...
+        .iter()
+        .find(|c_vote_plan| {
+            let vote_plan: VotePlan = vote_plan.clone().into();
+            c_vote_plan.id == vote_plan.to_id()
+        })
+        .unwrap();
+
+    c.bench_function(&format!("decrypt_tally{}", bench_name_suffix), |b| {
+        b.iter(|| {
+            decrypt_tally(vote_plan_status, &members);
```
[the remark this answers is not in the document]
eugene-babichenko: So what you suggest is to break down this step into producing shares and then producing decrypted tally out of them? Actually, that's a good point, such benchmark would make much more sense.

chain-impl-mockchain/benches/tally.rs
```diff
@@ -0,0 +1,193 @@
...
+    let mut wallet_builders: Vec<WalletTemplateBuilder> =
+        voters_aliases.iter().map(|alias| wallet(alias)).collect();
+
+    for wallet_builder in wallet_builders.iter_mut() {
+        wallet_builder.with(voting_power_per_voter);
+        wallets.push(wallet_builder);
+    }
+
+    let mut rng = ChaCha20Rng::from_seed([0u8; 32]);
+    let members = CommitteeMembersManager::new(&mut rng, THRESHOLD, MEMBERS_NO);
```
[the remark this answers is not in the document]
eugene-babichenko: I do not think that having 1000 members is a realistic scenario to be honest.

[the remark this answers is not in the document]
Zeegomo: Agree, just pointed that out. We can skip this

chain-impl-mockchain/benches/tally.rs
Zeegomo: What I mean is that the number of voters does not influence the decryption speed. The advantage of having two benchmarks with similar values in both voters and total voting power (`128_voters_1000_ada` and `200_voters_1000_ada`) is not much IMHO if we already have  bigger tests like `200_votes_1_000_000_ada` or `1000_voters_1000_ada`, but maybe I am missing something. 

Anyway, this is just a nit.

chain-impl-mockchain/benches/tally.rs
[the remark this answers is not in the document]
Zeegomo: Yep, if we want to have an accurate benchmark, that would be a better representation of the actual workload

chain-impl-mockchain/benches/tally.rs
eugene-babichenko: > What I mean is that the number of voters does not influence the decryption speed

But it does influence the speed of construction of the encrypted tally.

Zeegomo (approved):
Great work!

PR #524, opened 2021-02-11
parallelize private tally computation

eugene-babichenko:


chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -199,29 +199,46 @@ impl ProposalManager {
...
+                            }
+                            vote::Payload::Private {
+                                encrypted_vote,
+                                proof: _,
+                            } => Some(Ok((encrypted_vote.as_inner(), stake.0))),
                         }
+                    } else {
+                        None
                     }
+                } else {
```
Zeegomo, line 228: Nit: you could return from inside the match and avoid else branches

Zeegomo (approved):

PR #525, opened 2021-02-12
Fix clippy 1.50 new lints

eugene-babichenko:


Zeegomo (approved):

Zeegomo (changes requested):
There are some lints still failing

chain-impl-mockchain/src/transaction/payload.rs
```diff
@@ -91,8 +91,8 @@ impl<'a, P: Payload> PayloadAuthSlice<'a, P> {
 }
 
 impl<'a, P: Payload> PayloadSlice<'a, P> {
-    pub fn into_certificate_slice(self) -> Option<CertificateSlice<'a>> {
-        <P as Payload>::to_certificate_slice(self)
+    pub fn inpayload_to_certificate_slice(self) -> Option<CertificateSlice<'a>> {
```
Zeegomo, line 94: Is this intentional or caused by some search and replace :)? I would also keep the `into_certificate_slice` name as it is here

Zeegomo (changes requested):

chain-impl-mockchain/src/transaction/payload.rs
eugene-babichenko: Yes, my bad :D

Zeegomo (approved):

PR #527, opened 2021-02-17
use fake cryptographic PRNG for preparatory work in tally benches

eugene-babichenko:


Zeegomo (approved):

eugene-babichenko:
Will also look into how test accounts are build: may need to change the generator for them as well.

PR #531, opened 2021-02-22
Big Bench

eugene-babichenko:


chain-impl-mockchain/benches/tally.rs
```diff
@@ -201,27 +258,130 @@ fn tally_benchmark(voters_count: usize, voting_power_per_voter: u64, c: &mut Cri
...
+        let range_no = ranges_sampler.sample(rng);
+        let lower_bound = range_no
+            .checked_sub(1)
+            .map(|i| self.ranges_bounds[i])
+            .unwrap_or(self.threshold);
+        let current_range_sampler = self
+            .ranges_bounds
+            .get(range_no)
+            .copied()
+            .map(|upper_bound| Uniform::from(lower_bound..upper_bound))
```
eugene-babichenko, line 307: Need a discussion on this. Ideally, we do not need an upper bound,but we can easily go out ~to space~ of the `u64` range.

chain-impl-mockchain/benches/tally.rs
```diff
@@ -18,101 +18,136 @@ use chain_impl_mockchain::{
...
+
+    let should_vote_distribution =
+        WeightedIndex::new(&[1.0 - proposals_per_voter_ratio, proposals_per_voter_ratio]).unwrap();
+
+    let mut total_votes_per_proposal = vec![0; n_proposals];
+
+    for (i, proposal) in vote_plan.proposals().iter().enumerate() {
+        for (alias, voting_power) in voters_aliases.iter().zip(voting_powers.iter()) {
+            let should_vote = should_vote_distribution.sample(&mut rng);
+            if should_vote == 0 {
```
Zeegomo, line 129: I think this is simpler https://docs.rs/rand/0.8.3/rand/trait.Rng.html#method.gen_bool

chain-impl-mockchain/benches/tally.rs
```diff
@@ -124,12 +125,13 @@ fn tally_benchmark(
             }
 
             let mut private_voter = controller.wallet(&alias).unwrap();
+            let choice = Choice::new(votes_distribution.sample(&mut rng) as u8);
```
Zeegomo, line 133: same here.
We also previously used choices 1/2, while I think this results in choices 0/1 being used. Don't really now why we used 1/2 though, 0/1 seems more sensible to me.

chain-impl-mockchain/benches/tally.rs
```diff
@@ -283,11 +285,103 @@ fn tally_benchmark_1000_voters_1000_ada(c: &mut Criterion) {
...
+struct FundDistribution<'a, 'b> {
+    threshold: u64,
+    ranges_bounds: &'a [u64],
+    ranges_weights: &'b [f64],
+}
+
+impl<'a, 'b> Distribution<u64> for FundDistribution<'a, 'b> {
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> u64 {
+        assert_eq!(self.ranges_weights.len(), self.ranges_bounds.len());
+        let ranges_sampler = WeightedIndex::new(self.ranges_weights).unwrap();
```
Zeegomo, line 313: this could be moved to `FundDistribution` initialization

chain-impl-mockchain/benches/tally.rs
Zeegomo: With the current ADA -> voting power conversion, I think we can actually fit around 6e8 voters each with all ADA available (~3e10) in a u64 (~1.8e19)

Zeegomo (changes requested):
Looks good! Left a few notes

Zeegomo:
I think we should also consider benchmarking vote casting, last time I checked this was by far the heaviest operation in this bench

chain-impl-mockchain/benches/tally.rs
eugene-babichenko: What is more concerning is the actual maximum stake in the dataset (we are making an almost realistic benchmark, aren't we ;) ).

eugene-babichenko:
It definitely is, but with the rate of 5 votes per second or something like that this looks acceptable to me. Also I think it would be better to separate votes to a separate benchmark. They don't need all this setup and we can test them in isolation.

chain-impl-mockchain/benches/tally.rs
eugene-babichenko: `WeightedIndex` doesn't implement `Rng` and we need a particular distribution here. So we cannot do that unfortunately.

chain-impl-mockchain/benches/tally.rs
eugene-babichenko: > We also previously used choices 1/2, while I think this results in choices 0/1 being used

This doesn't really matter for a benchmark.

chain-impl-mockchain/benches/tally.rs
Zeegomo: Yes, I meant that we could allow the upper bound to be the maximum ADA available in circulation an still fit in a u64. This is still realistic in the sense that it would be impossible to have more than that, but is obviously quite strange to have that much concentrated in a single voter

chain-impl-mockchain/benches/tally.rs
Zeegomo: But we already have a `Rng`  https://github.com/input-output-hk/chain-libs/blob/f5791cd3597709fe981851e7b22ca00ec9168cf3/chain-impl-mockchain/benches/tally.rs#L41. As far as I understand weighted index is only used to get a boolean that is true with certain probability, that is exactly what `rng::gen_bool` does

eugene-babichenko: Ouch, I missed the part about the probability parameter in `gen_bool`. You're right!

eugene-babichenko:
Another observation: it uses only one CPU core to generate the workload. This is kinda painful when we start working with huge workloads. Will attempt to parallelize certain parts.

Possible options are:

- Key generation
- Transaction generation (not application!)

chain-impl-mockchain/benches/tally.rs
```diff
@@ -143,53 +182,74 @@ fn tally_benchmark(voters_count: usize, voting_power_per_voter: u64, c: &mut Cri
...
-            .unwrap()
-            .private_encrypted()
-            .unwrap()
-            .0
-            .state();
         let table = chain_vote::TallyOptimizationTable::generate_with_balance(total_votes, 1);
-        chain_vote::tally(total_votes, &tally_state, &decrypt_shares, &table).unwrap()
+        vote_plan_status
+            .proposals
+            .iter()
```
eugene-babichenko, line 211: errr, sequential tally decryption

chain-impl-mockchain/benches/tally.rs
```diff
@@ -125,12 +120,13 @@ fn tally_benchmark(
 
     for (i, proposal) in vote_plan.proposals().iter().enumerate() {
         for (private_voter, voting_power) in voters_and_powers.iter_mut() {
-            let should_vote = should_vote_distribution.sample(&mut rng);
-            if should_vote == 0 {
+            let should_vote = rng.gen_bool(proposals_per_voter_ratio);
+            if should_vote {
```
Zeegomo, line 124: Isn't this swapped? Like it skips voting when `should_vote` is true?

Zeegomo (changes requested):

Zeegomo (approved):

eugene-babichenko:
The panic is fixed by the latest commit, so this is unblocked now.

mzabaluev (approved):

PR #534, opened 2021-02-26
chain-vote: fix GroupElement::zero deserialization

eugene-babichenko:


Zeegomo (approved):
Maybe it would be a good idea to use the constant also when serializing, something like 
```
fn to_bytes(&self) -> [u8] {
  match self.0.to_affine() {
    None => ZERO_CONSTANT,
    Some(_) => ...
  }
}
```

dkijania (approved):

dkijania:
Tested on test branch. Fixed the problem with empty proposal

PR #538, opened 2021-03-02
Use default balance

Zeegomo:
Using default balance will help us decrypt the tally faster


eugene-babichenko:
What is the change compared to how it was done before?

mzabaluev (approved):

Zeegomo:
> What is the change compared to how it was done before?

For fund3 scenario is around 50%

eugene-babichenko (approved):

PR #539, opened 2021-03-02
Add benchmark for vote casting

Zeegomo:
Registering vote casts is a very expensive operation, so this PR adds a benchmark for it.

Benches structure is also changed so that we can reuse the same configuration for different tests.

### DIFF
--- a/chain-impl-mockchain/benches/tally.rs
+++ b/chain-impl-mockchain/benches/tally.rs
@@ -18,7 +18,7 @@ use chain_impl_mockchain::{
 };
 use criterion::{criterion_group, criterion_main, Criterion};
 use rand::{
-    distributions::{Distribution, Uniform, WeightedIndex},
+    distributions::{Bernoulli, Distribution, Uniform, WeightedIndex},
     Rng, SeedableRng,
 };
 use rayon::prelude::*;
@@ -109,7 +109,7 @@ fn tally_benchmark(
 
     // cast votes
     let vote_plan_def = controller.vote_plan(VOTE_PLAN).unwrap();
-    let vote_plan: VotePlan = vote_plan_def.clone().into();
+    let vote_plan: VotePlan = vote_plan_def.into();
 
     let mut total_votes_per_proposal = vec![0; n_proposals];
     let mut voters_and_powers: Vec<_> = voters_aliases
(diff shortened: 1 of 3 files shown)


chain-impl-mockchain/benches/vote/vote_cast.rs
```diff
@@ -0,0 +1,105 @@
...
+                ));
+            }
+        }
+        res
+    };
+
+    c.bench_function(&format!("vote_cast_{}", benchmark_name), |b| {
+        b.iter(|| {
+            let mut ledger = ledger.clone();
+            for vote in &vote_fragments {
```
eugene-babichenko, line 54: It looks like I had a wrong impression of what you were going to measure. For this kind of benchmark it is fine to include it into the existing benchmark function. It can be filtered out if needed.

Zeegomo: What did you mean? (maybe it's worthwhile to add that as well :))

eugene-babichenko: I mean that you generate exactly the same data we use for the vote tally benchmark, so you can just insert that benchmark in the sequence of benchmarking the vote tally. If someone does not want to run this benchmark, and only benchmark the tally steps, they can always do this: `cargo bench --manifest-path chain-impl-mockchain/Cargo.toml --features with-bench -- tally`

Zeegomo: Yes, that makes sense, but you hinted that there was miscommunication about the subject of the benchmark (i.e. you thought to benchmark another part) and maybe we could add that as well

eugene-babichenko: I thought you were going to benchmark one individual vote, not the whole fund load.

Zeegomo: This can be discussed indeed. I though to benchmark the whole fund because the number of previous votes affect the size of the hamt used to store them for example, so I suspect that adding the 1e6th vote could be slower. I may as well look into that to see if we can avoid to benchmark the whole fund

eugene-babichenko: > I may as well look into that to see if we can avoid to benchmark the whole fund

Benchmarking the whole fund is perfectly fine, because this

> the number of previous votes affect the size of the hamt used to store them

is a perfectly valid concern.

I meant that I misunderstood what you wanted to do and gave you the wrong advice about structuring your benchmarks.

eugene-babichenko (approved):

PR #542, opened 2021-03-13
Add decryption share verification

vincenthz:
* introduce a decryption nizk
* introduce a new `verify_decrypt_share` to verify decrypt share using the decryption nizk
* change format of tallydecryptshare to contains the proof
* also finish is now randomized due to randomized nizk

### DIFF
--- a/chain-vote/src/decr_nizk.rs
+++ b/chain-vote/src/decr_nizk.rs
@@ -1,0 +1,124 @@
+use super::gang::{GroupElement, Scalar};
+use super::gargamel::Ciphertext;
+use cryptoxide::digest::Digest;
+use cryptoxide::sha2::Sha512;
+
+/// Proof of discrete logarithm equivalence
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Proof {
+    a1: GroupElement,
+    a2: GroupElement,
+    z: Scalar,
+}
+
+pub(crate) const PROOF_SIZE: usize = 162; // Scalar is 32 bytes
+
+impl Proof {
+    pub fn to_bytes(&self) -> [u8; PROOF_SIZE] {
+        let mut output = [0u8; PROOF_SIZE];
+        output[0..65].copy_from_slice(&self.a1.to_bytes());
+        output[65..130].copy_from_slice(&self.a2.to_bytes());
+        output[130..162].copy_from_slice(&self.z.to_bytes());
+        output
+    }
+
+    pub fn to_slice_mut(&self, output: &mut [u8]) {
+        assert_eq!(output.len(), PROOF_SIZE);
+        output[0..65].copy_from_slice(&self.a1.to_bytes());
+        output[65..130].copy_from_slice(&self.a2.to_bytes());
+        output[130..162].copy_from_slice(&self.z.to_bytes());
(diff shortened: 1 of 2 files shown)


iquerejeta (commented):
I would recommend making some changes in the way we define the ZKPs. One way I see it is by defining the low level ZKPs (such as discrete log) in `chain-crypto`, and the more specific (such as correct decryption proof) here. This would allow us to use DLEQ proof in other contexts.  

I feel it's confusing to be in the context of a discrete logarithm proof, and have variables called `share`, `decrypted_share` or `pk`. In my opinion it would be clearer to explicitly distinguish between the discrete log proof and the decryption proof, as the former might be used in a context different correct decryption

chain-vote/src/lib.rs
```diff
@@ -121,15 +128,25 @@ impl EncryptedTally {
...
+        &self,
+        rng: &mut R,
+        secret_key: &OpeningVoteKey,
+    ) -> (TallyState, TallyDecryptShare) {
         let mut dshares = Vec::with_capacity(self.r.len());
         let mut r2s = Vec::with_capacity(self.r.len());
         for r in &self.r {
             let (r1, r2) = r.elements();
-            dshares.push(r1 * &secret_key.0.sk);
+            let decrypted_share = r1 * &secret_key.0.sk;
```
iquerejeta, line 140: This is also computed within the `generate()` function. We could pass it as input to avoid computing it twice.

iquerejeta (changes requested):
Generator and public key should be given as input of the challenge computation

chain-vote/src/decr_nizk.rs
```diff
@@ -0,0 +1,124 @@
...
+    ctx.input(&c.to_bytes());
+    ctx.input(&d.to_bytes());
+    ctx.input(&a1.to_bytes());
+    ctx.input(&a2.to_bytes());
+    ctx.result(&mut out);
+    Challenge(Scalar::from_bytes(&out[0..32]).unwrap())
+}
+
+/// Generate a decryption zero knowledge proof
+pub fn generate(w: &Scalar, share: &GroupElement, sk: &Scalar) -> Proof {
```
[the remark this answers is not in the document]
mzabaluev: Will this make `rand` (or another RNG library) an explicit dependency? OTOH you probably need it anyway to come up with the random input.

My take is, if the proof is sensitive to the choice of a random value or the value is derived from other input data, make it internal to the API.

chain-vote/src/decr_nizk.rs
```diff
@@ -0,0 +1,124 @@
...
+    let mut ctx = Sha512::new();
+    ctx.input(&c.to_bytes());
+    ctx.input(&d.to_bytes());
+    ctx.input(&a1.to_bytes());
+    ctx.input(&a2.to_bytes());
+    ctx.result(&mut out);
+    Challenge(Scalar::from_bytes(&out[0..32]).unwrap())
+}
+
+/// Generate a decryption zero knowledge proof
```
iquerejeta, line 73: Can't we use the proof in `chain-crypto/src/algorithms/vrf/dleq.rs`?

mzabaluev:
Superseded by #574.

PR #543, opened 2021-03-15
ledger: allow to apply block in separate steps

eugene-babichenko:


chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -723,66 +723,53 @@ impl Ledger {
...
-        // Apply all the fragments
-        for content in contents.iter() {
-            new_ledger = new_ledger.apply_fragment(ledger_params, content, metadata.block_date)?;
-        }
+        Ok(new_ledger)
+    }
+
+    /// This function must be called after applying fragments to the ledger. Do not use it if you
+    /// use `apply_block`.
+    pub fn apply_block_step3(&self, metadata: &HeaderContentEvalContext) -> Self {
```
enzoc4, line 771: so... step2 would be apply_fragments? Maybe `apply_block_finish` is better? Although that would suggest `start` instead of `step1`. The lack of `step2` threw me off a little at first.

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -723,66 +723,53 @@ impl Ledger {
         Ok(())
     }
 
-    /// Try to apply messages to a State, and return the new State if successful
-    pub fn apply_block(
+    /// This function must be called before applying fragments to the ledger. Do not use it if you
+    /// use `apply_block`.
+    pub fn apply_block_step1(
```
mzabaluev, line 728: It looks like this exposes an interim `Ledger` state in public API.

I think there should be a more ergonomic API: this method, renamed to, e.g. `begin_block()`, would return a wrapper type, say `BlockApply<'a>`.
That type would keep track of the metadata in a `&'a HeaderContentEvalContext` member, and likewise for the ledger parameters. An `fn apply_fragment(&mut self)` member should be present on this type, and the final method call, say, `fn finalize(self) -> Ledger`, would return the finallized block ledger state.

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -723,66 +723,53 @@ impl Ledger {
         Ok(())
     }
 
-    /// Try to apply messages to a State, and return the new State if successful
-    pub fn apply_block(
+    /// This function must be called before applying fragments to the ledger. Do not use it if you
```
enzoc4, line 726: This seems like a good case for a wrapping type, like
```rust
pub struct LedgerApplyBlock(Ledger);

impl LedgerApplyBlock {
  pub fn apply_fragments(self, ...) -> LedgerApplyBlock { ... }
  pub fn finish(self, ...) -> Ledger { ... }
}
```
But considering that `apply_fragment` is already public and unrestricted, it doesn't add too much anyway... So I also wonder how useful is the comment here.

enzoc4: I didn't notice @mzabaluev commented basically the same when I was writing this, so just answer to him (?)

eugene-babichenko:
In practice this would also need a version `ApplyBlockLedger` that owns `LedgerParameters`.

chain-impl-mockchain/src/chaineval.rs
```diff
@@ -38,7 +38,7 @@ impl AsRef<[u8]> for PraosNonce {
 
 /// Genesis Praos related data extract from the header
 #[derive(Debug, Clone)]
-pub(crate) struct HeaderGPContentEvalContext {
+pub struct HeaderGPContentEvalContext {
```
mzabaluev, line 41: Bikeshed: since it's going to be a public type, maybe name it more descriptively,
like `HeaderPraosContentEvelContext?`

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -1610,6 +1622,39 @@ impl Ledger {
...
+        let ledger = self
+            .ledger
+            .apply_fragment(&self.ledger_params, fragment, self.block_date)?;
+        Ok(ApplyBlockLedger {
+            ledger,
+            ..self.clone()
+        })
+    }
+
+    pub fn finish(self, maybe_gp_content: Option<&HeaderGPContentEvalContext>) -> Ledger {
```
mzabaluev, line 1636: The maybe_gp_content parameter type could may be an enum like

```
enum ConsensusEvalContext {
    Bft,  // nothing to see here, move along
    Praos {
        // ...
    }
}
```

To make it less ad-hoc and provide a pattern for extension in the future.

chain-impl-mockchain/src/chaineval.rs
mzabaluev: See the idea below about an enum.

chain-impl-mockchain/src/ledger/ledger.rs
eugene-babichenko: done

chain-impl-mockchain/src/header/header.rs
```diff
@@ -237,26 +237,30 @@ impl Header {
         }
     }
 
-    pub fn to_content_eval_context(&self) -> HeaderContentEvalContext {
-        let gp_content = match self.block_version() {
+    pub fn to_consensus_eval_context(&self) -> ConsensusEvalContext {
```
mzabaluev, line 240: This `to_` naming bugs me a little: we don't convert the whole header to another thing, it's more like a `get_` method (or a simple getter, that would have no verb prefix per good Rust style, but this one may do some work to get the value).

mzabaluev (approved):

PR #548, opened 2021-03-31
Migrate to circle ci images

filip-dulic-bloxico:
All CircleCI tests are green.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   rust-stable:
     docker:
-      - image: circleci/rust:latest
+      - image: cimg/rust:1.51.0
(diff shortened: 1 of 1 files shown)


Zeegomo (approved):
LGTM, just a nit about the version used

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   rust-stable:
     docker:
-      - image: circleci/rust:latest
+      - image: cimg/rust:1.51.0
```
Zeegomo, line 6: Nit: this was a mistake in the other pr, I think using `1.51` is better so that we get fixes without having to change the version

eugene-babichenko (approved):

PR #551, opened 2021-04-13
Hash to group

iquerejeta:
Implement hash to group for sec2 curve, and adaptation of the rest of the code-base to new function. Changes: 
* hash to group by hashing a given slice, and generating a point out of it
* removed random generations of commitment key to the deterministic `from_hash()`
* generate and verify vote take as input the CRS
* adapt rest of the code base to the changes above

### DIFF
--- a/chain-impl-mockchain/Cargo.toml
+++ b/chain-impl-mockchain/Cargo.toml
@@ -57,6 +57,7 @@ rand_chacha = "0.3"
 lazy_static = "1.3.0"
 tempfile = "3.1.0"
 rand = "0.8"
+cryptoxide = "0.3"
 
 
 [[bench]]
--- a/chain-impl-mockchain/Cargo.toml
+++ b/chain-impl-mockchain/Cargo.toml
@@ -57,8 +57,6 @@ rand_chacha = "0.3"
 lazy_static = "1.3.0"
 tempfile = "3.1.0"
 rand = "0.8"
-cryptoxide = "0.3"
-
 
 [[bench]]
 harness = false
--- a/chain-vote/src/commitment.rs
+++ b/chain-vote/src/commitment.rs
@@ -1,5 +1,4 @@
 use crate::gang::{GroupElement, Scalar};
-use rand_core::{CryptoRng, RngCore};
 use std::ops::{Add, Mul};
 
 /// Pedersen commitment
@@ -17,17 +16,6 @@ impl CommitmentKey {
     pub fn generate_from_seed(buffer: &mut [u8]) -> Self {
         CommitmentKey {
             h: GroupElement::from_hash(buffer),
-        }
-    }
-
-    pub fn commit<R: RngCore + CryptoRng>(&self, rng: &mut R, message: &Scalar) -> Commitment {
-        let randomness = Scalar::random(rng);
-        self.commit_with_randomness(message, &randomness)
-    }
-
-    pub fn commit_with_randomness(&self, message: &Scalar, randomness: &Scalar) -> Commitment {
-        Commitment {
-            c: GroupElement::generator() * message + &self.h * randomness,
         }
     }
 }
--- a/chain-vote/src/commitment.rs
+++ b/chain-vote/src/commitment.rs
@@ -1,6 +1,7 @@
-use crate::gang::{GroupElement, Scalar};
+use crate::gang::{GroupElement, Scalar, IncorrectHashLengthError};
 use rand_core::{CryptoRng, RngCore};
 use std::ops::{Add, Mul};
+use cryptoxide::digest::Digest;
 
 /// Pedersen commitment
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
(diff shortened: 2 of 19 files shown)


chain-vote/src/gang/p256k1.rs
```diff
@@ -10,6 +11,10 @@ pub struct Scalar(IScalar);
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct GroupElement(Point);
 
+#[derive(Debug, thiserror::Error)]
+#[error("The length of the hash is incorrect. It should be at least BYTES_LEN of Coordinate.")]
```
mzabaluev, line 15: I think you can use BYTES_LEN in a format parameter expression here:
```suggestion
#[error("The length of the hash is incorrect. It should be at least {} bytes", Coordinate::BYTES_LEN)]
```

mzabaluev (commented):
Looks good in general. We should consider changes in the workflow and possibly in the chain as the CRS needs to be present on encryption and verification as well.

chain-vote/src/commitment.rs
```diff
@@ -14,9 +15,25 @@ pub struct CommitmentKey {
...
+        match GroupElement::from_hash(hash){
+            Ok(h) => Ok(CommitmentKey { h }),
+            Err(e) => Err(e)
+        }
+
+    }
+
+    pub fn commit<R: RngCore + CryptoRng>(&self, rng: &mut R, message: &Scalar) -> Commitment {
+        let randomness = Scalar::random(rng);
+        Commitment {
```
Zeegomo, line 28: Nit: could this call `commit_with_randomness` with the calculated randomness?

Zeegomo (commented):
This looks good.
I wonder if it is better to put the `CRS` seed in the `VotePlan` creation transaction instead of using `VotePlan.to_id()`, so that it is more flexible if we want to change that in the future. This comes at the cost of a greater size of the transaction, especially if the seed is long. @mzabaluev

chain-impl-mockchain/Cargo.toml
```diff
@@ -57,6 +57,7 @@ rand_chacha = "0.3"
 lazy_static = "1.3.0"
 tempfile = "3.1.0"
 rand = "0.8"
+cryptoxide = "0.3"
```
vincenthz, line 60: I failed to find relevant diff that requires it, but should it not be abstracted anyway in chain-vote / chain-crypto ?

chain-vote/src/gang/p256k1.rs
```diff
@@ -45,23 +57,46 @@ impl GroupElement {
...
+            hash.input(buffer);
+            hash.input(&i.to_be_bytes());
+            hash.result(&mut result);
+            hash.reset();
+            // arbitrary encoding of sign
+            let sign = if result[32] & 1 == 0 {
+                Sign(Positive)
+            } else {
+                Sign(Negative)
+            };
```
vincenthz, line 75: doesn't seem particularly useful to get the sign from the binary data and at the same time pass the binary data unmodified; why not just assume positive / negative value ?

vincenthz (approved):
looking fine, a bit hard to review as a diff of everything due to renaming and reformating tagging along the useful diff

chain-vote/src/gang/p256k1.rs
Zeegomo: Wouldn't that make the output not uniform?

chain-impl-mockchain/Cargo.toml
iquerejeta: Good catch. Removed

chain-vote/src/gang/p256k1.rs
iquerejeta: Agree with @Zeegomo , assuming positive or negative would reduce the number of points we can randomly choose from. The sign should also be randomly selected. 

And I am not sure I understand the first part of your comment. We get the sign from the last byte, and then we pass all bytes but the last (unmodified) to compute the point. What is not useful about this?

vincenthz: ok one thing I missed is that it's using blake2b264, not blake2b256. which make sense to get the bit from that extra byte. however, whilst it doesn't have the re-use bit that I originally thought about, it still have no practical use. losing one bit of randomness on E(Fp) make no difference in the grand scheme, and it doesn't make a difference on uniformity as you don't introduce a bias (all points / {0} have a negative, you just take away uniformly the negative).

iquerejeta: We made that choice to go according to the algorithm presented in [Short signatures from the Weil pairing](https://www.iacr.org/archive/asiacrypt2001/22480516.pdf) (Algorithm MapToGroup). And indeed, by preselecting the sign you choose uniformly from the negative, but we designed the algorithm so that it chooses uniformly from the prime order group. I agree that losing one bit of randomness makes not practical difference, but I would keep it like that unless you see an important drawback on this scheme.

PR #545, opened 2021-03-30
run cargo fmt &clippy 1.51

eugene-babichenko:

PR #552, opened 2021-04-16
Peers update

Zeegomo:
Make `GossipService::peers` return a valid gossip instead of just the address of a peer. A gossip includes also the node public key and is signed with that key, so that it cannot be tampered.

### DIFF
--- a/chain-network/proto/node.proto
+++ b/chain-network/proto/node.proto
@@ -61,31 +61,8 @@ message FragmentIds {
 // Request for peers
 message PeersRequest { uint32 limit = 1; }
 
-// Responses as a bunch of peers
-message PeersResponse { repeated Peer peers = 1; }
-
-// A Peer on the network
-message Peer {
-  // FIXME: make the port field common, indicate TCP protocol selection
-  // by the field number.
-  oneof peer {
-    PeerV4 v4 = 1;
-    PeerV6 v6 = 2;
-  }
-}
-
-// An ipv4 peer
-message PeerV4 {
-  fixed32 ip = 1;
-  fixed32 port = 2; // FIXME: uint32 is more efficient for port numbers
-}
-
-// An ipv6 peer
-message PeerV6 {
-  fixed64 ip_high = 1;
-  fixed64 ip_low = 2;
-  fixed32 port = 3; // FIXME: uint32 is more efficient for port numbers
-}
+// Responses as a bunch of peers, similar to Gossip
+message PeersResponse { repeated bytes peers = 2; }
 
 // Request message for method PullHeaders.
 // This message can also be send by the service as a BlockEvent variant.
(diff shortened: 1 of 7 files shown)


chain-network/proto/node.proto
```diff
@@ -61,31 +61,8 @@ message FragmentIds {
...
-}
-
-// An ipv6 peer
-message PeerV6 {
-  fixed64 ip_high = 1;
-  fixed64 ip_low = 2;
-  fixed32 port = 3; // FIXME: uint32 is more efficient for port numbers
-}
+// Responses as a bunch of peers, similar to Gossip
+message PeersResponse { repeated bytes peers = 2; }
```
eugene-babichenko, line 65: Is that to keep the backwards compatibility? Not sure if we care about it much right now.

chain-network/src/grpc/convert.rs
```diff
@@ -173,97 +156,18 @@ impl IntoProtobuf for Gossip {
...
-                    ip_low,
-                })
-            }
+impl FromProtobuf<proto::PeersResponse> for Gossip {
+    fn from_message(message: proto::PeersResponse) -> Result<Self, Error> {
+        let gossip = Gossip {
+            nodes: message
+                .peers
+                .into_iter()
+                .map(Node::from_bytes)
```
eugene-babichenko, line 165: Is there `try_from_bytes`? It would make much more sense since this is the input from the outside world (we are not sure it valid) and the calling function is already fallible.

chain-network/proto/node.proto
Zeegomo: if you are talking about the field number yes, I used 2 so that it is (kind of) backward compatible.
 With regards to using a custom response message instead of reusing `Gossip`, I think it may be cleaner, and it is also what `TipResponse` does.

chain-network/src/grpc/convert.rs
Zeegomo: well, `Node` is just a `Box<[u8]>` in this context, I guess we can't really enforce properties at this level

eugene-babichenko: Ah, well, I did not look at the type definition.

eugene-babichenko (approved):

eugene-babichenko:
Better check it with @mzabaluev as well.

mzabaluev (approved):
Looks good to me. Could peer entries be different from authenticated gossip, though, e.g. lacking signatures?

Zeegomo:
> Could peer entries be different from authenticated gossip, though, e.g. lacking signatures?

As far as `chain-network` is concerned yes, but I think we would want that in Jormungandr to prevent nodes from sending malicious information.

mzabaluev:
Fix lints and this is good to go.

Zeegomo:
The lints are fixed by https://github.com/input-output-hk/chain-libs/pull/545 and will require extensive update in jormungandr before , I would rather fix them after merging this

eugene-babichenko:
Agree. I will work on updating Jormungandr after we are done with this
merge.

PR #555, opened 2021-04-26
Ristretto flag

iquerejeta:
Remove the existing code of `zerocaf` which did not compile. Moreover, there exists better implementations (more complete and maintained) of the ristretto group. 

I've made the changes necessary so that the code can run over Curve25519 using ristretto255 prime order group (using the z
`curve25519-dalek-ng` crate). The benches show that elliptic curve operations (point-scalar multiplication and point addition) are better using ristretto255. This results in a ~x2 improvement in the proof generation and verification. 

However, scalar-scalar multiplication works much better using eccoxide, meaning that when the number of scalar exponentiations becomes big (when the number of candidates increases), using ristretto255 performs worse than the default (at around 512 candidate, default performs better).


Zeegomo:
If I understand correctly and 'candidates' are the number of vote options for a proposal, then it seems like 512 would be quite uncommon.
There are more voting benchmarks in `chain-impl-mockchain`, would be interesting to see effects there as well.

Zeegomo (changes requested):

chain-vote/src/gang/p256k1.rs
```diff
@@ -110,6 +118,11 @@ impl GroupElement {
         })
     }
 
+    pub(super) fn encode_hash_map(&self) -> Option<[u8; FieldElement::SIZE_BYTES]> {
```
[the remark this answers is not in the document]
Zeegomo: I'm still a bit skeptical about this. The name of this method suggest to me a behavior that should be implemented already by the `Hash` trait. 
The reason we were not using the `Hash` trait directly in bsgs is to do some optimizations, but those optimizations are only valid in that algorithm context and should be explicit (i.e. it's not something that the `GroupElement` can hide from bsgs, otherwise we won't be able to recover the point later).
I would rather keep the old approach of explicitly using only the x coordinate for sec2 curves

Zeegomo (changes requested):

chain-vote/src/gang/p256k1.rs
```diff
@@ -32,6 +32,13 @@ impl Hash for Scalar {
     }
 }
 
+#[allow(clippy::derive_hash_xor_eq)]
+impl Hash for Coordinate {
```
Zeegomo, line 36: Is this used?

chain-vote/src/gang/p256k1.rs
```diff
@@ -53,6 +60,7 @@ impl Coordinate {
 impl GroupElement {
     /// Size of the byte representation of `GroupElement`.
     pub const BYTES_LEN: usize = 65;
+    pub const HASH_MAP_LEN: usize = FieldElement::SIZE_BYTES;
```
Zeegomo, line 63: I think this should be removed as well, we should not require a struct to commit to a specific representation in a hashmap, also because the current implementation seems tied to a specific optimization

chain-vote/src/gang/ristretto255.rs
```diff
@@ -0,0 +1,366 @@
...
+impl Hash for Scalar {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        state.write(&self.to_bytes())
+    }
+}
+
+impl GroupElement {
+    /// Size of the byte representation of `GroupElement`. We always encode the compressed value
+    pub const BYTES_LEN: usize = 32;
+    pub const HASH_MAP_LEN: usize = Self::BYTES_LEN;
```
Zeegomo, line 41: same as above

Zeegomo (changes requested):
We can remove the optimization for sec2 if that is too much of a burden and we plan to switch to ristretto anyway

Zeegomo (approved):

chain-vote/Cargo.toml
```diff
@@ -11,25 +11,22 @@ thiserror = "1.0"
...
 default = ["p256k1"]
 p256k1 = ["eccoxide"]
-ed = ["zerocaf"]
-with-bench = ["criterion"]
+ristretto255 = ["curve25519-dalek-ng"]
 
 [[bench]]
 harness = false
 name = "shvzk"
-required-features = ["with-bench"]
```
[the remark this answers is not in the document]
eugene-babichenko: No practical need for that if we do not have benchmarks-only interfaces or dependencies.

PR #556, opened 2021-04-28
Fiat shamir

### LINKED ISSUE #553: Use Merlin to automate Fiat-Shamir transform
To avoid bugs such as the one existing in #542, we could use [Merlin](https://github.com/zkcrypto/merlin). This will make the ZKP implementations less error prone.

iquerejeta:
Closes #553 . Use merlin for the fiat-shamir transcripts.

### DIFF
--- a/chain-impl-mockchain/Cargo.toml
+++ b/chain-impl-mockchain/Cargo.toml
@@ -29,6 +29,7 @@ cardano-legacy-address = { path= "../cardano-legacy-address" }
 rand_chacha = { version = "0.3", optional = true }
 chain-test-utils = { path = "../chain-test-utils", optional = true }
 rayon = "1.5.0"
+merlin = "3"
 criterion = { version = "0.3.0", optional = true }
 
 [features]
--- a/chain-impl-mockchain/src/certificate/test.rs
+++ b/chain-impl-mockchain/src/certificate/test.rs
@@ -8,6 +8,7 @@ use crate::vote;
 use chain_core::mempack::{ReadBuf, Readable};
 use chain_crypto::{testing, Ed25519};
 use chain_time::DurationSeconds;
+use merlin::Transcript;
 #[cfg(test)]
 use quickcheck::TestResult;
 use quickcheck::{Arbitrary, Gen};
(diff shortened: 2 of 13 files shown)


Zeegomo:
I have a bunch of questions:
* How does the use of a `Transcript` prevent the implementor from forgetting to add some data to the proof context like it could happen with standard methods?
* With this flexible, builder like approach, I think this is prone to errors in each of the points in the code where you actually have to construct challenges, whereas the old `ChallengeContext` could only be constructed if all the parameters were supplied, so you just needed to check that it is correct in one place.

iquerejeta:
I agree that it is not an absolute winner compared to the old approach, and I believe it falls to personal preference. But the reasons I prefer using the `transcript` are: 

* The way the `transcript` is defined and used, allows for the proof developer to include each element independently in the transcript, i.e.: we use a public key? then we add it; we generate a commitment of the first round of the proof? then we add it; we define a commitment key? then we add it, etc. Of course, it can still happen that the developer forgets to add the commitment key. This does _not_ enforce correct usage. Nonetheless, IMO it is less error prone because of that; the proof starts? then we include the whole statement in the transcript, we "send" a value to the verifier? then we include it in the transcript.
* The current approach requires the proof developer to define independent functions for each challenge computation. I find it cleaner to have everything centralised in `transcript.rs`, and simply define functions for appending the different types to the transcript. Then depending on the statement/proof, we do not need to generate a new function, but simply append the different values to the transcript. 

I see your point of building a different structure for each challenge, so that it enforces it's correct usage, and I understand the reasoning that it is less error prone. Currently we have a single proof in the stack, but there is already a PR for another proof, and yet another one will come when we fully implement the DKG. This would require three different challenge contexts in the library. 

As I said, this is not a required change, but rather a design proposal. If the usage of a ChallengeContext for each proof is preferred, I am happy to keep it that way.

Zeegomo:
Another things that bugs me a little is that I guess the order in which you add things to the transcript is important, even for things that do not have a predefined order (e.g. multi part statements).
While this will be likely caught early with really simple tests,  you end up with some "shadowy" (at least in my eyes) dependency between the prover and verifier, that I do not see at first glance and that is not enforced by the transcript itself.

Since the flow for the transcript is the same, could those part be refactored in a shared function, that branches only for either computing or verifying the challenges?

Or if we go for a different struct for each challenge, could we internally reuse merlin in some way that speeds up / simplify the development?

iquerejeta:
I think that if we decide to go for specific functions for each challenge computation, including the merlin dependency is not required. The main motivation to use Merlin was to avoid having to define single function for each challenge generation. If we decide that that is the best way to go (with respect to design and reducing possible errors), we should close this PR and simply resolve the current instantiation. Maybe we could reorganise the ChallengeContext so that we have all challenge contexts in one place.

vincenthz:
I think Merlin is great for doing the transcript, but the problem is that it's a non-standardised component that is not, rightfully, expected to offer stability or compatibility (forward or backward), which is fundamentally incompatible with the blockchain data.

Effectively using Merlin, the blockchain is at the mercy of upstream maintenance in term of having your previous data still replayable, and/or stuck with a very specific version; I don't think it's indicated in this settings to do this, unless you're planning to vendor it.

iquerejeta:
Makes sense 👍  I'll close this PR then, and resolve the problem using the current solution of defining a challenge context for each proof.

PR #560, opened 2021-05-03
non-node clients grpc api proposal

enzoc4:
# Context 

It seems there is a need for some API's to support heterogeneous clients (not nodes, basically), besides the one's that we have in rest.

See https://github.com/input-output-hk/jormungandr/issues/3227 and maybe https://github.com/input-output-hk/jormungandr/issues/2354 for some context.

The kinds of clients we could support are:

- explorers
- wallets
- a mempool watching service (which could be considered a subsystem of a wallet...)

## Disclaimers

Just a PoC to get a conversation started. I think this belongs in `chain-libs` and I wanted to put it in `chain-network`, but I think it doesn't fit as it is, and I don't want to spend time reorganizing things without having some discussion about it first.

Also, the name `chain-watch` is only half-serious... I just couldn't think of something better yet.

## Extra comments

The `SyncMultiverse` procedure is somewhat a placeholder to bring out discussion about it... I think it would be a valid way to bootstrap something like the explorer, but there are more than one way we could go about that.
For example, we could re-expose some of the procedures in the `node` service. (None does exactly the same, but there are some similar things). Or have a third common service for those things... Or just let clients use the `node` service.

Also, I think this may be also missing a way to get the current fragments in the mempool (which we already have in rest, so that brings out another question).

Another thing I think could be nice for these API's, is that we could try to send

### DIFF
--- a/chain-watch/Cargo.toml
+++ b/chain-watch/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "chain-watch"
+version = "0.1.0"
+authors = ["Enzo Cioppettini <ecioppettini@atixlabs.com>"]
+edition = "2018"
+
+[dependencies]
+prost = "0.7"
+
+[dependencies.tonic]
+version = "0.4"
+default-features = false
+features = ["codegen", "prost"]
+
+[build-dependencies.tonic-build]
+version = "0.4"
+default-features = false
+features = ["prost"]
+
+[features]
+default = ["tonic/transport", "tonic-build/transport"]
--- a/chain-watch/proto/watch.proto
+++ b/chain-watch/proto/watch.proto
@@ -1,0 +1,60 @@
+syntax = "proto3";
+
+package iohk.chain.watch;
+
+
+message Block {
+  bytes content = 1;
+}
+
+message BlockSubscriptionRequest {}
+
+message TipSubscriptionRequest {}
+
+message MempoolSubscriptionRequest {}
+
+message SyncMultiverseRequest {
+  // send blocks with this chain_length or greater
+  uint32 from = 1;
+}
+
+message BlockId {
+  bytes content = 1;
+}
+
+message MempoolEvent {
+    bytes fragment_id = 1;
+    oneof event {
+        MempoolFragmentInserted inserted = 2;
+        MempoolFragmentRejected rejected = 3;
+        MempoolFragmentInABlock in_a_block = 4;
+    };
+}
+
+
+message MempoolFragmentInserted {}
+
+message MempoolFragmentRejected {
+    string reason = 1;
+}
+
+message MempoolFragmentInABlock {
+    BlockId block = 1;
+}
+
+
+service SubscriptionService {
+  // get a stream of blocks succesfully processed by the node, this means they
+  // are already validated.
+  // the parent of a block will always be streamed before the block itself.
+  rpc BlockSubscription(BlockSubscriptionRequest) returns (stream Block);
+
+  // get tip updates
+  rpc TipSubscription(TipSubscriptionRequest) returns (stream BlockId);
+
+  rpc MempoolSubscription(MempoolSubscriptionRequest) returns (stream MempoolEvent);
+
+  // fetch all blocks from the given initial chainlength to the tip, from all
+  // possible branches and in increasing order 
+  rpc SyncMultiverse(SyncMultiverseRequest) returns (stream Block);
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ members = [
     "cardano-legacy-address",
     "sparse-array",
     "typed-bytes",
+    "chain-watch",
 ]
 
 [profile.bench]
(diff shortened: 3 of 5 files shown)


chain-watch/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+default-features = false
+features = ["codegen", "prost"]
+
+[build-dependencies.tonic-build]
+version = "0.4"
+default-features = false
+features = ["prost"]
+
+[features]
+default = ["tonic/transport", "tonic-build/transport"]
```
eugene-babichenko, line 21: A newline please :)

chain-watch/proto/watch.proto
```diff
@@ -0,0 +1,60 @@
...
+
+  // get tip updates
+  rpc TipSubscription(TipSubscriptionRequest) returns (stream BlockId);
+
+  rpc MempoolSubscription(MempoolSubscriptionRequest) returns (stream MempoolEvent);
+
+  // fetch all blocks from the given initial chainlength to the tip, from all
+  // possible branches and in increasing order 
+  rpc SyncMultiverse(SyncMultiverseRequest) returns (stream Block);
+}
```
eugene-babichenko, line 60: same here

chain-watch/proto/watch.proto
```diff
@@ -0,0 +1,60 @@
...
+  rpc BlockSubscription(BlockSubscriptionRequest) returns (stream Block);
+
+  // get tip updates
+  rpc TipSubscription(TipSubscriptionRequest) returns (stream BlockId);
+
+  rpc MempoolSubscription(MempoolSubscriptionRequest) returns (stream MempoolEvent);
+
+  // fetch all blocks from the given initial chainlength to the tip, from all
+  // possible branches and in increasing order 
+  rpc SyncMultiverse(SyncMultiverseRequest) returns (stream Block);
```
eugene-babichenko, line 59: There might be consistency issues, e.g. branches starting from different roots. But I guess we just let the client to deal with that.

chain-watch/proto/watch.proto
```diff
@@ -0,0 +1,60 @@
...
+service SubscriptionService {
+  // get a stream of blocks succesfully processed by the node, this means they
+  // are already validated.
+  // the parent of a block will always be streamed before the block itself.
+  rpc BlockSubscription(BlockSubscriptionRequest) returns (stream Block);
+
+  // get tip updates
+  rpc TipSubscription(TipSubscriptionRequest) returns (stream BlockId);
+
+  rpc MempoolSubscription(MempoolSubscriptionRequest) returns (stream MempoolEvent);
```
eugene-babichenko, line 55: And now I am not sure that this should belong to chain-libs. We usually did not introduce the concept of mempool at the chain-libs level.

eugene-babichenko (commented):
The plan is basically to reuse some already existing implementations, right? At least, I can imagine that with `BlockSubscription` and other stuff must be trivial.

chain-watch/proto/watch.proto
enzoc4: I'm curious about when that can happen, you mean branches with different genesis?

chain-watch/proto/watch.proto
enzoc4: Well, current plan is to move it to chain-network, actually. If we put this in jormungandr... And we want to re-use from a different repository, then we would have to include jormungandr there somehow?

Although, @mzabaluev motivation to put it in `chain-network` was the dependencies/build process, I think.

chain-watch/proto/watch.proto
eugene-babichenko: No, I meant something like that:

```
o - o - o - o - o - o - o
         \ - o - o - o
                 ^
                 |
                 We make request from this height and
                 get two incomplete branches instead
                 of a "proper" tree
```

chain-watch/proto/watch.proto
eugene-babichenko: > motivation to put it in `chain-network` was the dependencies/build process, I think

Yeah, mostly that I think.

> then we would have to include jormungandr there somehow

Not a huge problem, we are doing that for `jormungandr-lib` already.

chain-watch/proto/watch.proto
enzoc4: Until the explorer has a database I guess you need to start from 0. But after that, I was thinking on starting from `min(stability_depth, last_block_I_have)`. It may lead to sending a lot of redundant blocks, though... And you would need a way of knowing the current length (or the current stable block)

Also, I was thinking on sending all blocks of a given height before the next one, separate branches can be applied potentially in parallel, so I think it's better than sending one branch at a time.

I did write this because it was the simplest thing I could think of, though... Originally I thought about the explorer sending the list of tips it has, but that's not that simple I think.

enzoc4: I've been also been wondering if this endpoint shouldn't be merged with the *live* block subscription, because otherwise there is a chance of missing a block in the middle if you switch from one stream to the other.

In that case the block subscription could have an optional parameter to sync before, and then start sending new blocks.

But clients can work around that by starting both I guess, I'm not really sure what's idiomatic here.

PR #564, opened 2021-05-07
Reestructure and review of the unit vector zero knowledge proof

iquerejeta:
Reestructure the code of the unit vector zero knowledge proof, including docs and testing. Resolution of bugs of ZKP.


chain-impl-mockchain/src/vote/payload.rs
```diff
@@ -131,36 +130,7 @@ impl ProofOfCorrectVote {
...
-            zwvs.push(zwv);
-        }
-        let r_buf = buf.get_slice(Scalar::BYTES_LEN)?;
-        let r = Scalar::from_bytes(r_buf).ok_or_else(|| {
-            ReadError::StructureInvalid("Invalid Proof encoded R scalar".to_string())
-        })?;
-        Ok(Self(chain_vote::ProofOfCorrectVote::from_parts(
-            ibas, bs, zwvs, r,
-        )))
+        chain_vote::ProofOfCorrectVote::from_buffer(buf).map(|proof| Self(proof))
```
danielSanchezQ, line 133: Probably just `map(Self)` or `map(ProofOfCorrectVote)` would work.

eugene-babichenko:
Please rebase it onto the latest master. We did several naming changes to conform the Rust coding style, you should check with `cargo clippy`.

chain-vote/src/commitment.rs
```diff
@@ -1,85 +1,61 @@
...
+    /// Return a commitment with the given opening, `o`
+    pub fn commit_with_open(&self, o: &Open) -> GroupElement {
+        self.commit(&o.m, &o.r)
     }
 
-    pub fn new(ck: &CommitmentKey, m: &Scalar, r: &Scalar) -> Self {
-        let c = GroupElement::generator() * m + &ck.h * r;
-        Commitment { c }
+    /// Return a commitment with the given message, `m`,  and opening key, `r`
+    pub fn commit(&self, m: &Scalar, r: &Scalar) -> GroupElement {
```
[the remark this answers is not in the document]
iquerejeta: I thought about this [rng as input]. The problem of a commitment (contrarily to encryption) is that the randomness needs to be stored to memory, as it is needed to prove knowledge of the committed value. So we would need to return the group element and the opening (and still keep this function for verification procedure).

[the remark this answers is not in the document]
iquerejeta: And for constructing `Announcement` and `BlindingRandomness` together, it might make sense. I'll give it a try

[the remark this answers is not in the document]
Zeegomo: Yep, I was thinking of something like `encrypt_return_r` we use in encryption. Then the `commit` function that accepts the raw randomness instead of a PRNG could be made private and only used internally (for both commit and verificatiol)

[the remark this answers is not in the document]
iquerejeta: ah, that is also a good point 👍  I'll do that. Check out the code with the `Announcement` and `BlindingRandomness` as discussed above. I think it is better as you said 👌  (the last commit)

chain-vote/src/private_voting/unit_vector_zkp.rs
```diff
@@ -0,0 +1,459 @@
...
+        ciphertexts: &Ptp<Ciphertext>,
+        challenge_x: &Scalar,
+        challenge_y: &Scalar,
+    ) -> bool {
+
+        // check commitments are 0 / 1
+        for (iba, zwv) in self.ibas.iter().zip(self.zwvs.iter()) {
+            let com1 = commitment_key.commit_with_random(&zwv.z, &zwv.w);
+            let lhs = &iba.i * challenge_x + &iba.b;
+            if lhs != com1 {
```
Zeegomo, line 233: Could this (and the following check) be rewritten as `commitment_key.verify(iba.i * challenge_x + &iba.b, Open {m: zwv.z, r: zwv.w })`?
If so, `commit_with_random` could be made even private

iquerejeta: Very good catch 👌

chain-vote/src/private_voting/unit_vector_zkp.rs
```diff
@@ -0,0 +1,475 @@
...
+                .chain(self.zwvs.iter().map(|zwv| zwv.w))
+                .chain(iter::repeat(challenge_x.negate()).take(bits))
+                .chain(iter::repeat(Scalar::one().negate()).take(bits))
+                .chain(self.zwvs.iter().map(|zwv| zwv.v))
+                .chain(self.zwvs.iter().map(|zwv| zwv.z - challenge_x))
+                .chain(iter::repeat(Scalar::one().negate()).take(bits))
+                .chain(powers_cy.take(length).map(|s| s * cx_pow))
+                .chain(powers_cy.take(length).map(|s| s * cx_pow))
+                .chain(powers_cy.take(length).map(|s| s))
+                .chain(powers_cx.take(bits).map(|s| s))
```
Zeegomo, line 203: What's up with `map(|s| s)`?

chain-vote/src/private_voting/unit_vector_zkp.rs
```diff
@@ -0,0 +1,475 @@
...
+        let mega_check = GroupElement::multiscalar_multiplication(
+            self.zwvs
+                .iter()
+                .map(|zwv| zwv.z)
+                .chain(self.zwvs.iter().map(|zwv| zwv.w))
+                .chain(iter::repeat(challenge_x.negate()).take(bits))
+                .chain(iter::repeat(Scalar::one().negate()).take(bits))
+                .chain(self.zwvs.iter().map(|zwv| zwv.v))
+                .chain(self.zwvs.iter().map(|zwv| zwv.z - challenge_x))
+                .chain(iter::repeat(Scalar::one().negate()).take(bits))
```
Zeegomo, line 199: Wondering if it is possible to provide a value that was not the committed one (i.e. first part of the multiscalar mul is not zero), while still getting `GroupElement::zero` in the end of the computation

Zeegomo (commented):
I haven't actually checked the specific of the proof

chain-vote/src/private_voting/unit_vector_zkp.rs
iquerejeta: woops, first implementation was over `RistrettoPoint`s, so it was `map(|s| s.0)`. Removed 👍

chain-vote/src/private_voting/unit_vector_zkp.rs
iquerejeta: yes, absolutely. Forgot an important detail.

iquerejeta: so what I wanted to do initially (that I forgot, and it is absolutely crucial that you figured it out, thanks 🙏 ) was to use a random scalar, computed by the verifier to be able to include all checks in a single multiscalar mult (see current code). However, with this new random number I thought the code did not read very well, so I thought of breaking it into `bits` multiscalar multiplications for the first part of the verification. But efficiency was slightly affected, and it did not read much better. So I reverted to the initial idea---everything in a single multiscalar mult. WDYT?

iquerejeta: Also, with this verification procedure, the verifier needs to compute a random number. Also had a self debate there. Should the verifier directly use `thread_rng`, or should we give as input a `CryptoRng + RngCore`? I chose the former, as otherwise we had to handle RNGs in the mockchain code. So basically, `fn verify_statements()` directly uses `thread_rng()`.

chain-vote/Cargo.toml
```diff
@@ -5,31 +5,31 @@ authors = ["Vincent Hanquez <vincent.hanquez@iohk.io>"]
...
 [[bench]]
 harness = false
 name = "shvzk"
-required-features = ["with-bench"]
 
-[[bench]]
-harness = false
-name = "curve"
-required-features = ["with-bench"]
+#[[bench]]
```
Zeegomo, line 33: Is this intentional?

chain-vote/src/commitment.rs
```diff
@@ -33,6 +33,20 @@ impl CommitmentKey {
         (self.commit_with_random(m, &r), r)
     }
 
+    /// Return a commitment of a boolean value, and the used randomness, `r`, where the latter is computed
+    /// from a `Rng + CryptoRng`
+    pub(crate) fn commit_bool<R>(&self, m: &bool, rng: &mut R) -> (GroupElement, Scalar)
```
Zeegomo, line 38: I don't think a immutable reference to a bool makes much sense, but I thought clippy would catch that

Zeegomo (commented):
There are also a few formatting issues reported by clippy

Zeegomo (approved):
Good work!

chain-vote/Cargo.toml
```diff
@@ -5,31 +5,31 @@ authors = ["Vincent Hanquez <vincent.hanquez@iohk.io>"]
...
 # TODO replace with the crates.io version once it has faster scalar multiplication
 eccoxide = { git = "https://github.com/eugene-babichenko/eccoxide.git", branch = "fast-u64-scalar-mul", features = ["fast-u64-scalar-mul"], optional = true }
 curve25519-dalek-ng = { version = "4.0.1", optional = true}
-criterion = { version = "0.3", optional = true }
 
 [dev-dependencies]
 rand_chacha = "0.3"
 smoke = "^0.2.1"
 
+criterion = { version = "0.3" }
```
eugene-babichenko, line 22:
```suggestion
criterion = "0.3"
```

chain-vote/src/private_voting/challenge_context.rs
```diff
@@ -0,0 +1,52 @@
+//! Challenge context for the Unit Vector Zero Knowledge Proof. The common reference string
+//! is a commitment key, and the statement consists of a public key, and the encryption of each
+//! entry of the vector.
```
eugene-babichenko, line 3: I would rather make this documentation for ChallengeContext, not for the entire module.

chain-vote/src/private_voting/mod.rs
```diff
@@ -0,0 +1,14 @@
+//! Implementation of the Unit Vector ZK argument presented by
```
eugene-babichenko, line 1: YES YES YES. Good to have these links here :)

chain-vote/src/private_voting/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+//! Zhang, Oliynykov and Balogum in
+//! ["A Treasury System for Cryptocurrencies: Enabling Better Collaborative Intelligence"](https://www.ndss-symposium.org/wp-content/uploads/2019/02/ndss2019_02A-2_Zhang_paper.pdf).
+//! We use the notation presented in the technical
+//! [spec](https://github.com/input-output-hk/treasury-crypto/blob/master/docs/voting_protocol_spec/Treasury_voting_protocol_spec.pdf),
```
eugene-babichenko, line 5: Unfortunately, this link gets me to 404.

eugene-babichenko: This may be due to limited permissions.

iquerejeta: ah, yes, I thought this was public. Do you think it makes sense to note that it needs permissions (giving that we don't expect this to be a public library)? Or rather modify the docs and notation to fit with the published paper?

chain-vote/src/lib.rs
```diff
@@ -17,8 +17,8 @@ pub mod debug {
     pub mod gang {
         pub use crate::gang::*;
     }
-    pub mod shvzk {
-        pub use crate::shvzk::*;
+    pub mod private_voting {
```
vincenthz, line 20: much prefer the old name personally ¯\_(ツ)_/¯

vincenthz (approved):
It's looking fine, apart from the creative disagreement into scoping and naming. should go ahead

chain-vote/src/private_voting/mod.rs
iquerejeta: I'll talk with the research team then 👍

chain-vote/src/lib.rs
iquerejeta: I don't have a strong feeling about `private_voting`, but special honest verifier zero knowledge (`shvzk`) was a bit generic. The reason I changed the name is because the proofs used in the DKG will also be special honest verifier zero knowledge, so wanted to have a clear distinction of it.

chain-vote/Cargo.toml
```diff
@@ -5,31 +5,31 @@ authors = ["Vincent Hanquez <vincent.hanquez@iohk.io>"]
 edition = "2018"
 
 [dependencies]
+chain-core = { path = "../chain-core" }
+rand = "0.8"
```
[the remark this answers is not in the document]
iquerejeta: ah! problem is that `ristretto255` flag does use it in `unit_vector_zkp.rs` to compute the `batch_challenge` scalar.

eugene-babichenko (approved):

PR #565, opened 2021-05-18
add explicit spending counter to account transaction inputs

eugene-babichenko:
This is required for reordering transactions that were submitted out of the expected order.

The thing that is **very** debatable in this approach is that currently it introduces 4 bytes for each input that are going to be unused by UTXO inputs. This is by no means a good design and needs to be implemented in a less wasteful way. But good enough for the purpose of demonstrating the approach.


mzabaluev:
Closing this for now, we have #571 and more upcoming changes in the works to enable more flexible ordering of transactions.

PR #566, opened 2021-05-18
Fix lints reported by clippy 0.1.52

mzabaluev:


eugene-babichenko (approved):
Shall we also fix the ones from "Unchanged files with check annotations"?

PR #580, opened 2021-06-22
Vote refactor

iquerejeta:
Major refactor included:

* new cryptography module with zkps, commitment and elgamal
* zkps module contains decryption and unit_vector zkps
* moved code in lib.rs to its own module, under tally.rs
    * all functions are now associated functions to a given structure
    * removed the TallyState structure, which was redundant
    * added documentation
    * Deleted EncryptingVote structure
    * made encrypt_and_prove_vote as an associated function to ElectionPublicKey
* Generalised UnitVectorZkp to be abstracted from election notation
* Added intermediary struct ValidatedTally, to enforce that decryption only happens given valid shares (this will be helpful in the future to handle threshold decryption)
* Enforced that only valid votes can be added to the encrypted tally (this still fails in the mockchain-repo---requires a few more changes)


chain-vote/src/tally.rs
```diff
@@ -1,15 +1,20 @@
...
 use rand_core::{CryptoRng, RngCore};
-use crate::cryptography::PublicKey;
 
 /// Secret key for opening vote
 pub type OpeningVoteKey = MemberSecretKey;
 
+/// A proof of correct decryption share consists of a dleq zkp, where the committee member proves
+/// that the `DecryptionShare` is honestly derived from the `EncryptedTally` and the committee private
+/// key correspondig to its public key without disclosing it.
+pub type ProofOfCorrectShare = DleqZkp;
```
iquerejeta, line 17: This is where I doubt we can use this notation. As discussed, the proof of correct decryption hashes the two points of a ciphertext to compute the challenge, while a Dleq  zkp does not. So we can not directly use a Dleq proof for a proof the correct decryption described in the paper.

iquerejeta:
Superseded by #584

PR #581, opened 2021-06-22
impl IntoIterator for imhamt

enzoc4:
Not a huge deal, but it's nice (and expected) to be able to use `for (k, v) in ...` without having to add `iter()`.


danielSanchezQ (approved):
LGTM

PR #582, opened 2021-06-23
Move EC backends to chain-crypto

iquerejeta:


chain-crypto/src/algorithms/vrf/vrf.rs
```diff
@@ -163,28 +158,28 @@ impl PublicKey {
         if input.len() != PUBLIC_SIZE {
             return Err(PublicKeyError::SizeInvalid);
         }
-        let ristretto = CompressedRistretto::from_slice(&input);
-        match ristretto.decompress() {
+        let ristretto = GroupElement::from_bytes(input);
```
Zeegomo, line 161: Nit: the variable name does not reflect anymore what it is

chain-crypto/src/algorithms/vrf/vrf.rs
```diff
@@ -67,27 +62,27 @@ impl SecretKey {
...
+        let pk = GroupElement::generator() * sk;
         SecretKey {
             secret: sk,
             public: pk,
         }
     }
 
     pub fn as_bytes(&self) -> &[u8] {
-        self.secret.as_bytes()
+        &self.secret.as_bytes()
```
Zeegomo, line 73: I think the `&` at the beginning is redundant

chain-crypto/src/algorithms/vrf/vrf.rs
```diff
@@ -32,7 +27,7 @@ impl AsRef<[u8]> for SecretKey {
 
 /// VRF Public Key
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct PublicKey(Point, CompressedRistretto);
+pub struct PublicKey(GroupElement, [u8; PUBLIC_SIZE]);
```
Zeegomo, line 30: Why does it have both a point and it's compressed form?

chain-crypto/src/algorithms/vrf/vrf.rs
```diff
@@ -244,20 +239,20 @@ impl OutputSeed {
...
         buf.extend_from_slice(suffix);
 
         Blake2b256::new(&buf)
     }
 }
 
-fn make_message_hash_point(data: &[u8]) -> Point {
+fn make_message_hash_point(data: &[u8]) -> GroupElement {
     let m_hash = {
         let mut c = Sha512::new();
```
Zeegomo, line 251: `from_hash` already hashes data in input. I think it's redundant to hash it with `Sha512` before (and also a bit odd since we use blake everywhere else)

Zeegomo (commented):
Have you tried to reuse the dleq proof for decryption share verification? IIRC that one has one more parameter and I was wondering how that would work

chain-crypto/src/algorithms/vrf/vrf.rs
iquerejeta: I thought the same, that it was redundant. The problem is that if the public key is only a `GroupElement`, then we cannot implement the `AsRef` trait. The other option would be to only have it's compressed form, but then every time we use the pk we need to decompress.

chain-crypto/src/algorithms/vrf/vrf.rs
```diff
@@ -163,44 +165,44 @@ impl PublicKey {
...
+            g2: &GroupElement::from_hash(&input),
             h2: &self.u.0,
         };
         dleq::verify(&dleq, &self.dleq_proof)
     }
 
     pub fn to_buffer(&self, output: &mut [u8]) {
         assert_eq!(output.len(), PROOF_SIZE);
-        output[0..32].copy_from_slice(self.u.0.compress().as_bytes());
+        output[0..32].copy_from_slice(&self.u.0.to_bytes());
```
Zeegomo, line 199: This will panic at runtime if `p256k1` feature is selected because `src` (`&self.u.0.to_bytes()`) is not the same length as `self` (`output[0..32]`)

chain-crypto/src/algorithms/vrf/vrf.rs
```diff
@@ -163,44 +165,44 @@ impl PublicKey {
...
         self.dleq_proof.to_bytes(&mut output[32..96]);
     }
 
     pub fn bytes(&self) -> [u8; PROOF_SIZE] {
         let mut output = [0u8; PROOF_SIZE];
-        output[0..32].copy_from_slice(self.u.0.compress().as_bytes());
+        output[0..32].copy_from_slice(&self.u.0.to_bytes());
```
Zeegomo, line 205: same here

chain-crypto/src/algorithms/vrf/vrf.rs
```diff
@@ -209,7 +211,7 @@ impl ProvenOutputSeed {
         if bytes.len() != PROOF_SIZE {
             return None;
         }
-        let u = CompressedRistretto::from_slice(&bytes[0..32]).decompress()?;
+        let u = GroupElement::from_bytes(&bytes[0..32])?;
```
Zeegomo, line 214: This do not work for feature `p256k1`

chain-crypto/src/algorithms/vrf/dleq.rs
```diff
@@ -11,26 +9,26 @@ pub struct Proof {
...
-
 impl Proof {
+    pub const PROOF_SIZE: usize = Scalar::BYTES_LEN * 2;
+
     pub fn to_bytes(&self, output: &mut [u8]) {
-        assert_eq!(output.len(), PROOF_SIZE);
-        output[0..32].copy_from_slice(self.c.0.as_bytes());
-        output[32..64].copy_from_slice(self.z.as_bytes());
+        assert_eq!(output.len(), Self::PROOF_SIZE);
+        output[0..32].copy_from_slice(&self.c.0.to_bytes());
```
Zeegomo, line 17: I would rather use `Scalar::BYTES_LEN` instead of hardcoding `32`. If we change the length of the non-default backend such error would be caught very late.
Same thing applies to wherever we use such constants.

chain-vote/src/encrypted.rs
```diff
@@ -0,0 +1,108 @@
...
+        }
+        Self {
+            unit_vector: *vote,
+            ciphertexts: ciphers,
+            random_elements: rs,
+        }
+    }
+
+    /*
+    pub fn pad<F>(mut self, extended_value: F) -> PtpEncryptingVote
```
Zeegomo, line 88: leftover?

chain-vote/src/error.rs
```diff
@@ -0,0 +1,93 @@
+//! Errors related to chain-vote.
```
Zeegomo, line 1: Was this file included on purpose?

Zeegomo (changes requested):

chain-crypto/src/algorithms/vrf/dleq.rs
iquerejeta: great, I've added tests to ensure these would have been cached

chain-crypto/src/algorithms/vrf/dleq.rs
```diff
@@ -11,26 +9,26 @@ pub struct Proof {
...
-        let mut c_array = [0u8; 32];
-        c_array.copy_from_slice(&slice[0..32]);
-        let c = Scalar::from_canonical_bytes(c_array)?;
+        let mut c_array = [0u8; Scalar::BYTES_LEN];
+        c_array.copy_from_slice(&slice[0..Scalar::BYTES_LEN]);
+        let c = Scalar::from_bytes(&c_array)?;
 
-        let mut z_array = [0u8; 32];
-        z_array.copy_from_slice(&slice[32..64]);
-        let z = Scalar::from_canonical_bytes(z_array)?;
```
Zeegomo, line 33: The old implementation was enforcing the use of canonical bytes but the p256k1 backend implementation of `Scalar::from_bytes` does not, is this of concern?

chain-crypto/src/multilock.rs
```diff
@@ -268,7 +264,7 @@ mod test {
         let encrypted = encrypt(&mut r, app_info, &participant_pks, msg);
         for (i, sk) in participants.iter().enumerate() {
             let mut out = vec![0; msg.len()];
-            //let pk = RISTRETTO_BASEPOINT_POINT * sk;
+            //let pk = BASEPOINT_POINT * sk;
```
Zeegomo, line 267: leftover?

chain-crypto/src/algorithms/vrf/dleq.rs
iquerejeta: p256k1 also returns an error if the scalar is non-canonical (higher than the group order). You can find it [here](https://github.com/vincenthz/eccoxide/blob/master/src/curve/fiat/field_macros.rs#L404)

Zeegomo (approved):

chain-crypto/src/algorithms/vrf/vrf.rs
mzabaluev: Could `PUBLIC_SIZE` be an associated const: `PublicKey::SIZE`?

chain-crypto/src/ec/mod.rs
```diff
@@ -1,9 +1,14 @@
-mod babystep;
+//! Module defining the Group Elements and Scalar structures in one primer order group (over sec2
+//! curves), or the other (ristretto255).
+#[macro_use]
+mod macros;
 #[cfg(not(feature = "ristretto255"))]
```
mzabaluev, line 5: I don't like the idea of an absence of feature bringing in a different backend implicitly.

Maybe there should be backend selection features `ec-p256k1` and `ec-ristretto255`, where the first one is listed among default features?

chain-vote/src/lib.rs
```diff
@@ -1,27 +1,28 @@
+// These two need to be here because the points of sec2 backend do not implement Copy
+#![allow(clippy::clone_on_copy)]
+#![allow(clippy::op_ref)]
```
mzabaluev, line 3: This should be redundant here now?
I don't think the blanket clippy overrides for the whole library is a good idea, too. It's better to keep them close in the modules where the overrides are needed (and maybe think twice whether they should be removed by redesign, as long as we're changing the API anyway)

chain-vote/src/tally.rs
```diff
@@ -156,6 +162,37 @@ impl EncryptedTally {
     }
 }
 
+impl ValidatedTally {
+    /// Given the shares of the committee members, returns the decryption of all the
+    /// election options in the form of `GroupElements`. To get the final results, one
+    /// needs to compute the discrete logarithm of these values, which is performed in
+    /// `decrypt_tally`.
```
mzabaluev, line 169: Good doc; for a moment, I was confused by this being a doc comment, thinking that you are adding a new public method. For private methods, plain comments should be made as a rule.

mzabaluev (commented):
Good changes, I have a few objections about management of features and minor points on code quality.

chain-vote/src/lib.rs
iquerejeta: Why redundant? You mean that not both are needed? As far as I can tell, the first (`clone_on_copy`) is required for when you clone a structure that implements copy (and is therefore not required). The second, is when you unnecessarily use a reference. 

I also agree that it's better to no override the whole library, but in this case, it seems required. The problem is that the two ec backends differ in the `GroupElement` and `Scalar` structure. `eccoxide` does not implement the copy trait, so we need to clone everything, and use references rather than values. On the other hand, `curve25519_dalek` does. Given that the latter is our default feature, it will complain. And the reason to have it in the library is because this flag is needed in almost every operation where `GroupElements` or `Scalar`. However, if it is preferable, and can put it in every module individually 👍

chain-crypto/src/ec/mod.rs
iquerejeta: I'm not sure I understood this. Do you mean that we should do `#[cfg(feature = "p256k1")]` instead of `#[cfg(not(feature = "ristretto255"))]`?

Zeegomo: there are already different features for the two backends with p256k1 being among default ones. AFAIU The reason for `#[cfg(not(...))` is to guarantee that one and only one backend is selected at a time

chain-vote/src/lib.rs
mzabaluev: OK then, leave it.
Would it be annoying to not implement `Copy` instead?

chain-vote/src/tally.rs
```diff
@@ -243,8 +243,8 @@ impl TallyDecryptShare {
         self.elements.len()
     }
 
-    /// Size of the byte representation for a tally decrypt share
-    /// with the given number of options.
+    // Size of the byte representation for a tally decrypt share
+    // with the given number of options.
```
mzabaluev, line 247: Now this is a public method, one edit too many? :)

mzabaluev (approved):

mzabaluev:
There are merge conflicts now, probably caused by #587.

chain-vote/src/lib.rs
iquerejeta: very good point 👍  much cleaner now.

chain-crypto/src/algorithms/vrf/mod.rs
```diff
@@ -1,3 +1,7 @@
+// These two need to be here because the points of sec2 backend do not implement Copy
+#![allow(clippy::clone_on_copy)]
```
mzabaluev, line 2: Do these overrides still need to be present?

mzabaluev (approved):

PR #584, opened 2021-06-25
Chain vote modules refactor

Zeegomo:
This is mostly a non functional refactor for modules and names in the chain-vote crate


chain-impl-mockchain/src/vote/committee.rs
```diff
@@ -22,7 +22,7 @@ impl CommitteeId {
     pub const COMMITTEE_ID_SIZE: usize = 32;
 
     /// returns the identifier encoded in hexadecimal string
-    pub fn to_hex(&self) -> String {
+    pub fn to_hex(self) -> String {
```
iquerejeta, line 25: doesn't it make sense to take by reference instead here?

Zeegomo: Kind of, but clippy complains about naming with `to_*` and the type being copy

iquerejeta (approved):
Looks good 👍

PR #586, opened 2021-06-29
Typestate enforced ValidatedTally

iquerejeta:
Ensures that a tally is decrypted only with valid shares. This structure will be of particular interest when we progress with the DKG, as committee members might have to perform actions between verification and decryption.

### DIFF
--- a/chain-impl-mockchain/src/testing/builders/vote.rs
+++ b/chain-impl-mockchain/src/testing/builders/vote.rs
@@ -28,6 +28,12 @@ pub fn decrypt_tally(
         .unwrap();
     let table = chain_vote::TallyOptimizationTable::generate_with_balance(absolute_max_votes, 1);
 
+    let members_pks: Vec<chain_vote::MemberPublicKey> = members
+        .members()
+        .iter()
+        .map(|member| member.public_key())
+        .collect();
+
     let proposals = encrypted_tally
         .into_iter()
         .map(|(encrypted_tally, max_votes)| {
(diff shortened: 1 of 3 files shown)


chain-impl-mockchain/src/testing/builders/vote.rs
```diff
@@ -37,9 +43,10 @@ pub fn decrypt_tally(
                 .map(|member| member.secret_key())
                 .map(|secret_key| encrypted_tally.partial_decrypt(&mut thread_rng(), secret_key))
                 .collect::<Vec<_>>();
-            let tally = encrypted_tally
-                .decrypt_tally(max_votes, &decrypt_shares, &table)
-                .unwrap();
+            let validated_tally = encrypted_tally
+                .validate_partial_decryptions(&members_pks, &decrypt_shares)
+                .expect("Invalid shares");
```
danielSanchezQ, line 48: I think this should leverage some kind of error instead of just unwraping here. It may lead an attack in where  some fake shares are sent and then node just crash(?)

Zeegomo: This is testing code where we directly control committee members.

chain-vote/src/tally.rs
```diff
@@ -37,8 +37,23 @@ pub struct TallyDecryptShare {
...
-// `ProvenDecryptShare` consists of a group element (the partial decryption), and `ProofOfCorrectShare`,
-// a proof of correct decryption.
+/// `ValidatedTally` can only be constructed by valid `TallyDecryptShare`s, and the
+/// corresponding `EncryptedTally`. This intermediate structure ensures that only
+/// validated decryptions are used to compute the election outcome, i.e. if the
+/// committee members do not present valid shares, the tally decryption cannot be
+/// computed.
+/// This intermediate structure is particularly of interest during the distributed
+/// decryption protocol, where, in case there is a misbehaving party, one needs to
+/// perform certain actions between the verification of a vote, and its inclusion in
```
Zeegomo, line 47: Shouldn't it be 
> between the verification of a decryption share, and its inclusion in the tally?

votes are already verified through other means

PR #587, opened 2021-06-30
ditch custom eccoxide version

Zeegomo:
Port the fast u64 trick to chain-vote to avoid maintaining a branch of eccoxide.
There is a little cost in performance for doing so (~18% for building the encrypted tally), but ristretto will probably be the default backend in the future anyway


mzabaluev (approved):
:raised_hands:

PR #588, opened 2021-06-30
Test crypto backends

Zeegomo:
Now that we have the choice of elliptic curve backends for curve-agnostic protocols, we should add tests for both in the CI to ensure we do not forget about the non-default one and it does not lag behind when we add new features.

Changelog:
* Reversed the precedence of compile time features in `chain-crypto`. Rust features were not meant to be used this way but this is required when we change the default feature to be able to select the secondary backend with `--feature p256k1`.
* Rename the Vrf function we use from `Cruve25519_2HashDh` to `EllipticCurve2hashDh`. The curve used will now depend on the backend chosen so let's remove any reference to it from the name (+ clippy camel case suggestions).
* Adapt constants in `chain-impl-mockchain/src/hear/cstruct.rs` to the backend used. I'm not really sure on this one. It seems like a lot of those constants are defined independently from the underlying objects used (i.e. raw bytes sequences with a hardcoded length) and I do not understand the rationale there. 
Should we tie at least the length of those byte sequences to the object they represent like
```
type BftSignature = [u8; Signature<_, Ed25519>::SIGNATURE_SIZE]
```
instead of 
```
type BftSignature = [u8; 64]
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,7 +70,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         toolchain: [stable]
-        mode: ['', '--release', '--features p256k1']
+        mode: ['', '--release', '--features chain-crypto/p256k1']
         experimental: [false]
         include:
           - os: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,7 +70,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         toolchain: [stable]
-        mode: ['', '--release']
+        mode: ['', '--release', '--features p256k1']
         experimental: [false]
         include:
           - os: ubuntu-latest
@@ -153,6 +153,7 @@ jobs:
           args: >-
             ${{ matrix.mode }} --locked
             --manifest-path chain-network/Cargo.toml --no-default-features
+
 
   lints:
     name: Lints
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,12 +70,16 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         toolchain: [stable]
-        mode: ['', '--release', '--features chain-crypto/p256k1']
+        mode: ['', '--release']
         experimental: [false]
         include:
           - os: ubuntu-latest
             toolchain: nightly
             experimental: true
+          - os: ubuntu-latest
+            toolchain: stable
+            experimental: false
+            mode: '--features chain-crypto/p256k1'
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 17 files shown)


chain-crypto/src/algorithms/vrf/mod.rs
```diff
@@ -11,9 +11,9 @@ use rand_core::{CryptoRng, RngCore};
 pub use vrf::ProvenOutputSeed;
 
 /// VRF
-pub struct Curve25519_2HashDh;
```
eugene-babichenko, line 14: Yes, that's definitely better naming :)

eugene-babichenko (approved):

chain-crypto/src/asymlock.rs
```diff
@@ -20,10 +20,10 @@ pub enum DecryptionError {
 
 fn shared_key_to_symmetric_key(app_level_info: &[u8], p: &GroupElement) -> ChaCha20Poly1305 {
     // use the compressed point as PRK directly
-    #[cfg(feature = "ristretto255")]
+    #[cfg(not(feature = "p256k1"))]
```
mzabaluev, line 23: As I wrote in another PR, I don't like features that implicitly negate alternative dependencies. This is also not scalable: if we add a third backend in the future, we'd have to expand this tedious exclusive logic everywhere.

Can we have both `p256k1` and `ristretto255`, with one enabled by default? It would be OK to have the compilation break if both are enabled, so the test configuration would use `--no-default-features --features p256k1`.

.github/workflows/main.yml
```diff
@@ -70,7 +70,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         toolchain: [stable]
-        mode: ['', '--release']
```
mzabaluev, line 73: This expands the test matrix by 50%.

Let's treat the p256k1 as a "tier 2" configuration, and only test if it compiles and runs tests on a single Linux runner? I presume the implementation is pure Rust, so we don't get much by testing it on other OSes.

chain-crypto/src/asymlock.rs
Zeegomo: Dumb question, is it possible to use `--no-default-features` while targeting only one package in a workspace like I can with `--features chain-crypto/p256k1`?

Zeegomo: I'm having a hard time making this work in a workspace. Maybe using https://crates.io/crates/cfg-if would make selecting the backend at compile time better?

Zeegomo:
I've offloaded the choice of the backend to a `build.rs` script so that we don't have to write custom exclusive logic everywhere in the code where we need different behavior based on the backend.
The behavior is still the original one though (one feature might disable the other one), as I haven't find a way to disable default features for a single crate in a workspace

mzabaluev (approved):
Looks good, let's leave the improvements at that.

PR #589, opened 2021-06-30
disallow multiple ballots for same proposal from same address

eugene-babichenko:
This change opens the way for incremental tally implementation that would resolve several performance issues.

### DIFF
--- a/chain-impl-mockchain/src/vote/manager.rs
+++ b/chain-impl-mockchain/src/vote/manager.rs
@@ -79,6 +79,9 @@ pub enum VoteError {
         vote: VoteCast,
     },
 
+    #[error("This account already voted for this proposal")]
+    AlreadyVoted,
+
     #[error("Invalid vote proposal, only {num_proposals} available in the vote plan")]
     InvalidVoteProposal {
         num_proposals: usize,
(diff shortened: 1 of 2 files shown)


chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -141,10 +144,13 @@ impl ProposalManager {
         identifier: UnspecifiedAccountIdentifier,
         payload: ValidatedPayload,
     ) -> Result<Self, VoteError> {
-        // we don't mind if we are replacing a vote
-        let votes_by_voters =
-            self.votes_by_voters
-                .insert_or_update_simple(identifier, payload.clone(), |_| Some(payload));
+        // Part of DDoS protection: do not record a new ballot if the account already voted for this
+        // proposal. This protects the system from flooding in a system with cheap/free voting
```
Zeegomo, line 148: Does this on its own provide any advantage for DDoS protection?

Zeegomo (approved):

chain-impl-mockchain/src/vote/manager.rs
eugene-babichenko: Not right now, but flooding the system with encrypted votes would quite resource-heavy due to NIZK validation.

eugene-babichenko: Another concern there is not getting non-malicious transactions into a block & having the blockchain storage full of garbage.

mzabaluev (approved):

PR #590, opened 2021-07-01
uniform keys API

Zeegomo:
Uniform API across different key structs and expose `MemberCommunicationPublicKey::from_bytes`


iquerejeta (approved):
If it is a matter of unifying style, I think I prefer implementing `From` rather than `to_*`. But it is a matter of personal taste, so no objections. Looks good to me

PR #592, opened 2021-07-05
Generalise dleq

iquerejeta:
As agreed with the research team, a proof of correct ElGamal decryption can be generated through a proof of discrete logarithm equality. This allows us to unify the ZKP used in VRF and the one used for ElGamal decryption. We slightly modify the proof of correct decryption in `chain-crypto` and make the corresponding modifications to the proof of correct decryption in `chain-vote`. 

Similarly, we define proof of share generation (which is slightly different to that of correct decryption), and use it in the tally protocol.


chain-crypto/src/algorithms/vrf/vrf.rs
```diff
@@ -176,30 +160,27 @@ impl PublicKey {
...
+            &self.u.0,
+        )
     }
 
     pub fn to_buffer(&self, output: &mut [u8]) {
         assert_eq!(output.len(), Self::BYTES_LEN);
         output[0..GroupElement::BYTES_LEN].copy_from_slice(&self.u.0.to_bytes());
         self.dleq_proof
-            .to_bytes(&mut output[GroupElement::BYTES_LEN..]);
+            .to_mut_slice(&mut output[GroupElement::BYTES_LEN..]);
```
mzabaluev, line 178: I don't like the `to_*` names, both before and after: `to_*` is conventionally used for conversion methods that return a value (most `to_bytes` methods elsewhere conform as they return a fixed-size array).

Can we use the opportunity to start renaming these slice-filling methods to something like `write_to_bytes`?

chain-crypto/src/algorithms/vrf/vrf.rs
```diff
@@ -208,7 +189,7 @@ impl ProvenOutputSeed {
             return None;
         }
         let u = GroupElement::from_bytes(&bytes[0..GroupElement::BYTES_LEN])?;
-        let proof = dleq::Proof::from_bytes(&bytes[GroupElement::BYTES_LEN..])?;
+        let proof = dleq::Zkp::from_slice(&bytes[GroupElement::BYTES_LEN..])?;
```
mzabaluev, line 192: Bikeshedding: I think `from_bytes` is better as it's clear we mean to convert from a byte array in this serialization API, while `from_slice` typically is used for slices of elements of some non-primary type.
We use `from_bytes` everywhere, let's keep to the convention.

chain-vote/src/cryptography/zkps/correct_share_generation/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod correct_share_generation_zkp;
```
mzabaluev, line 1: This internal mod and its source file could be named `zkp` for brevity. The IDE tools usually disambiguate if more than one `zkp.rs` is open.

mzabaluev (commented):
Looks good, I only have remarks on style.

chain-vote/src/tally.rs
```diff
@@ -1,6 +1,6 @@
 use crate::{
     committee::*,
-    cryptography::{Ciphertext, CorrectElGamalDecrZkp},
+    cryptography::{Ciphertext, CorrectShareGenerationZkp},
```
Zeegomo, line 3: I thought we tried not to tie too much the stuff in cryptography with vote-specific domain (e.g. `UnitVectorProof` instead of `CorrectVoteZkp`). I don't have a strong opinion for this, but I would like to be consistent across proofs

iquerejeta: I agree. But `CorrectShareGenerationZkp` is not vote-specific. It is specific to threshold decryption, so I think it should be fine (note that there is a difference between `CorrectElGamalDecrZkp` and `CorrectShareGenerationZkp` in the proof itself, not only in the naming).

mzabaluev (approved):

Issue #593, opened 2021-07-08
Unalias ECC backends in chain-crypto


mzabaluev:
So now we have explicitly selected algorithms for most cryptographic things in jormungandr, but an alias for the ECC backend, selected by chain-crypto feature, which affects both the vote protocol and Praos.

This diverges from the general approach of providing generic crypto APIs in the libraries and selecting specific implementations in chain-impl-mockchain and jormungandr.

Can we export both ECC backends in chain-libs unaliased side by side, so that jormungandr can select one of them explicitly? This will remove the issue with being unable to build the library with `--all-features` because of the conflicting backend aliases.

_Originally posted by @mzabaluev in https://github.com/input-output-hk/jormungandr/pull/3404#discussion_r665976115_

mzabaluev:
Cc @Zeegomo @iquerejeta

PR #594, opened 2021-07-08
De alias crypto

Zeegomo:
closes https://github.com/input-output-hk/chain-libs/issues/593


chain-crypto/src/algorithms/mod.rs
```diff
@@ -13,4 +12,4 @@ pub use ed25519::Ed25519;
 pub use ed25519_derive::Ed25519Bip32;
 pub use ed25519_extended::Ed25519Extended;
 pub use sumed25519::SumEd25519_12;
-pub use vrf::EllipticCurve2hashDh;
+pub use vrf::Curve25519_2HashDh;
```
mzabaluev, line 15: Right, can you deuglify the names while we are at it?
Maybe a submodule namespace like `vrf::curve25519::Curve2HashDh` could help.

iquerejeta: I would use `ristretto` instead of `curve25519`, as it is more representative of what we use.

chain-crypto/src/algorithms/vrf/dleq.rs
```diff
@@ -0,0 +1,145 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct Challenge(Scalar);
+
+fn challenge(
+    h1: &GroupElement,
+    h2: &GroupElement,
+    a1: &GroupElement,
+    a2: &GroupElement,
+) -> Challenge {
```
iquerejeta, line 54: challenge should be computed using the two bases and the two points (`g1, g2, h1, h2`) together with the two announcements (`a1, a2`). I would keep the latest version of the ZKP we had, even if we repeat the code in `chain-vote`.

chain-crypto/src/algorithms/vrf/vrf.rs
```diff
@@ -100,34 +100,50 @@ impl SecretKey {
...
+            h2: &output.0,
+        };
+        let dleq_proof = dleq::generate(&r, &self.secret, &dleq);
+        ProvenOutputSeed {
+            u: output.clone(),
+            dleq_proof,
+        }
+    }
+
+    pub fn prove_simple<T: RngCore + CryptoRng>(
```
iquerejeta, line 121: I would use the latest version of this as well. Remove the possibility of inputing the randomness at call time. This should never be required.

chain-crypto/src/algorithms/vrf/vrf.rs
```diff
@@ -100,34 +100,50 @@ impl SecretKey {
...
-            dleq_proof,
-        }
+        let w = Scalar::random(rng);
+        self.prove(&w, m_point, output)
     }
 
     /// Generate a Proof
     ///
-    pub fn evaluate<T: RngCore + CryptoRng>(&self, rng: &mut T, input: &[u8]) -> ProvenOutputSeed {
+    /// the proof is randomized, so need a freshly randomly scalar for random.
```
iquerejeta, line 133: Same as above. We should not allow for inputing the ZKP randomness.

chain-crypto/src/algorithms/vrf/vrf.rs
```diff
@@ -160,27 +176,30 @@ impl PublicKey {
...
+        };
+        dleq::verify(&dleq, &self.dleq_proof)
     }
 
     pub fn to_buffer(&self, output: &mut [u8]) {
         assert_eq!(output.len(), Self::BYTES_LEN);
         output[0..GroupElement::BYTES_LEN].copy_from_slice(&self.u.0.to_bytes());
         self.dleq_proof
-            .write_to_bytes(&mut output[GroupElement::BYTES_LEN..]);
+            .to_bytes(&mut output[GroupElement::BYTES_LEN..]);
```
iquerejeta, line 195: Also, we got the `to_*` functions renamed for keeping consistency across crates.

chain-crypto/src/ec/macros.rs
```diff
@@ -27,3 +27,113 @@ macro_rules! std_ops_gen {
...
+                            let g = (GroupElement::generator() * e1) * e1.inverse();
+                            property::equal(g, GroupElement::generator())
+                        })
+                        .test(ctx);
+                        */
+                })
+            }
+        }
+    };
+}
```
iquerejeta, line 139: Nice, better place to have this indeed 👍

chain-impl-mockchain/src/header/cstruct.rs
```diff
@@ -20,8 +19,7 @@ pub(super) type BftLeaderId = [u8; 32];
 pub(super) type BftSignature = [u8; 64];
 
 pub(super) type GpNodeId = [u8; 32];
-pub(super) type GpVrfProof =
-    [u8; <EllipticCurve2hashDh as VerifiableRandomFunction>::VerifiedRandomOutput::BYTES_LEN];
+pub(super) type GpVrfProof = [u8; 96];
```
iquerejeta, line 22: I would keep using the constant of the VRF proof. If we ever decide to implement the VRF as defined in the IETF standard, this size would change.

chain-vote/Cargo.toml
```diff
@@ -26,3 +29,8 @@ name = "shvzk"
 [[bench]]
 harness = false
 name = "curve"
+
+[features]
+default = ["ristretto255"]
+ristretto255 = []
+p256k1 = ["chain-crypto/p256k1"]
```
iquerejeta, line 36: new line?

chain-vote/src/cryptography/zkps/correct_decryption/correct_decryption_zkp.rs
```diff
@@ -0,0 +1,116 @@
+//! Non-interactive Zero Knowledge proof for correct ElGamal
```
iquerejeta, line 1: The Treasury voting spec is going to be modified so that the proof of ElGamal decryption will be the same as the DLEQ proof. I would keep the latest docs, and (unfortunately) repeat the DLEQ proof code here.

iquerejeta (commented):
Unfinished review, but I would not revert everything to how it was before.

chain-impl-mockchain/src/header/cstruct.rs
Zeegomo: The thing is that no other constant in that file is bound to anything, so I thought to remove it not to break consistency, but either way is fine for me.

iquerejeta (approved):

mzabaluev (approved):

PR #600, opened 2021-07-12
Changes on hybrid encryption

iquerejeta:
Following the changes of the [spec](https://github.com/input-output-hk/treasury-crypto/commit/bb2c281df96bbbc28c5f7bd8622946fd07029056#diff-eda0c8f7bb9d0d01d564eb29a764a9cce7a40c56bf174ff61d63a472ca1c0ef8), we've modified the hybrid encryption algorithm, and the zkp of correct symmetric key generation.


chain-vote/src/cryptography/zkps/mod.rs
```diff
@@ -1,9 +1,12 @@
-// Correct encryption will be used once we proceed with DKG protocol
+// CorrectElGamalDecrZkp used for disclosing randomness by committee members
 // mod correct_decryption;
+// CorrectHybridDecrKeyZkp used for DKG algorithm
+// mod correct_hybrid_decryption_key;
```
Zeegomo, line 4: why is this commented?

iquerejeta: linter complains that the code is not being used

Zeegomo (commented):
Why are the files containing the hybrid decryption zkp not included in the module system? I would rather avoid that, as it will likely be forgotten in future refactors since it is not checked by the CI. 

We can add a `#[allow(dead_code)]` attribute with a brief explanation of its purpose if the problem is the linter complaining

chain-vote/src/cryptography/elgamal.rs
```diff
@@ -245,6 +254,21 @@ impl Ciphertext {
     }
 }
 
+impl HybridCiphertext {
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut r = Vec::new();
```
mzabaluev, line 259: `Vec::with_capacity` would be better for performance.

Zeegomo (approved):

PR #601, opened 2021-07-14
Added test coverage CI job

mrzenioszeniou:
Introduced a simple job to run [tarpaulin](https://github.com/xd009642/tarpaulin) and expose the resulting report as an artifact.

It uses the [actions-rs/tarpaulin](https://github.com/actions-rs/tarpaulin) action and the caching scheme that is already present in the workflow. The artifact name includes part of the commit's SHA hash to allow distinction between different reports.


.github/workflows/main.yml
```diff
@@ -207,3 +207,49 @@ jobs:
...
+        with:
+          path: ~/.cargo/registry/cache
+          key: cargo-deps-v2-${{ hashFiles('Cargo.lock') }}
+
+      - name: Run tarpaulin
+        uses: actions-rs/tarpaulin@v0.1
+        with:
+          out-type: Html
+
+      - run: mv tarpaulin-report.html test_coverage_$SHA_SHORT.html
```
eugene-babichenko, line 249: I would reflect a branch name & commit time as well. It would be easier to navigate through artifacts that way.

.github/workflows/main.yml
```diff
@@ -207,3 +207,49 @@ jobs:
...
+
+      - name: Upload coverage report
+        uses: actions/upload-artifact@v2
+        with:
+          name: test_coverage_$SHA_SHORT
```
eugene-babichenko, line 254: You need to use `${{ env.SHA_SHORT }}` outside of shell scripts.

.github/workflows/main.yml
eugene-babichenko: You can actually drop this extra step by doing

```yaml
- uses: actions/upload-artifact@v2
  with:
    name: test_coverage_${{ env.SHA_SHORT }}
    path: tarpaulin-report.html
```

mzabaluev: The artefact is attached to the workflow run, which has the links to GitHub.

mrzenioszeniou: @eugene-babichenko I believe that skipping the `mv` would leave the actual file name as is, meaning all the reports would be `tarpaulin-report.html` after unzipping. That defeats the purpose of including the SHA and branch name in the artifact, don't you think?

@mzabaluev I'm not sure what you mean. The idea with the naming is to be able to distinguish them on our local filesystems or in case they're shuffled around over Slack, email, and whatnot. 

Branch name would be nice to have 👌 I'd replace any character that is not alphanumeric, `_`, or `-` with `-` for a cleaner name. I'm not so keen on the timestamp since it makes the file name quite verbose – unless it's something compact. For example:
```
test_coverage_feature-branch-name_20210713T112706Z_3df40ba2.html
```

What do you think?

.github/workflows/main.yml
mrzenioszeniou: Done, thanks

mzabaluev (approved):
Looks good.

It could be good to also obtain coverage for the entire OS matrix, but there is not much value added as we don't have many target-specific differences in the chain-libs code.

eugene-babichenko (approved):

PR #605, opened 2021-07-21
chain-evm: implement basic storage backend

eugene-babichenko:


chain-evm/src/state/trie.rs
```diff
@@ -0,0 +1,78 @@
...
+        // using two branches instead of `Hamt::insert_or_update` to avoid unnecessary cloning
+        let new_state = if self.0.contains_key(&key) {
+            self.0
+                .update(&key, |_| Ok::<_, Infallible>(Some(value)))
+                .expect("we already checked that the key is present")
+        } else {
+            self.0
+                .insert(key, value)
+                .expect("we already checked that the key does not exist")
+        };
```
mzabaluev, line 36: Could use `.insert_or_update`.

eugene-babichenko: I left a comment in the beginning of this function as per why I did not use it. It would result in a code like `self.0.insert_or_update(key, value.clone(), |_| Some(value))`. Cloning is completely unnecessary here and should be avoided IMO.

mzabaluev: Hmm, there should be some Entry-like API to Hamt to deal with this.

mzabaluev (approved):

PR #607, opened 2021-07-23
[Test] Vote manager conflict payloads

dkijania:
Increased coverage in vote/manager module and added 3 new test cases on api level 1 for public vote 2 for private vote.

### DIFF
--- a/chain-impl-mockchain/src/testing/e2e/vote_private.rs
+++ b/chain-impl-mockchain/src/testing/e2e/vote_private.rs
@@ -1,6 +1,8 @@
 use crate::testing::decrypt_tally;
 use crate::testing::TestGen;
 use crate::testing::VoteTestGen;
+use crate::vote::VoteError::AlreadyVoted;
+use crate::vote::VotePlanLedgerError::VoteError;
 use crate::{
     certificate::VotePlan,
     fee::LinearFee,
@@ -13,6 +15,7 @@ use crate::{
     value::Value,
     vote::{Choice, PayloadType},
 };
+use imhamt::UpdateError::ValueCallbackError;
 
 const ALICE: &str = "Alice";
 const STAKE_POOL: &str = "stake_pool";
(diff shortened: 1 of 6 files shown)


chain-impl-mockchain/src/testing/e2e/vote_private.rs
```diff
@@ -143,3 +132,129 @@ pub fn private_vote_plan_without_keys() {
...
+    assert!(controller
+        .cast_vote_private(
+            &alice,
+            &vote_plan,
+            &proposal.id(),
+            favorable,
+            &mut ledger,
+            &mut rng
+        )
+        .is_err());
```
mzabaluev, line 193: Could have matched the error case to check it is as expected.

chain-impl-mockchain/src/testing/e2e/vote_public.rs
```diff
@@ -376,6 +376,67 @@ pub fn vote_on_same_proposal() {
         .is_err());
 }
 
+#[test]
+pub fn vote_on_differenet_proposal() {
```
mzabaluev, line 380: Typo:
```suggestion
pub fn vote_on_different_proposal() {
```

chain-impl-mockchain/src/testing/gen/vote.rs
```diff
@@ -83,6 +86,32 @@ impl VoteTestGen {
...
+        CommitteeMembersManager::new(&mut rng, crs_seed, threshold, members_no)
+    }
+
+    pub fn private_vote_plan() -> VotePlan {
+        Self::private_vote_plan_with_committees_manager(&Self::committee_members_manager(3, 1))
+    }
+
+    pub fn private_vote_plan_with_committees_manager(
+        manager: &CommitteeMembersManager,
+    ) -> VotePlan {
```
mzabaluev, line 104: Now this could be more elegant a method `with_committee_manager` modifying `self` in a builder pattern, but the rest of the impl of this type defines `self`-less methods. The `VoteTestGen` type currently has no purpose, for namespacing a module is more typically used in Rust.

dkijania: can i propose to change it in another pr as VoteTestGen is spread around several crate?

PR #608, opened 2021-07-26
[Test] Add API tests

mrzenioszeniou:
Introduced two API tests:
1. Verifies that after a transaction in a ledger **without** fees, the total funds do not change and the fee pots remain empty.
2. Verifies that after a transaction in a ledger **with** fees, the total funds do not change and the fee pots contain the fee.

### DIFF
--- a/chain-impl-mockchain/src/testing/e2e/fees/mod.rs
+++ b/chain-impl-mockchain/src/testing/e2e/fees/mod.rs
@@ -182,3 +182,91 @@ pub fn owner_delegates_fee() {
 
     ledger_verifier.total_value_is(&Value(expected_total_funds_after));
 }
+
+#[test]
+/// Verifies that after a transaction in a ledger without fees, the total funds do not change and
+/// the fee pots remain empty.
+fn transaction_without_fees() {
+    const ALICE: &str = "Alice";
+    const ALICE_FUNDS: u64 = 42;
+
+    const BOB: &str = "Bob";
+    const BOB_FUNDS: u64 = 13;
+
+    const TRANSFER: u64 = 10;
+
+    let (mut ledger, controller) = prepare_scenario()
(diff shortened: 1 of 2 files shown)


chain-impl-mockchain/src/testing/e2e/fees/mod.rs
```diff
@@ -182,3 +182,91 @@ pub fn owner_delegates_fee() {
 
     ledger_verifier.total_value_is(&Value(expected_total_funds_after));
 }
+
+#[test]
+/// Verifies that after a transaction in a ledger without fees, the total funds do not change and
+/// the fee pots remain empty.
+fn transaction_without_fees() {
```
dkijania, line 189: Looks like those two methods are almost the same in term of the flow. the only diff is FEE value. Maybe we should introduce helper method with signature:

`fn verify_total_funds_after_transaction_with_fee(fee: u64)`

and then use it for two cases (non zero fees and zero fees).

chain-impl-mockchain/src/testing/e2e/fees/mod.rs
```diff
@@ -182,3 +182,91 @@ pub fn owner_delegates_fee() {
...
+    const ALICE: &str = "Alice";
```
dkijania, line 190: This is my mistake, done in some earlier PR, but you can fix it in this one. Please consider moving ALICE and BOB consts to the module level and use it in all tests in current module

dkijania (changes requested):

dkijania:
Apart above minor remarks code looks very good

mrzenioszeniou (commented):
@dkijania, I noticed that`chain-impl-mockchain/src/testing/e2e/fees` was redundant since it only contained a `mod.rs` file. I moved it up to `e2e` to get rid of the unnecessary nesting.

dkijania (approved):

PR #609, opened 2021-07-30
WIP: Prototype EVM abstractions

saibatizoku:
Add code to chain-evm crate in chain-libs abstracting execution of EVM smart contracts, wrapping the details of SputnikVM integration in a way that will allow an alternative implementation to be swapped in later if needed.

It can expose the EVM data types such as addresses, hashes, bytecode payloads etc., and it can pass and return account state around as HAMTs, but the details of runtime initialization and running of a smart contract should be abstracted away from the user.

### DIFF
--- a/chain-evm/Cargo.toml
+++ b/chain-evm/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2018"
 
 [dependencies]
+evm = "0.28.0"
 imhamt = { path = "../imhamt" }
 primitive-types = "0.10.1"
 
--- a/chain-evm/Cargo.toml
+++ b/chain-evm/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2018"
 [dependencies]
 evm = "0.28.0"
 imhamt = { path = "../imhamt" }
-primitive-types = "0.10.1"
+primitive-types = "0.9.1"
 
 [dev-dependencies]
 proptest = "1.0.0"
--- a/chain-evm/Cargo.toml
+++ b/chain-evm/Cargo.toml
@@ -4,9 +4,9 @@ version = "0.1.0"
 edition = "2018"
 
 [dependencies]
-evm = "0.28.0"
+evm = "0.30.0"
 imhamt = { path = "../imhamt" }
-primitive-types = "0.9.1"
+primitive-types = "0.10.1"
 
 [dev-dependencies]
 proptest = "1.0.0"
--- a/chain-evm/src/machine.rs
+++ b/chain-evm/src/machine.rs
@@ -1,0 +1,8 @@
+//! Abstractions for the EVM runtime.
+//!
+//! This module contains types that abstract away implementation details from the `evm` crate,
+//! specially those involving node and stack configurations, and runtime execution.
+
+pub fn code_to_execute_evm_runtime() -> Result<(), ()> {
+    todo!("put together the puzzle of types needed to run evm code");
+}
--- a/chain-evm/src/machine.rs
+++ b/chain-evm/src/machine.rs
@@ -18,6 +18,7 @@ use primitive_types::{H160, H256, U256};
 
 use crate::state::AccountTrie;
 
+/// Environment values for the machine backend.
 pub type Environment = MemoryVicinity;
 
 /// The context of the EVM runtime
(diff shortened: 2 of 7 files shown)


saibatizoku:
So far, evm::backend traits are implemented in a first version, and the state is represented by the `AccountTrie` struct. New methods were added to the `Trie<K, V>` type, to be able to access the inner `Hamt` struct using a mutable reference.

Still pending:
* Implement the `evm::executor::StackState<'_>` trait:
  + keep track of substates and their data
  + expose substate information thru the `StackState<'_>` trait methods
* Implement the `evm::Handler` trait so that the `evm::Runtime` can run
* Put the code together to parse input arguments and execute the runtime

mzabaluev:
> to be able to access the inner Hamt struct using a mutable reference.

The API of `Hamt` is meant to produce new instances at a low cost whenever a modification is made, so there should not be a need to get it by mutable reference. I'll check if it is too difficult to reconcile with the backend traits.

chain-evm/Cargo.toml
```diff
@@ -4,8 +4,9 @@ version = "0.1.0"
 edition = "2018"
 
 [dependencies]
+evm = "0.28.0"
 imhamt = { path = "../imhamt" }
-primitive-types = "0.10.1"
+primitive-types = "0.9.1"
```
mzabaluev, line 9: Why the version dowgrade?

chain-evm/src/machine.rs
```diff
@@ -0,0 +1,211 @@
...
+    backend::{Apply, ApplyBackend, Backend, Basic, Log, MemoryVicinity},
+    executor::StackState,
+    Context,
+};
+use primitive_types::{H160, H256, U256};
+
+use crate::state::AccountTrie;
+
+/// Environment values for the machine backend.
+pub type Environment = MemoryVicinity;
```
mzabaluev, line 22: Good naming, IDK why do the SputnikVM developers use "vicinity".

chain-evm/src/machine.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+    fn reset_balance(&mut self, _address: H160) {
+        todo!();
+    }
+    fn touch(&mut self, _address: H160) {
+        todo!();
+    }
+}
+
+pub fn code_to_execute_evm_runtime() -> Result<(), String> {
```
mzabaluev, line 209: Please add a doc comment explaining this public function.
It could have a shorter name.

chain-evm/src/state/account.rs
```diff
@@ -15,11 +15,51 @@ pub struct Account {
     /// Account data storage.
     pub storage: Storage,
     /// EVM bytecode of this account.
-    pub code: Box<[u8]>,
+    pub code: Vec<u8>,
```
mzabaluev, line 18: `Box` had some benefits: a smaller struct size (which could get significant if the ledger has a lot of value accounts), and a speed bump against any accidental attempts to modify the code in place, which should never be needed. Just like `Vec<u8>`, `Box<[u8]>` provides all slice methods through deref.

chain-evm/src/state/trie.rs
```diff
@@ -47,6 +47,44 @@ impl<K: Clone + Hash + Eq, V: Clone> Trie<K, V> {
             }
         }
     }
+
+    /// Put a value into a trie, replacing an existing value if there was any.
+    /// This method is useful when you have a mutable reference to the Trie.
+    ///
+    /// This method is similar to `Trie::put`, except that it uses `&mut self` to
+    /// modify the inner state.
+    pub fn insert_or_update(&mut self, key: K, value: V) {
```
mzabaluev, line 56: As `Trie` is a thin wrapper over `Hamt`, I think it makes sense to preserve the ownership pattern: the new state should be returned rather than assigned in place.

chain-evm/Cargo.toml
saibatizoku: The reason for the version downgrade is that the evm crate uses "v0.9.1" as types returned in the traits needed to get the EVM runtime going. Specifying a different version in our manifest causes errors when compiling.

chain-evm/src/machine.rs
saibatizoku: Thanks, I got the name from reading the "yellow paper".

chain-evm/src/state/account.rs
saibatizoku: I mainly changed it because the evm crate uses Vec<u8>, and it seemed simpler to avoid having to make Vecs out of Boxes whenever data has to go through the evm code. I'll look into this.

chain-evm/src/state/trie.rs
saibatizoku: Should be resolved in https://github.com/input-output-hk/chain-libs/pull/609/commits/2e2c5103f15d6250238c99a5a9f3fdccfc0d7deb

chain-evm/src/machine.rs
saibatizoku: Yeah, this can be ignored, it was a placeholder that never should've made it to the commit.

saibatizoku:
I added a simple to test that seems to run fine:

A black contract with no data, and default values for everything except the gas limit, which is set to `u64::max_value()`.

chain-evm/src/state/account.rs
saibatizoku: `Box<[u8]>` is restored in the latest commit.

saibatizoku:
Added a few more lines to the test in src/machine.rs

chain-evm/Cargo.toml
saibatizoku: I updated evm to 0.30.0 and primitive-types to 0.10.1

chain-evm/src/machine.rs
```diff
@@ -0,0 +1,248 @@
...
+            state,
+            logs: Default::default(),
+        }
+    }
+    /// Returns an initialized instance of `evm::executor::StackExecutor`.
+    pub fn executor<'backend, 'config>(
+        &'backend self,
+        gas_limit: u64,
+        config: &'config Config,
+    ) -> StackExecutor<'config, MemoryStackState<'backend, 'config, VirtualMachine>> {
```
mzabaluev, line 51: So we expose `StackExecutor` from `evm` in a public method. What is this method used for?

chain-evm/src/machine.rs
```diff
@@ -0,0 +1,248 @@
...
+        StackExecutor::new(memory_stack_state, config)
+    }
+    /// Returns an initialized instance of `evm::Runtime`.
+    pub fn runtime<'config>(
+        &self,
+        code: Rc<Vec<u8>>,
+        data: Rc<Vec<u8>>,
+        context: RuntimeContext,
+        config: &'config Config,
+    ) -> Runtime<'config> {
```
mzabaluev, line 63: Same as above. It would be good to have a test that exercises the abstract API to see if we can really abstract.

saibatizoku: Same as above.

chain-evm/src/machine.rs
saibatizoku: I just added this method to be able to get an executor running, there's no need for it to be public. I've been waiting for feedback on how to proceed.

mzabaluev (approved):
The functions are not used anywhere currently, but at least there are lints to remind us of that.

Next thing is to try to integrate this, after the EVM accounts are added to the -mockchain ledger. I'm on this.

saibatizoku:
> The functions are not used anywhere currently, but at least there are lints to remind us of that.
> 
> Next thing is to try to integrate this, after the EVM accounts are added to the -mockchain ledger. I'm on this.

The only place where the functions are in use are in a unit test on src/machine.rs

https://github.com/input-output-hk/chain-libs/pull/609/files#diff-fc5c4769cbac0a26fbc297b34e18e3ba0489827a98518df38ac95ab5ad8ddfb9R196-R248

PR #612, opened 2021-08-10
Separate SpendingCounter from the spending strategy

vincenthz:
the spending counter increasing strategy is the same as before, but now this distinguish the use of a spending counter, from the usage of the sequentially increasing strategy in the account state

now that the transaction are limited in time, we also allow wrapping of the counter back to zero without the need for the final emptying transaction, whereas before this was a security risk.

the spending counter has a concept of lanes, which distinguish various independently increasing number between 0 and `2^(32-bit_for_lanes)`, each counter becomes self-identifying in term of lanes and allow up to `2^bit_for_lanes` independent transaction in flight without the risk of not having transaction failing sequentially if a previous one is failing.


chain-impl-mockchain/src/accounting/account/spending.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// number of bits reserved for lanes
+const LANES_BITS: usize = 3;
+// number of bits reserved for counter (unrespective of the lane)
+const UNLANES_BITS: usize = 32 - LANES_BITS;
+
+impl SpendingCounterIncreasing {
+    /// number of parallel lanes of increasing counters
+    pub const LANES: usize = 1 << LANES_BITS;
```
danielSanchezQ, line 17: Not sure if useful in this case, but now that we have `const generics` it could be implemented with them?
```rust
impl<const LANES_BITS: usize> SpendingCounterIncreasing  {
    pub const LANES: usize = 1 << LANES_BITS;
    ...
}
```

chain-impl-mockchain/src/accounting/account/spending.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    fn unlaned_counter(self) -> u32 {
+        self.0 & Self::UNLANED_MASK
+    }
+
+    pub(crate) fn new(lane: usize, counter: u32) -> Self {
+        assert!(lane < (1 << LANES_BITS));
+        assert!(counter < (1 << UNLANES_BITS));
+        SpendingCounter((lane << UNLANES_BITS) as u32 | (counter & Self::UNLANED_MASK))
```
danielSanchezQ, line 112: Probably the same here

chain-impl-mockchain/src/accounting/account/spending.rs
vincenthz: it's not terribly useful for anything though, if it's not hardcoded here, it would be very likely in the user, otherwise it will be snowball'ed to every type in the ledger

danielSanchezQ: We could split it in a generic implementation and a type alias for the specific lenght.

```rust
type SpendingCounterIncreasing = GenericSpendingCounterIncreasing<3>;
```
But it is ok anyway.

mzabaluev (approved):

PR #613, opened 2021-08-11
make Ledger::settings return an immutable ref

eugene-babichenko:


enzoc4 (approved):
I was just thinking about this yesterday. `settings_mut` would seem more appropriate if they need to be changed.

eugene-babichenko:
I don't think they are really needed to change. At least, I did not find an example.

PR #616, opened 2021-08-13
Refactor ledger checks to apply to all fragments

Zeegomo:
Use `apply_fragment` in all fragments paths so that we reuse the general checks everywhere


enzoc4 (approved):
LGTM

mzabaluev (approved):
Looks great.

PR #618, opened 2021-08-19
start ApplyBlockLedger with new ledger.date

eugene-babichenko:
This helps to fix a couple inconsistencies in the BlockDate-dependent
fragments processing. Right now this includes TTL checks and voting
periods.

### DIFF
--- a/chain-impl-mockchain/src/ledger/ledger.rs
+++ b/chain-impl-mockchain/src/ledger/ledger.rs
@@ -98,7 +98,6 @@ pub struct ApplyBlockLedger {
     ledger: Ledger,
     ledger_params: LedgerParameters,
     chain_length: ChainLength,
-    block_date: BlockDate,
 }
 
 // Dummy implementation of Debug for Ledger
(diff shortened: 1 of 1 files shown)


chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -1555,37 +1556,28 @@ impl Ledger {
...
 impl ApplyBlockLedger {
-    pub fn block_date(&self) -> BlockDate {
-        self.block_date
-    }
-
     pub fn apply_fragment(&self, fragment: &Fragment) -> Result<Self, Error> {
         let ledger = self
             .ledger
-            .apply_fragment(&self.ledger_params, fragment, self.block_date)?;
+            .apply_fragment(&self.ledger_params, fragment, self.ledger.date)?;
```
Zeegomo, line 1562: Wait, if the fragments were applied using the new `block_date` isn't it equivalent to the new version?

Zeegomo: I mean that I don't think this PR actually changes the behavior of the ledger.
The old `ApplyBlockLedger` didn't update the date of the ledger before applying the transactions, but evaluation for dates is done not on the current ledger date but on the date passed as an argument to `Ledger::apply_fragment` https://github.com/input-output-hk/chain-libs/blob/ecab5ca5cc1c786a2e855b3f2550c87853513b03/chain-impl-mockchain/src/ledger/ledger.rs#L826

As you can see from this diff, it only moves the assignment of the ledger date early and then use the that value instead of saving the new block date, which may be preferred stylistically but I don't think has any functional impact

PR #619, opened 2021-08-19
TTL Max Expiry Epochs Test

mrzenioszeniou:
Ensures that transactions are rejected, if their expiration block date is more than `transaction-maximum-expiry-epochs` epochs away.

Sorted out some clippy warnings too.


mzabaluev (approved):
Looks good

PR #620, opened 2021-08-20
Initial integration of EVM ledger

mzabaluev:


chain-impl-mockchain/src/testing/builders/tx_builder.rs
```diff
@@ -128,8 +128,8 @@ impl TestTxBuilder {
         )];
         let tx_builder = TxBuilder::new()
             .set_payload(&NoExtra)
-            .set_expiry_date(validity.unwrap_or(BlockDate::first()))
-            .set_ios(&inputs, &destination);
+            .set_expiry_date(validity.unwrap_or_else(BlockDate::first))
```
Zeegomo, line 131: Not that it's related to this PR, but I found interesting that the default option is only valid if you are in fact building block 0

Zeegomo (approved):
This looks good but I wonder how this approach will play out for the rest of the functionalities. If I understand correctly the approach is to extend the ledger to always execute EVM code where appropriate, and the evm-responsible ledger will then only contains meaningful code if the evm feature is enabled. This certainly has its own advantages, namely not having to deal with feature gated branches all over then main ledger code, which will hurt readability, and confine this logic in the evm specific code. 
OTOH, maybe there are better ways to make certain methods a no-op than rewrite the entire struct (as the 2 structs are not tied together) but I've not found anything clearly better yet, and for this simple case I think this is ok.

saibatizoku (approved):
Looks good as initial code to build upon, still need to deal with configuration and environment settings needed to get the EVM runtime going ( #625 ), and these will need to go into other parts of the chain-impl-mockchain library, so we should be able to see what the best strategy for feature gating the EVM-related code might be for us.

We should this merge code.

mzabaluev:
> This looks good but I wonder how this approach will play out for the rest of the functionalities. If I understand correctly the approach is to extend the ledger to always execute EVM code where appropriate, and the evm-responsible ledger will then only contains meaningful code if the evm feature is enabled. This certainly has its own advantages, namely not having to deal with feature gated branches all over then main ledger code, which will hurt readability, and confine this logic in the evm specific code.

I think nicer ways are possible, but needs creation of dynamic extension points and boxing. I'll give more thought to making ledger functionality composable.

> OTOH, maybe there are better ways to make certain methods a no-op than rewrite the entire struct (as the 2 structs are not tied together) but I've not found anything clearly better yet, and for this simple case I think this is ok.

It was an alternative to putting lots of conditional code in the main ledger module. I haven't come up with a better approach yet, but as mentioned above there should be better ways, even if with some runtime cost.

PR #621, opened 2021-08-20
EVM account trie in chain-libs

saibatizoku:
Define data structures to represent state of EVM accounts, in a new chain-libs crate chain-evm (coordinate with EAS-68) to support EVM smart contract features. The accounts are identified by 160-bit addresses in the EVM format.

It could be a specialization of account::Ledger with EVM-specific additions, modeled by the MemoryAccount struct in the evm library.


saibatizoku:
This dupliclates https://github.com/input-output-hk/chain-libs/pull/620. I'm closing it.

PR #622, opened 2021-08-24
Encode fragment size with u32

Zeegomo:
Some large fragments (private `VoteTally` with lots of proposals/shares) can overflow the current fragment size (`u16`) when serializing. This PR encodes the size of a `Fragment` as a `u32` to overcome such limitations.

In addition, add some tests that could help detect such issues in the future:
* impl `Arbitrary` for private `VoteTally`, so that we can test its ser/de implementations and stress the behavior of other parts with very large payloads.
* Add a ser/de test for `Fragment`


chain-impl-mockchain/src/block/mod.rs
```diff
@@ -145,15 +145,15 @@ impl Readable for Block {
...
+            let message_size = buf.get_u32()?;
             let mut message_buf = buf.split_to(message_size as usize)?;
 
             // return error here if message serialize sized is bigger than remaining size
 
             let message = Fragment::read(&mut message_buf)?;
             contents.push(message);
 
-            remaining_content_size -= 2 + message_size as u32;
+            remaining_content_size -= 4 + message_size as u32;
```
danielSanchezQ, line 156: Maybe we could add a `const` value for this `4` (`2`), just to name it.

chain-impl-mockchain/src/certificate/test.rs
```diff
@@ -226,7 +231,35 @@ impl Arbitrary for VoteCast {
...
+                    buffer.extend(&Scalar::from_u64(u64::arbitrary(g)).to_bytes());
+                    buffer.extend(&Scalar::from_u64(u64::arbitrary(g)).to_bytes());
+                }
+                inner.push(DecryptedPrivateTallyProposal {
+                    tally_result: (0..n_options.get())
+                        .map(|_| u64::arbitrary(g))
+                        .collect::<Box<[_]>>(),
+                    decrypt_shares: Box::new([TallyDecryptShare::from_bytes(&buffer).unwrap()]),
+                });
+            }
```
danielSanchezQ, line 258: Maybe extract this into a function?

danielSanchezQ (approved):
Great,  thanks!

Zeegomo:
I'm not actually too happy with `Block::read` implementation, as it uses more details about how the size of a fragment than what I would be comfortable with. 
I'm not exactly sure of what the `Readable` trait tries to represent, but I think that implementing `Readable` for `FragmentRaw` and following the same structure as `Block::deserialize` would be the best option here (unless there's a reason why `Readable` was not implemented for `FragmentRaw`)

chain-impl-mockchain/src/certificate/test.rs
```diff
@@ -223,10 +226,53 @@ impl Arbitrary for VoteCast {
     }
 }
 
+fn arbitrary_decrypted_private_tally<G: Gen>(g: &mut G) -> DecryptedPrivateTally {
+    let proposals_n = u8::arbitrary(g);
+    let mut inner = Vec::new();
+    let mut rng = ChaChaRng::seed_from_u64(u64::arbitrary(g));
+    let crs_seed = String::arbitrary(g).into_bytes();
+    let committee_size = 1; // very time consuming
```
mzabaluev, line 234: Maybe try `1..=2` just to exercise the multi-member case?

mzabaluev (approved):

mzabaluev:
> I'm not actually too happy with `Block::read` implementation, as it uses more details about how the size of a fragment than what I would be comfortable with.

IIRC there was an idea to be able to iterate through raw fragments in a block without copying their bodies out of a raw block, and the read implementation for `Block` may reflect this.

eugene-babichenko (approved):

Zeegomo:
> I'm not actually too happy with Block::read implementation, as it uses more details about how the size of a fragment than what I would be comfortable with.

This was related to a temporary implementation that I coded before changing it in https://github.com/input-output-hk/chain-libs/pull/622/commits/03e27ba8dc5e74944f2423195acbebb61193bce2. I'm reasonably happy with the current one

Zeegomo:
I've moved the `Arbitrary` impl for `VoteTally` which was causing problems to https://github.com/input-output-hk/chain-libs/pull/626 and retained here only the encoding fix and a general test for fragments ser/de

PR #624, opened 2021-08-24
Fix CI only testing alternative backend on ubuntu

Zeegomo:
The 'include' field has a tricky behavior. It creates a new job
only if the tags do not match any point in the existing matrix,
and modifies that particular config otherwise. This caused our
current configuration to only test the alternative backend on
ubuntu.

To solve this, add a (empty) 'features' field to the original
matrix.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -72,6 +72,7 @@ jobs:
         toolchain: [stable]
         mode: ['', '--release']
         experimental: [false]
+        features: ['']
         include:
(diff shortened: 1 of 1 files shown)


mzabaluev:
Using a separate matrix job to test outlier features is not the best approach I think.
Coincidentally, I have reworked this to run feature-gated build/test steps as part of each test matrix job in #620.

mzabaluev:
An alternative solution has been merged in #620.

PR #625, opened 2021-08-24
EVM ledger transactions

saibatizoku:


chain-impl-mockchain/src/fragment/mod.rs
```diff
@@ -111,6 +114,7 @@ impl Fragment {
             Fragment::VoteCast(_) => FragmentTag::VoteCast,
             Fragment::VoteTally(_) => FragmentTag::VoteTally,
             Fragment::EncryptedVoteTally(_) => FragmentTag::EncryptedVoteTally,
+            Fragment::SmartContract(_) => FragmentTag::SmartContract,
```
eugene-babichenko, line 117: I think it would be more correct to call it `SmartContractDeploy` or something similar.

mzabaluev: And maybe have an enum to discriminate smart contract types, with EVM as the only currently defined variant.

saibatizoku: I changed the variant and the type as suggested, and updated the code.

chain-evm/src/machine.rs
```diff
@@ -20,6 +20,12 @@ use primitive_types::{H160, H256, U256};
 
 use crate::state::AccountTrie;
 
+/// EVM Configuration.
+pub type Configuration = Config;
```
mzabaluev, line 24: The renaming seems to have no purpose and will make the code less greppable. Leave it at `Config`.

chain-impl-mockchain/src/smartcontract/mod.rs
```diff
@@ -0,0 +1,47 @@
+//! EVM Smart Contract transactions
+
+use chain_core::mempack::Readable;
+#[cfg(feature = "evm")]
+use chain_evm::{Configuration, Environment, GasLimit};
+
+use crate::transaction::Payload;
+
+pub enum Contract {
+    EVM {
```
mzabaluev, line 10: I think the whole variant should be put behind the feature gate, if it's practical.

mzabaluev (commented):
I like what I see so far;
the `todo!()` sites need to be completed, of course.

chain-evm/src/machine.rs
saibatizoku: Left it as `Config`

saibatizoku:
Looking at the JSON-RPC API, the method of interest here is **eth_sendTransaction** which has the following fields:


    from: DATA, 20 Bytes - The address the transaction is send from.
    to: DATA, 20 Bytes - (optional when creating new contract) The address the transaction is directed to.
    gas: QUANTITY - (optional, default: To-Be-Determined) Integer of the gas provided for the transaction execution. It will return unused gas.
    gasPrice: QUANTITY - (optional, default: To-Be-Determined) Integer of the gasPrice used for each payed gas
    value: QUANTITY - (optional) Integer of the value send with this transaction
    data: DATA - (optional) The compiled code of a contract

This is what I want to incorporate next.

saibatizoku:
Added an implementation of the `Readable` trait for the `Contract` enum. It only works if the `evm` feature is enabled.

chain-evm/src/machine.rs
```diff
@@ -14,12 +14,27 @@ use std::rc::Rc;
...
     executor::{MemoryStackState, StackExecutor, StackSubstateMetadata},
-    Config, Context, Runtime,
+    Config as EvmConfig, Context, Runtime,
 };
 use primitive_types::{H160, H256, U256};
 
 use crate::state::AccountTrie;
 
+/// EVM Configuration.
+pub type Config = EvmConfig;
```
mzabaluev, line 24: No need for naming gymnastics:
```suggestion
pub type Config = evm::Config;
```

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -497,6 +497,7 @@ impl Ledger {
                 Fragment::EncryptedVoteTally(_) => {
                     return Err(Error::Block0(Block0Error::HasVoteTally));
                 }
+                Fragment::SmartContractDeploy(_) => todo!(),
```
mzabaluev, line 500: So, this is the next frontier. What we want is to initialize the ledger state with the smart contract data contained in the fragment.

mzabaluev (changes requested):

chain-impl-mockchain/src/ledger/ledger.rs
saibatizoku: Cool. In order to get this working, I'm adding EVM params to the block's Config types.

mzabaluev:
> there seems to be an issue when running the `ledger::recovery::test::ledger_serialize_deserialize_bijection` test.

This usually means there is a bug in serialization/deserialization implementation, or maybe that the `Arbitrary` impl triggers some unimplemented code.

saibatizoku:
> > there seems to be an issue when running the `ledger::recovery::test::ledger_serialize_deserialize_bijection` test.
> 
> This usually means there is a bug in serialization/deserialization implementation, or maybe that the `Arbitrary` impl triggers some unimplemented code.

Yeah, I found that by increasing the maximum `TagLen`, the serialization/deserialization of the ledger works, and that the tests pass with this modification.

chain-impl-mockchain/src/testing/ledger.rs
```diff
@@ -100,6 +109,26 @@ impl ConfigBuilder {
...
+                    block_hashes: BlockHashes::new(),
+                    block_number: BlockNumber::zero(),
+                    block_coinbase: BlockCoinBase::zero(),
+                    block_timestamp: BlockTimestamp::zero() + 1,
+                    block_difficulty: BlockDifficulty::from(131_072),
+                    // FIXME: need to set a real limit
+                    block_gas_limit: GasLimit::max_value(),
+                    // FIXME: adequate base fee
+                    block_base_fee_per_gas: 1u32.into(),
+                },
```
eugene-babichenko, line 130: Hm, this one is a bit confusing. We are going to use this entire structure for blockchain configuration if I understood correctly. But some of these parameters can be derived from Jormungandr. Or maybe even should to avoid a strange/meaningless initial state. This effectively means we need two separate structures: one to configure the blockchain and the other to represent the environment state.

chain-impl-mockchain/src/fragment/mod.rs
```diff
@@ -36,6 +36,7 @@ pub enum Fragment {
     VoteCast(Transaction<certificate::VoteCast>),
     VoteTally(Transaction<certificate::VoteTally>),
     EncryptedVoteTally(Transaction<certificate::EncryptedVoteTally>),
+    SmartContractDeploy(Transaction<smartcontract::Contract>),
```
eugene-babichenko, line 39: Should be renamed to just `Evm` or something like that maybe.

chain-impl-mockchain/src/testing/ledger.rs
saibatizoku: Yes, the values from this were left as default values pending further definition, so it makes perfect sense to include it in future work.

chain-impl-mockchain/src/fragment/mod.rs
saibatizoku: This is done.

chain-evm/src/precompiles/blake2.rs
```diff
@@ -0,0 +1,280 @@
...
+        }
+
+        let cost = Self::required_gas(input)?;
+        if let Some(target_gas) = target_gas {
+            if cost > target_gas {
+                return Err(ExitError::OutOfGas);
+            }
+        }
+
+        let mut rounds_bytes = [0u8; 4];
```
Mr-Leshiy, line 58: maybe better use our `Codec` for all these reading operations ?
or at least move it into another separate functions ?

chain-evm/src/precompiles/blake2.rs
```diff
@@ -0,0 +1,280 @@
...
+        rounds_bytes.copy_from_slice(&input[0..4]);
+        let rounds = u32::from_be_bytes(rounds_bytes);
+
+        let mut h = [0u64; 8];
+        for (mut x, value) in h.iter_mut().enumerate() {
+            let mut word: [u8; 8] = [0u8; 8];
+            x = x * 8 + 4;
+            word.copy_from_slice(&input[x..(x + 8)]);
```
Mr-Leshiy, line 66: same

chain-evm/src/precompiles/blake2.rs
```diff
@@ -0,0 +1,280 @@
...
+            *value = u64::from_le_bytes(word);
+        }
+
+        let mut m = [0u64; 16];
+        for (mut x, value) in m.iter_mut().enumerate() {
+            let mut word: [u8; 8] = [0u8; 8];
+            x = x * 8 + 68;
+            word.copy_from_slice(&input[x..(x + 8)]);
```
Mr-Leshiy, line 74: same

chain-evm/src/precompiles/blake2.rs
```diff
@@ -0,0 +1,280 @@
...
+            *value = u64::from_le_bytes(word);
+        }
+
+        let mut t: [u64; 2] = [0u64; 2];
+        for (mut x, value) in t.iter_mut().enumerate() {
+            let mut word: [u8; 8] = [0u8; 8];
+            x = x * 8 + 196;
+            word.copy_from_slice(&input[x..(x + 8)]);
```
Mr-Leshiy, line 82: same

chain-evm/src/precompiles/secp256k1.rs
```diff
@@ -0,0 +1,197 @@
...
+/// See: https://etherscan.io/address/0000000000000000000000000000000000000001
+// Quite a few library methods rely on this and that should be changed. This
+// should only be for precompiles.
+pub fn ecrecover(hash: H256, signature: &[u8]) -> Result<Address, ExitError> {
+    assert_eq!(signature.len(), 65);
+
+    internal_impl(hash, signature)
+}
+
+fn internal_impl(hash: H256, signature: &[u8]) -> Result<Address, ExitError> {
```
Mr-Leshiy, line 25:
```suggestion
fn internal_impl(hash: H256, signature: &[u8; 64]) -> Result<Address, ExitError> {
```

chain-evm/src/precompiles/secp256k1.rs
```diff
@@ -0,0 +1,197 @@
...
+mod consts {
+    pub(super) const INPUT_LEN: usize = 128;
+}
+
+/// See: https://ethereum.github.io/yellowpaper/paper.pdf
+/// See: https://docs.soliditylang.org/en/develop/units-and-global-variables.html#mathematical-and-cryptographic-functions
+/// See: https://etherscan.io/address/0000000000000000000000000000000000000001
+// Quite a few library methods rely on this and that should be changed. This
+// should only be for precompiles.
+pub fn ecrecover(hash: H256, signature: &[u8]) -> Result<Address, ExitError> {
```
Mr-Leshiy, line 19:
```suggestion
pub fn ecrecover(hash: H256, signature: &[u8; 64]) -> Result<Address, ExitError> {
```

chain-evm/src/precompiles/secp256k1.rs
```diff
@@ -0,0 +1,197 @@
...
+    assert_eq!(signature.len(), 65);
```
Mr-Leshiy, line 20:
```suggestion
```

chain-evm/Cargo.toml
```diff
@@ -4,10 +4,28 @@ version = "0.1.0"
...
 
 [dependencies]
-evm = "0.30.0"
+evm = "0.33.0"
 imhamt = { path = "../imhamt" }
 primitive-types = "0.10.1"
+base64 = { version = "0.13.0", default-features = false, features = ["alloc"] }
+blake2 = { git = "https://github.com/near/near-blake2.git", rev = "736ff607cc8160af87ffa697c14ebef85050138f", default-features = false }
+borsh = { version = "0.8.2", default-features = false }
+bn = { package = "aurora-bn", git = "https://github.com/aurora-is-near/aurora-bn.git", default-features = false, features = ["std"] }
```
danielSanchezQ, line 13: I would try to avoid this if possible or at least:
* Bind to a version
* Bind to a commit

chain-evm/Cargo.toml
```diff
@@ -4,10 +4,28 @@ version = "0.1.0"
...
+libsecp256k1 = { version = "0.3.5", default-features = false }
+num = { version = "0.4.0", default-features = false, features = ["alloc"] }
+ripemd160 = { version = "0.9.1", default-features = false }
+rlp = { version = "0.5.0", default-features = false }
+sha2 = { version = "0.9.3", default-features = false }
+sha3 = { version = "0.9.1", default-features = false }
+wee_alloc = { version = "0.4.5", default-features = false }
+logos = { version = "0.12", default-features = false, features = ["export_derive"] }
+ethabi = { git = "https://github.com/darwinia-network/ethabi", branch = "xavier-no-std", default-features = false }
```
danielSanchezQ, line 22: Likewise here, although it is bind to a branch already.

chain-evm/src/precompiles/blake2.rs
Zeegomo: Mmm, not sure that applies here. This is the format EVM code expects to interact with a precompile function

chain-impl-mockchain/src/config.rs
```diff
@@ -101,6 +109,61 @@ pub enum RewardParams {
     },
 }
 
+#[cfg(feature = "evm")]
+#[derive(Clone, Debug, PartialEq)]
+/// EVM Configuration parameters needed for execution.
+pub struct EvmConfigParams {
+    /// EVM Block Configuration. It is boxed to reduce
```
Zeegomo, line 116: I think the comment does not apply anymore

chain-evm/src/precompiles/mod.rs
```diff
@@ -0,0 +1,341 @@
+//! EVM precompiles based on the aurora-engine project.
```
Zeegomo, line 1: Is all the code in `chain-evm/src/precompiles` (included this file) taken from the aurora-engine project?

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -1,15 +1,106 @@
...
     #[cfg(feature = "evm")]
     pub(crate) accounts: AccountTrie,
+    #[cfg(feature = "evm")]
+    pub(crate) logs: Vec<Log>,
 }
 
 impl Ledger {
     pub fn new() -> Self {
-        Default::default()
+        Self {
```
Zeegomo, line 21: nit: Isn't this equivalent but longer?

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -1,15 +1,106 @@
...
+                access_list,
+            } => {
+                if let Some((new_state, logs, _byte_code_msg)) =
+                    vm.transact_call(caller, address, value, data, gas_limit, access_list, true)
+                {
+                    // update ledger state
+                    self.accounts = new_state.clone();
+                    self.logs.extend_from_slice(logs);
+                }
+                Ok(())
```
Zeegomo, line 92: Does this return an error originated from deploying the contract?

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -1,15 +1,106 @@
...
+                    access_list,
+                    true,
+                ) {
+                    // update ledger state
+                    self.accounts = new_state.clone();
+                    self.logs.extend_from_slice(logs);
+                }
+                Ok(())
+            }
+            Contract::Call {
```
Zeegomo, line 77: nit: don't know if this is standard in the evm ecosystem (and forget me if it is) but it sounds a bit off for me that a call to a contract is in a function called `deploy`

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -1023,6 +1037,22 @@ impl Ledger {
...
+                    let contract = tx.as_slice().payload().into_payload();
+                    let config = &new_ledger.settings.evm_params.config.into();
+                    let environment = &new_ledger.settings.evm_params.environment;
+                    new_ledger
+                        .evm
+                        .deploy_contract(contract, config, environment)?;
+                }
+                #[cfg(not(feature = "evm"))]
+                {
+                    let _ = tx;
```
Zeegomo, line 1053: I think we should not allow any EVM transaction in this case or include some fees anyway otherwise it would be possible to bloat up the storage without having to pay anything

chain-impl-mockchain/src/smartcontract/mod.rs
```diff
@@ -0,0 +1,239 @@
+//! EVM Smart Contract transactions
+
+use chain_core::mempack::{ReadError, Readable};
+#[cfg(feature = "evm")]
```
Zeegomo, line 4: This is general and not only for this particular import but I wonder if we should instead feature-gate e.g. the whole `smartcontract` module instead of the specific logic inside each function

chain-evm/src/precompiles/blake2.rs
saibatizoku: Good calls, thanks. This code was brought in as-is from the Aurora Engine (`CC0` licensed), with the intention to refactor it later. I think that these comments should be moved into a new issue @eugene-babichenko .

chain-evm/src/machine.rs
```diff
@@ -12,64 +12,245 @@
...
+                    let state = executor.into_state();
+                    // Next, we consume the stack state and extract the values and logs
+                    // used to modify the accounts trie in the VirtualMachine.
+                    let (values, logs) = state.deconstruct();
+
+                    self.apply(values, logs, delete_empty);
+                    //_exit_reason
+                    Some((&self.state, &self.logs))
+                }
+                _ => None,
```
Zeegomo, line 158: Could it be useful for debugging to return the error here?

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -1,15 +1,106 @@
...
+    machine::{Config, Environment, Log, VirtualMachine},
+    state::{AccountTrie, Balance},
+};
 
 #[derive(Default, Clone, PartialEq, Eq)]
 pub struct Ledger {
     #[cfg(feature = "evm")]
     pub(crate) accounts: AccountTrie,
+    #[cfg(feature = "evm")]
+    pub(crate) logs: Vec<Log>,
```
Zeegomo, line 16: This is not very fork friendly, I think we should use some persistent data structure (even a dummy HAMT with indexes as keys).

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -513,6 +513,20 @@ impl Ledger {
                     check::valid_block0_cert_transaction(&tx)?;
                     ledger = ledger.mint_token_unchecked(tx.payload().into_payload())?;
                 }
+                Fragment::Evm(tx) => {
+                    #[cfg(feature = "evm")]
+                    {
+                        // WIP: deploying contract
+                        let contract = tx.as_slice().payload().into_payload();
```
Zeegomo, line 520: We should check this is a valid Jormungandr transaction anyway (i.e. ios are balanced, limits are respected, etc), even if we set the fees to zero

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -1,15 +1,106 @@
...
+        match contract {
+            Contract::Create {
+                caller,
+                value,
+                init_code,
+                gas_limit,
+                access_list,
+            } => {
+                //
+                if let Some((new_state, logs)) =
```
Zeegomo, line 45: where do we check the transaction was correctly signed by the caller? or is this left for future revisions?

chain-impl-mockchain/src/ledger/ledger.rs
eugene-babichenko: The plan here is that this is going to be not a Jormungandr transaction, but something living in its own EVM world. Not even a certificate.

chain-evm/src/machine.rs
```diff
@@ -12,64 +12,245 @@
...
         context: RuntimeContext,
-        config: &'config Config,
-    ) -> Runtime<'config> {
-        Runtime::new(code, data, context, config)
+    ) -> Runtime<'_> {
+        Runtime::new(code, data, context, self.config)
+    }
+
+    /// Execute a CREATE transaction
+    #[allow(clippy::boxed_local)]
```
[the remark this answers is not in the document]
Zeegomo: `ByteCode` is an alias to `Box<u8>`, not sure why it's flagged though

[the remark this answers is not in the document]
saibatizoku: This clippy lint came up with the latest version of Rust. It has to do with `ByteCode` being an alias of `Box<[u8]>`, like @Zeegomo  mentions.

chain-evm/src/machine.rs
```diff
@@ -12,64 +12,245 @@
...
+                    //_exit_reason
+                    Some((&self.state, &self.logs))
+                }
+                _ => None,
+            }
+        }
+    }
+
+    /// Execute a CREATE2 transaction
+    #[allow(clippy::boxed_local)]
```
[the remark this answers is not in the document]
saibatizoku: Same as above.

chain-impl-mockchain/src/ledger/evm.rs
saibatizoku: This has not been defined yet.

chain-evm/src/machine.rs
saibatizoku: Yes, the other variants for `ExitReason` need to be properly handled. It should be straightforward to convert into ledger errors or logs. The reason that this method returns an `Option<_>` is to keep things consistent with how the EVM crate works.

PR #626, opened 2021-08-25
impl Arbitrary for private VoteTally

Zeegomo:
Add an `Arbitrary` implementation for a private `VoteTally`, which, apart from adding tests for a transaction that was not covered before, also allows us to stress very big (i.e. > 64Kib) payloads.
We may even think of switching to proptest, but it would mean moving out all of the current tests, which is likely to take quite some time.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -267,6 +267,7 @@ jobs:
         uses: actions-rs/tarpaulin@v0.1
         with:
           out-type: Html
+          timeout: 600
 
       - name: Set branch name
         if: ${{ github.event_name == 'pull_request' }}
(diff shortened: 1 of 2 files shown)


mzabaluev (approved):

mzabaluev:
Looks good, but the tests detect a format discrepancy: "Error: deserialization: Structure invalid: transaction payload is invalid"

Zeegomo:
> Looks good, but the tests detect a format discrepancy: "Error: deserialization: Structure invalid: transaction payload is invalid"
That's the error we were having in production, this branch does not contain the fix for it

Mr-Leshiy (approved):

PR #627, opened 2021-08-31
impl Bech32 for vote keys

Zeegomo:
This way we can have nice HRP without tricks in downstream users like jormungandr

### DIFF
--- a/chain-vote/src/committee.rs
+++ b/chain-vote/src/committee.rs
@@ -80,7 +80,8 @@ impl Bech32 for ElectionPublicKey {
 
     fn try_from_bech32_str(bech32_str: &str) -> Result<Self, Error> {
         try_from_bech32_to_bytes::<Self>(bech32_str).and_then(|raw| {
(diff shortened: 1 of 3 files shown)


chain-vote/src/committee.rs
```diff
@@ -54,6 +56,38 @@ impl ElectionPublicKey {
         );
         (ciphertexts, proof)
     }
+
+    /// Create an election public key from all the participants of this committee
+    pub fn from_participants(pks: &[MemberPublicKey]) -> Self {
```
enzoc4, line 61: such a weird diff, this didn't change, right?

enzoc4 (approved):
LGTM

chain-vote/src/committee.rs
Zeegomo: No, I just grouped together all impl blocks for `ElectionPublicKey`, don't know why we had 2 in the same file

mzabaluev (approved):

PR #630, opened 2021-09-03
Update bech32 dependency to 0.8, fix potential encoding ambiguity

mzabaluev:
Record concerns about bech32m as comments.
The Bech32 weakness is unlikely to affect our use of it for keys and
addresses, all of which expect fixed data length.
The changes for the general-purpose decoding procedure make it
~accept both variants, but~ produce the original Bech32 from BIP-0173.
The fixed binary data length is reinforced through the `Bech32` trait and the functions of the `bech32` module, using a new associated const `BYTES_LEN`.
Address encoding and decoding is fixed to the original Bech32 to avoid
any ambiguity and preserve backward compatibility.

### DIFF
--- a/chain-crypto/src/bech32.rs
+++ b/chain-crypto/src/bech32.rs
@@ -16,12 +16,13 @@ pub trait Bech32 {
 }
 
 pub fn to_bech32_from_bytes<B: Bech32>(bytes: &[u8]) -> String {
-    bech32::encode(B::BECH32_HRP, bytes.to_base32())
+    // FIXME: should switch to Bech32m?
+    bech32::encode(B::BECH32_HRP, bytes.to_base32(), bech32::Variant::Bech32)
         .unwrap_or_else(|e| panic!("Failed to build bech32: {}", e))
 }
 
 pub fn try_from_bech32_to_bytes<B: Bech32>(bech32_str: &str) -> Result<Vec<u8>> {
-    let (hrp, data) = bech32::decode(bech32_str)?;
+    let (hrp, data, _variant) = bech32::decode(bech32_str)?;
     if hrp != B::BECH32_HRP {
         return Err(Error::HrpInvalid {
             expected: B::BECH32_HRP,
--- a/chain-crypto/src/bech32.rs
+++ b/chain-crypto/src/bech32.rs
@@ -5,18 +5,34 @@ use std::result::Result as StdResult;
 
 pub type Result<T> = StdResult<T, Error>;
 
+/// Bech32 encoding for fixed-size binary objects.
 pub trait Bech32 {
+    /// The human-readable prefix that is used to represent the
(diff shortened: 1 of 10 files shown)


chain-crypto/src/bech32.rs
```diff
@@ -16,12 +16,13 @@ pub trait Bech32 {
...
 pub fn to_bech32_from_bytes<B: Bech32>(bytes: &[u8]) -> String {
-    bech32::encode(B::BECH32_HRP, bytes.to_base32())
+    // FIXME: should switch to Bech32m?
+    bech32::encode(B::BECH32_HRP, bytes.to_base32(), bech32::Variant::Bech32)
         .unwrap_or_else(|e| panic!("Failed to build bech32: {}", e))
 }
 
 pub fn try_from_bech32_to_bytes<B: Bech32>(bech32_str: &str) -> Result<Vec<u8>> {
-    let (hrp, data) = bech32::decode(bech32_str)?;
+    let (hrp, data, _variant) = bech32::decode(bech32_str)?;
```
mzabaluev, line 25: I'm not sure this is the right thing to do: a key could be encoded in the Bech32m representation, but some other software may treat it as different if they compare it as a string to the same key encoded in Bech32. Here I allow the ambiguity.

Zeegomo: That was my concern as well (having to check manually the version each time is a bit fragile) but you can't misinterpret a Bech32 str as a Bech32m one, so I would say this is not a problem in deserialization.
If we want to guarantee equal keys -> equal strs we should do that while serializing.

OTOH these function work on variable length binary data, which is exactly what Bech32m improves on and, in general, I would say a bech32 malformed / checksum error is more informative than "invalid binary data" in case of insertion/deletion of characters from the string representation even for fixed size data.

For these functions we could add the variant as an input parameter so that:
* you can use Bech32m if you are serializing/deserializing variable length data or don't need backward compatibility
* use Bech32 if you are interested in backward compatibility and you are serializing fixed length data

Btw, what would we need to update in case we wanted to use Bech32m everywhere?

mzabaluev: > what would we need to update in case we wanted to use Bech32m everywhere?

Any keys and other serialized objects that can now be reused between node deployments.
Some tools (in Python or JS) may also use other libraries to generate bech32 formatted keys and addresses and expect them to interop with our deserialization, but I can't claim to have any clear picture on this.

mzabaluev: OK, I think two variants of the functions are called for: the current from/to_bech_32 fns use the classic Bech32 variant, and the decoding fn is added with a `required_length` parameter, to check the expectation for fixed-size objects. If needed, another couple of functions will be added for variable-length objects, that one encoding with Bech32m.

Zeegomo (approved):

PR #634, opened 2021-09-06
imhamt: migrate to proptest from quickcheck

eugene-babichenko:


mzabaluev:
Some of the converted tests are running for too long when instrumented for coverage. Consider tweaking the strategy to produce smaller workloads, or do we have a performance problem there?

imhamt/src/lib.rs
```diff
@@ -420,17 +363,35 @@ mod tests {
         (h, reference)
     }
 
-    #[quickcheck]
-    fn plan_equivalent(xs: Plan<String, u32>) -> bool {
-        let (h, reference) = arbitrary_hamt_and_btree(xs, next_u32, |v| v.wrapping_mul(2));
-        property_btreemap_eq(&reference, &h)
+    fn arbitrary_hamt_and_btree(
+    ) -> impl Strategy<Value = (Hamt<DefaultHasher, String, u32>, BTreeMap<String, u32>)> {
+        any_with::<Vec<PlanOperation>>(size_range(1000..5120).lift())
```
mzabaluev, line 368: I don't see why the size range must start at 1000 here. It should be possible to create a Hamt with 0 operations? On failures, we want reduction to the smallest failing input.

Zeegomo:
Tarpaulin default timeout is 60s per test, which I think is a bit tight

mzabaluev (changes requested):

mzabaluev:
> Tarpaulin default timeout is 60s per test, which I think is a bit tight

Tight, but right: it makes sense to refactor individual test cases to complete in less than a minute, so that the overall test time does not grow unusably long, and a hack-and-test cycle while working with individual tests does not involve excessive waiting times.

Zeegomo:
But tests on tarpaulin run considerably slower than a normal execution and I would bit a bit sad to reduce input size / quality just to be able to record test coverage. Maybe we can instruct the tarpaulin test to run less iterations with `PROPTEST_CASES` or compile tests with release optimizations?

mzabaluev:
> Maybe we can instruct the tarpaulin test to run less iterations with `PROPTEST_CASES`

Good idea. If running with fewer iterations will result in loss of coverage, we'll know from the output.

imhamt/src/lib.rs
eugene-babichenko: Makes sense 👍

eugene-babichenko:
> Maybe we can instruct the tarpaulin test to run less iterations with `PROPTEST_CASES`

Will not be required once we change the default RNG for `proptest`.

imhamt/src/lib.rs
```diff
@@ -281,8 +250,8 @@ mod tests {
         true
     }
 
-    #[quickcheck]
-    fn insert_equivalent(xs: Vec<(String, u32)>) -> bool {
+    #[proptest]
+    fn insert_equivalent(xs: Vec<(String, u32)>) {
```
mzabaluev, line 254: This is one of these places where we may not need the `String` generator; could do with a simpler key type, so that case reduction does not get lost trying to simplify the string as a Unicode christmas tree.

mzabaluev (approved):

eugene-babichenko:
CI failures are unrelated.

PR #635, opened 2021-09-07
chain-time: migrate to proptest and expose proptest APIs

eugene-babichenko:
quickcheck remains in place for the time being as a compatibility
measure. It should be removed once all crates are using proptest.


eugene-babichenko:
Failures are unrelated.

mzabaluev (approved):

PR #637, opened 2021-09-10
CircleCI fixes

mzabaluev:
- Update the build container to rust 1.54.0
- Fix the way the registry index head hash is updated.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Get top commit hash of cargo registry index
+          name: Get head hash of cargo registry index
           command: |
             git ls-remote --heads https://github.com/rust-lang/crates.io-index.git master |
               cut -f 1 | tee .circleci/crates.io-index.head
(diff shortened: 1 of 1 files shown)


Zeegomo (approved):

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   rust-stable:
     docker:
-      - image: cimg/rust:1.51.0
+      - image: cimg/rust:1.54.0
```
danielSanchezQ, line 6: 1.55 is available already

eugene-babichenko (approved):

.circleci/config.yml
```diff
@@ -26,12 +26,15 @@ jobs:
...
+          name: Generate Cargo.lock
           command: |
             cargo generate-lockfile
-            git -C /home/circleci/.cargo/registry/index/github.com-1ecc6299db9ec823 \
-                show-ref -s refs/remotes/origin/master |
-              tee .circleci/crates.io-index.head
+      - run:
+          name: Update head hash of cargo registry index
+          command: |
+            cat /home/circleci/.cargo/registry/index/*/.git/FETCH_HEAD |
```
eugene-babichenko, line 35: Shall we use the same approach in our GH Actions pipeline? No unnecessary accesses to GitHub servers which is nice.

.circleci/config.yml
mzabaluev: I think we can keep CircleCI 1-2 versions behind, just to provide more dissimilarity in testing and notice when the language features we are using are too new.

.circleci/config.yml
mzabaluev: GitHub Actions does not use this, as their cache [can't be saved with an updated key](https://github.com/actions/cache/issues/135). The HTTP request for the head hash still works.

PR #639, opened 2021-09-10
Replace Arc<Box<[T]>> fields with Arc<[T]>

mzabaluev:
Fixes some new lints, removes an unnecessary allocation layer.


eugene-babichenko (approved):
As discussed before, `Arc<[]>` is a fat pointer, but I think that in general case this is negligible compared to the performance impact of having two indirections.

PR #641, opened 2021-09-15
chain-crypto: migrate tests from quickcheck to proptest

eugene-babichenko:


chain-crypto/src/algorithms/ed25519_extended.rs
```diff
@@ -75,22 +75,26 @@ mod test {
...
+        prop_assert!(keypair_signing_ko(input))
     }
 
-    #[quickcheck]
-    /// `secret_from_binary` should fail if the provided byte array does not match the public key size
-    fn secret_from_binary_size_check(n: usize) {
+    // `secret_from_binary` should fail if the provided byte array does not match the public key size
+    // the size is limited to 1 MiB to avoid segfaults during testing
+    #[proptest]
+    fn secret_from_binary_size_check(#[strategy(..1_048_576usize)] n: usize) {
```
danielSanchezQ, line 94: Can we take the `1_048_576usize` into some const? It is used somewhere else too.

chain-crypto/src/algorithms/sumed25519/mod.rs
```diff
@@ -120,23 +120,27 @@ impl KeyEvolvingSignatureAlgorithm for SumEd25519_12 {
...
 
-    #[quickcheck]
-    /// `public_from_binary`should fail if the provided byte array does not match the public key size
-    fn public_from_binary_size_check(n: usize) {
+    use proptest::prelude::*;
+    use test_strategy::proptest;
+    // `public_from_binary`should fail if the provided byte array does not match the public key size
+    // the size is limited to 1 MiB to avoid segfaults during testing
+    #[proptest]
+    fn public_from_binary_size_check(#[strategy(..1_048_576usize)] n: usize) {
```
danielSanchezQ, line 128: Here

chain-crypto/src/algorithms/sumed25519/mod.rs
```diff
@@ -120,23 +120,27 @@ impl KeyEvolvingSignatureAlgorithm for SumEd25519_12 {
...
         );
     }
 
-    #[quickcheck]
-    /// `signature_from_bytes` should fail if the provided byte array does not match the public key size
-    fn signature_from_bytes_size_check(n: usize) {
+    // `signature_from_bytes` should fail if the provided byte array does not match the public key size
+    // the size is limited to 1 MiB to avoid segfaults during testing
+    #[proptest]
+    fn signature_from_bytes_size_check(#[strategy(..1_048_576usize)] n: usize) {
```
danielSanchezQ, line 140: And here too

danielSanchezQ (approved):
I left some nitpicking comments otherwise LGTM

chain-crypto/src/algorithms/sumed25519/mod.rs
danielSanchezQ: Well, if changing lets just change all 😃

chain-crypto/src/algorithms/sumed25519/mod.rs
mzabaluev: It's overkill for this and similar tests, though: most of the times it will test extravagantly large sizes with no difference to the branch coverage at all.
A more effective strategy would be biased, selecting alternatively from larger, smaller, and the correct size.

In fact, this is a waste of proptest, it could be just a few fixed test cases for the correct size, one-off cases, empty slice, and a large size.

mzabaluev (approved):

eugene-babichenko:
I added the happy path, a zero-sized corner-case and reduced proptests to sensible ranges.

mzabaluev (approved):

PR #642, opened 2021-09-15
Fix more dead_code lints

mzabaluev:
Left the lint on the unused signature field remain, as that one needs to be fixed by actually using the signature.


chain-storage/src/permanent_store.rs
```diff
@@ -25,14 +24,12 @@ impl PermanentStore {
...
 
         let root_id = root_id.into();
-        let id_length = root_id.as_ref().len();
 
         Ok(Self {
             blocks,
             chain_length_index,
             block_id_index,
             root_id,
-            id_length,
```
Zeegomo, line 35: I guess this could be used to avoid one in memory access if the database is of equal length elements, but I don't think datapile allows you to use the api in this way, so it's fine to remove

chain-vote/src/committee.rs
```diff
@@ -95,10 +95,7 @@ impl Bech32 for ElectionPublicKey {
 #[derive(Clone)]
 pub struct MemberState {
     sk: MemberSecretKey,
-    owner_index: usize,
     apubs: Vec<GroupElement>,
-    es: Vec<GroupElement>,
-    encrypted: Vec<(HybridCiphertext, HybridCiphertext)>,
```
Zeegomo, line 101: It's fine if we just want to get rid of the warnings, but I think it would make more sense to either remove the code that generates shares altogether or ignore the warning as to signal this is a temporary implementation and the fields would need to be used in the future

mzabaluev: What are these fields going to be used for? Decryption key recovery?

Zeegomo: This is the first part of the DKG protocol. Those are the values you're supposed to share at the end of the first phase, but without the rest of the protocol they are not really too helpful.
However, if we are going to remove them we might as well remove all the code that generates them

mzabaluev: Right, I will override the lint instead.

danielSanchezQ (approved):

eugene-babichenko (approved):

PR #643, opened 2021-09-17
make Block fields private

Zeegomo:


chain-impl-mockchain/src/testing/ledger.rs
```diff
@@ -512,10 +512,10 @@ impl TestLedger {
     }
 
     pub fn apply_block(&mut self, block: Block) -> Result<(), Error> {
-        let header_meta = block.header.get_content_eval_context();
+        let header_meta = block.header().get_content_eval_context();
         self.ledger = self.ledger.clone().apply_block(
             self.ledger.get_ledger_parameters(),
-            &block.contents,
+            &block.contents(),
```
mzabaluev, line 518:
```suggestion
            block.contents(),
```

mzabaluev (commented):
Looks good (besides a lint), this should have been done long ago.

There should be a jormungandr PR to confirm that we can use the more disciplined API.

mzabaluev (approved):

PR #646, opened 2021-09-20
 chain-addr: migrate tests to proptest 

eugene-babichenko:


chain-addr/src/lib.rs
```diff
@@ -29,11 +29,6 @@
 //!     DISCRIMINATION_BIT || SCRIPT_KIND_TYPE (7 bits) || SCRIPT_IDENTIFIER
 //!
 //! Address human format is bech32 encoded
-//!
-
-#[cfg(test)]
-#[macro_use]
-extern crate quickcheck;
```
mzabaluev, line 36: Do we still want quickcheck feature and dependencies after removing any quickcheck macro use?

mzabaluev (approved):
Needs rustfmt, otherwise looks OK.

chain-addr/src/lib.rs
eugene-babichenko: macro use not required. I am keeping the feature for the time being, will remove after we migrate chain-impl-mockchain and dependent repos.

PR #654, opened 2021-10-01
[Test] cover gaps in Leadership tests

dkijania:


chain-impl-mockchain/src/header/builder.rs
```diff
@@ -153,6 +153,14 @@ impl HeaderBftBuilder<HeaderSetConsensusData> {
         sret.set_signature(BftSignature(sig))
     }
 
+    /// Method introduced for negative testing. It does not derive public key from secret, but allow
+    /// not linked public keys
+    pub fn sign_using_unsafe(self, sk: &SecretKey<Ed25519>, pk: PublicKey<Ed25519>) -> HeaderBft {
```
Zeegomo, line 158: Can we make this a test-only interface?

Zeegomo (changes requested):

chain-impl-mockchain/src/header/builder.rs
dkijania: Fixed that, according to our discussion

Zeegomo (approved):

Issue #522, opened 2021-02-08
Validate decrypted private tally on construction


mzabaluev:
It's possible, at least in code, to create a vote tally payload where the vector lengths in `TallyDecryptShare` elements is different between each other, and from the number of options in the result in #520. This can wreak havoc down the road in tallying or serialization. Tighten the type definitions so that the invariant is upheld.

PR #656, opened 2021-10-05
Add TallyDecryptShare elements size validation

Mr-Leshiy:
Adding elements size validation for the TallyDecryptShare, elements size should be equal to the Tally options amount.
Related to this issue: https://github.com/input-output-hk/chain-libs/issues/522.

### DIFF
--- a/chain-vote/src/tally.rs
+++ b/chain-vote/src/tally.rs
@@ -648,6 +648,43 @@ mod tests {
     }
 
     #[test]
+    fn tally_wrong_elements_size() {
+        let mut rng = ChaCha20Rng::from_seed([0u8; 32]);
+
+        let shared_string =
+            b"Example of a shared string. This should be VotePlan.to_id()".to_owned();
+        let h = Crs::from_hash(&shared_string);
+
+        let mc = MemberCommunicationKey::new(&mut rng);
+        let mc = [mc.to_public()];
+
+        let threshold = 1;
+
+        let m = MemberState::new(&mut rng, threshold, &h, &mc, 0);
+
+        let participants = vec![m.public_key()];
+        let ek = ElectionPublicKey::from_participants(&participants);
+
+        println!("encrypting vote");
+
+        let vote_options = 2;
+        let e = get_encrypted_ballot(&mut rng, &ek, &h, Vote::new(vote_options, 0));
+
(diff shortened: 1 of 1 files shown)


dkijania (changes requested):

chain-vote/src/tally.rs
```diff
@@ -647,6 +656,43 @@ mod tests {
...
+
+        let mc = MemberCommunicationKey::new(&mut rng);
+        let mc = [mc.to_public()];
+
+        let threshold = 1;
+
+        let m = MemberState::new(&mut rng, threshold, &h, &mc, 0);
+
+        let participants = vec![m.public_key()];
+        let ek = ElectionPublicKey::from_participants(&participants);
```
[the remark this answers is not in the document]
Mr-Leshiy: But at this case, we will have cyclic package dependency, because `committee_members_manager` defined in the `chain-impl-mockchain` and this package use as a dependency `chain-vote` package

chain-impl-mockchain/src/certificate/vote_tally.rs
```diff
@@ -167,10 +172,24 @@ impl TallyProof {
...
+    ) -> Result<Self, DecryptedPrivateTallyError> {
+        let is_valid = proposals
+            .iter()
+            .all(|proposal| match proposal.decrypt_shares.first() {
+                Some(share) => proposal
+                    .decrypt_shares
+                    .iter()
+                    .all(|el| el.options() == share.options()),
+                None => true,
+            });
```
eugene-babichenko, line 186: I'd rather do:

```rust
let mut proposals = proposals.iter();
let first_proposal = proposals.next();
```

chain-impl-mockchain/src/certificate/vote_tally.rs
```diff
@@ -167,10 +172,24 @@ impl TallyProof {
...
+
+        is_valid
+            .then(|| Self {
+                inner: proposals.into_boxed_slice(),
+            })
+            .ok_or(DecryptedPrivateTallyError {})
```
eugene-babichenko, line 192: This is overuse of combinators, if-else should be just fine :)

eugene-babichenko (approved):

PR #668, opened 2021-10-21
add helper to get fragment serialized size

zeegomo:


Zeegomo:
> not sure if it really needs to be usize but looks good

the raw fragment data part could be up to u32::MAX, which means with fragment overhead we need at least a u64 (not considering custom bytes length). But also it's a in-memory vector, so can't be really more than a usize, which means our code only works on 64bits arch (I don't think we need to support other arch though).
Using a usize was just easier because of the return type of the underlying method

PR #678, opened 2021-10-31
Updadate prop cert

Zeegomo:
The core of this should be ready, apart from some pieces that are yet to be updated

(failures unrelated)


chain-impl-mockchain/src/update.rs
```diff
@@ -1,94 +1,75 @@
...
-use chain_core::property;
-use chain_crypto::{Ed25519, Signature, Verification};
-use std::collections::{HashMap, HashSet};
+use imhamt::Hamt;
+use std::collections::hash_map::DefaultHasher;
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct UpdateState {
-    pub proposals: HashMap<UpdateProposalId, UpdateProposalState>,
+    pub proposals: Hamt<DefaultHasher, UpdateProposalId, UpdateProposalState>,
```
Mr-Leshiy, line 10: Just a question. Why do we move from the std::HashMap to the Hamt as I see it is a custom implementation of the Hash map ? What is the difference between them ?

chain-impl-mockchain/src/certificate/update_proposal.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    pub fn changes(&self) -> &ConfigParams {
+        &self.changes
+    }
+
+    pub fn proposer_id(&self) -> &UpdateProposerId {
+        &self.proposer_id
+    }
+
+    pub fn serialize_in(&self, bb: ByteBuilder<Self>) -> ByteBuilder<Self> {
```
Mr-Leshiy, line 39: Maybe we need to move this function into the `property::Serialize` trait. But it is not related to this PR.

chain-impl-mockchain/src/certificate/update_proposal.rs
```diff
@@ -0,0 +1,114 @@
...
+        // Should be impossible to fail serialization
+        bb.bytes(self.changes.serialize_as_vec().as_ref().unwrap())
+            .bytes(self.proposer_id.as_ref())
+    }
+
+    pub fn serialize(&self) -> ByteArray<Self> {
```
Mr-Leshiy, line 45: Same

Mr-Leshiy:
Generally LGTM

Mr-Leshiy (approved):

chain-impl-mockchain/src/update.rs
enzoc4: haven't read the rest of the PR, but the hamt is immutable (or persistent), which means it can be cloned cheaply. A big chunk of the ledger uses it so it can be cloned once per block without doing a deep clone.

eugene-babichenko (changes requested):
Please add these transaction to `impl Arbitrary for Certificate` and `Fragment`.

Zeegomo:
I think `impl Arbitrary for Fragment` should already include those

eugene-babichenko (approved):

PR #681, opened 2021-11-03
Fix update proposal processing

Mr-Leshiy:


Mr-Leshiy:
Merged it as it is not a huge update, small fix.

PR #693, opened 2021-11-22
basic tokens implementation for voting

eugene-babichenko:


chain-impl-mockchain/src/tokens.rs
```diff
@@ -0,0 +1,202 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct TokenName(Vec<u8>);
+
+/// A minting policy consists of multiple entries defining different
+/// constraints on the minting process. An empty policy means that new tokens
+/// cannot be minted during the chain run.
+///
+/// Minting policies are meant to be ignored in block0 fragments.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MintingPolicy(Vec<MintingPolicyEntry>);
```
danielSanchezQ, line 29: I think it would be nice to actually reflect this as code:
> An empty policy means that new tokens cannot be minted during the chain run.

Just as example idea:

```rust
pub enum MintingPolicy {
    Constrained(Vec<MintingPolicyEntry>)
    Locked
}
```

eugene-babichenko: How do we handle an empty `Vec` in `Constrained`?

Although one solution is to wrap an `enum`, limit its visibility to `tokens` module and make an empty `Constrained` illegal somehow.

eugene-babichenko: I somewhat agree with you, but adding an explicit enum would require the mentioned verification and it brings too much additional complexity IMO.

Mr-Leshiy: I think it should be enough to wrap `Vac< MintingPolicyEntry>` into the `MintingPolicies(Vec< MintingPolicyEntry>)` (as an example) and there make an impossible to create and make an empty struct

eugene-babichenko: Which would lead to basically the same logic as we have now but with extra steps and some spaghetti code. What I like about the current structure is that:

- It is simple.
- It directly maps into its serialized representation.

eugene-babichenko:
Don't know what's with these failures, our CI is misbehaving.

chain-impl-mockchain/src/tokens.rs
eugene-babichenko: Just stop for a second and see how many layers we already invented to replace a single if-statement that did not seem nice.

danielSanchezQ: Well, I agree that the if statement may do ofc. Something like this would too https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=527415f9459e04778a03047bf4006015. 
But if it is too convoluted go ahead with the initial simple implementation.

eugene-babichenko: To me it looks a bit convoluted :)

chain-impl-mockchain/src/tokens.rs
```diff
@@ -0,0 +1,208 @@
+use chain_core::mempack::{ReadBuf, ReadError, Readable};
+use cryptoxide::{blake2b::Blake2b, digest::Digest};
+use thiserror::Error;
+use typed_bytes::ByteBuilder;
+
+use std::convert::TryInto;
+
+pub const POLICY_HASH_SIZE: usize = 28;
+pub const TOKEN_NAME_MAX_BYTES: usize = 32;
```
Zeegomo, line 9: nit: would be nice to have the same convention for names, e.g. both having `BYTES`/`SIZE` as suffix

chain-impl-mockchain/src/tokens.rs
```diff
@@ -0,0 +1,208 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct PolicyHash([u8; POLICY_HASH_SIZE]);
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct TokenName(Vec<u8>);
+
+/// A minting policy consists of multiple entries defining different
+/// constraints on the minting process. An empty policy means that new tokens
+/// cannot be minted during the chain run.
```
Zeegomo, line 25: saying that a minting policy is a constraint for minting and that no minting policies means no minting is a bit of a contradiction, but as long as it's clearly specified I'm ok with that

chain-impl-mockchain/src/tokens.rs
```diff
@@ -0,0 +1,208 @@
...
+        &self.0
+    }
+}
+
+impl Readable for PolicyHash {
+    fn read(buf: &mut ReadBuf) -> Result<Self, ReadError> {
+        let bytes = buf
+            .get_slice(POLICY_HASH_SIZE)?
+            .try_into()
+            .expect("already read 32 bytes");
```
Zeegomo, line 115: what's up with this error?

chain-impl-mockchain/src/tokens.rs
```diff
@@ -0,0 +1,208 @@
...
+pub const TOKEN_NAME_MAX_BYTES: usize = 32;
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub struct TokenIdentifier {
+    pub policy_hash: PolicyHash,
+    pub token_name: TokenName,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct PolicyHash([u8; POLICY_HASH_SIZE]);
```
Zeegomo, line 18: Maybe we could have a struct like `Blake2b224` (we already have `Blake2b256`) so that the hash type is also enforced

chain-impl-mockchain/src/tokens.rs
eugene-babichenko: Here we convert `&[u8]` to `[u8; 32]`. This conversion can fail but the only condition for this would be size mismatch. It should not be possible if our code is functioning correctly, so I put an `.expect()` here to catch some severe misbehaviour.

chain-impl-mockchain/src/tokens.rs
eugene-babichenko: Well, I do not see a particular reason to do this. We need to serialize and deserialize this type anyway. At this point there is no "hash of a particular type", just some bytes that may happen to be a valid hash.

chain-impl-mockchain/src/tokens.rs
Zeegomo: But isn't the policy hash 28 bytes?

eugene-babichenko: Oopsie, should've used `format!()` here maybe.

chain-impl-mockchain/src/certificate/mint_token.rs
```diff
@@ -0,0 +1,101 @@
...
+};
+
+use chain_core::mempack::{ReadBuf, ReadError, Readable};
+use typed_bytes::ByteBuilder;
+
+use std::marker::PhantomData;
+
+#[derive(Debug, Clone)]
+pub struct MintToken {
+    pub token: TokenIdentifier,
```
Zeegomo, line 16: What about making the fields private to enforce consistency between token and minting policy?
Also keeping the hash of the policy in `token` is a bit redundant at this point. Not storing a copy of the policy hash would solve both cases

eugene-babichenko: > Also keeping the hash of the policy in token is a bit redundant at this point.

It is a bit redundant _at this point_ but this will change in near future.

eugene-babichenko: > What about making the fields private to enforce consistency between token and minting policy?

How would it help?

Zeegomo: > It is a bit redundant at this point but this will change in near future.

What do you mean? Isn't the policy hash always derived from hashing data you already keep in the struct?

> What about making the fields private to enforce consistency between token and minting policy?
How would it help?

I  cannot write something like
```
MintToken { token: TokenIdentifier::new("token", TestGen::hash()), policy: somepolicy}
```

eugene-babichenko: > I cannot write something like
>
>     MintToken { token: TokenIdentifier::new("token", TestGen::hash()), policy: somepolicy}

Ah, got it, makes sense.

> What do you mean? Isn't the policy hash always derived from hashing data you already keep in the struct?

Wait, so just make it clear. You mean that in the `MintToken` certificate it doesn't make sense to keep the policy hash because the policy itself is already there? If so, it was a misunderstanding and you're right, the policy hash as a certificate field is redundant.

Zeegomo: > You mean that in the MintToken certificate it doesn't make sense to keep the policy hash because the policy itself is already there? If so, it was a misunderstanding and you're right, the policy hash as a certificate field is redundant.

yep, exactly

Zeegomo (approved):

PR #692, opened 2021-11-17
Update proposal scenario

dkijania:
Expose useful method to be used in Jormungandr project

PR #695, opened 2021-11-22
Port update proposals helpers to catalyst-fund7 (#692)

dkijania:
* add new certs to FragmentFactory

enhance controller with new methods for update proposals

* add new test

### DIFF
--- a/chain-impl-mockchain/src/testing/scenario/fragment_factory.rs
+++ b/chain-impl-mockchain/src/testing/scenario/fragment_factory.rs
@@ -183,18 +183,30 @@ impl FragmentFactory {
         &self,
         valid_until: BlockDate,
         owner: &Wallet,
+        signer: &Wallet,
         update_proposal: UpdateProposal,
     ) -> Fragment {
-        self.transaction_with_cert(valid_until, Some(owner), &update_proposal.into())
+        TestTxCertBuilder::new(self.block0_hash, self.fee).make_transaction_different_signers(
+            valid_until,
+            owner,
+            vec![signer],
+            &update_proposal.into(),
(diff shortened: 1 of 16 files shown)


chain-impl-mockchain/src/testing/scenario/fragment_factory.rs
```diff
@@ -183,18 +183,30 @@ impl FragmentFactory {
...
+            vec![signer],
+            &update_proposal.into(),
+        )
     }
 
     pub fn update_vote(
         &self,
         valid_until: BlockDate,
         owner: &Wallet,
+        signer: &Wallet,
```
Mr-Leshiy, line 201: Why does this needs to separate who will be an `owner` and who will sign this certificate ?
It should be the same person I do not think that we need to separate this entities.

chain-impl-mockchain/src/testing/scenario/fragment_factory.rs
```diff
@@ -183,18 +183,30 @@ impl FragmentFactory {
...
```
Mr-Leshiy, line 186: same

chain-impl-mockchain/src/testing/scenario/fragment_factory.rs
dkijania: Well not necessarily, for catalyst funds we are using bft nodes which have secrets without any funds. We have one account with some funds which will do that kind of transactions. So we need to send transaction using our account with auth for bft leaders. So there is at least one use case for different auth. Maybe the name is misleading....

Mr-Leshiy (approved):

PR #696, opened 2021-11-24
Refactor Serialize/Deserialize traits

Mr-Leshiy:
Remove redundant `Readable` trait with the `ReadBuf` reader struct, we can use `Codec` instead. Update `Serialize` and `Deserialize` interface, update error types.
Also add implementation of the `Serialize` trait for the `MintToken` structures, which is needed for the `jormungandr` update.

This is the first PR related to refactor/improvement of the serialization/deserialization objects. For the future improvements need to unify the serde interface.


Zeegomo:
I was confused by that at first as well, but they're not actually perfectly identical, `Readable` is for in-memory buffers, `Deserialize` for general purpose readers. 
I'm not aware of any type which has significant differences in the implementations for the two traits though and if we cannot justify its existence I'm in favor of dropping it

eugene-babichenko:
> I was confused by that at first as well, but they're not actually perfectly identical, `Readable` is for in-memory buffers, `Deserialize` for general purpose readers.

They are not that different at all. An in-memory buffer (e.g. `&[u8]`) is a reader as well so `Readable` actually looks redundant if you compare the two. As far as I'm concerned, I would only try to avoid unnecessary data copies and I need to compare `ReadBuf` vs `std:io::Read` in that regard.

eugene-babichenko:
So, here is a thing, that I consider to be problematic about this refactoring.

### Old behaviour

- We use `ReadBuf::get_slice(len: usize)` to get a slice of an underlying buffer. This is a zero-copy procedure.
- In all cases we pass that slice to something like `T::from_bytes(bytes: &[u8])` where it gets processed and a copy or processing of some kind.

The process involves **a single copy** to deserialize a single object, which is reasonable.

### New behaviour

- `std::io::Read` (and thus `Codec`) implement reading from a byte slice as copying that byte slice into a `Vec<u8>`.
- The second step is the same as with the old behaviour.

So, we end up having **an extra copy** at step 1 which is not desired.

### Conclusions

Having that said, I would either stick with the old approach or reconsider how functions taking a byte slice operate. The latter might not be feasible in all cases, since there might be some kind of actual parsing involved. For example, `chain-crypto` does a lot of splitting these slices into sub-slices and only then either copying them or parsing into integers.

Mr-Leshiy:
@eugene-babichenko , good point !
I assume we can save an old behavior even with the `Codec`. As we use for deserialization only `std::io::BufRead` trait bound for the reader, so we can add a new function for the `Codec` which should have the same behavior as `ReadBuf::get_slice`

```
fn get_slice(&mut self, n: usize) -> Result<&[u8], ReadError> {
        let slice = self.fill_buf()?;
        self.consume(n);
        Ok(&slice[..n])
}
```

eugene-babichenko:
@Mr-Leshiy that is an interesting approach.

You need to be careful with it though. The documentation `BufRead` says:

> Returns the contents of the internal buffer, filling it with more data from the inner reader if it is empty.

So it may not have `n` bytes you require depending on the nature of the underlying reader.

Mr-Leshiy:
@eugene-babichenko , I have updated codec with the new function and replace the usage of the `get_bytes()` with the new `get_slice()` function, to remove extra copying.

chain-ser/src/packer.rs
```diff
@@ -64,79 +65,89 @@ impl<R: std::io::BufRead> Codec<R> {
...
         let mut buf = vec![0u8; n];
-        self.0.read_exact(&mut buf)?;
+        self.inner.read_exact(&mut buf)?;
         Ok(buf)
     }
     #[inline]
-    pub fn get_slice(&mut self, slice: &mut [u8]) -> Result<(), ReadError> {
-        self.0.read_exact(slice)?;
+    /// This is a wrapper over the std::io::BufRead::fill_buf() function,
+    /// so be aware of that you need also execute consume() function to move the reader position
```
eugene-babichenko, line 73: I think we are starting to expose the internal complexity here. Calling `BufRead::consume(n)` inside of this function would be just fine IMO. The original goal of this function is to return a slice bytes and proceed forward.

chain-ser/src/packer.rs
```diff
@@ -64,79 +65,89 @@ impl<R: std::io::BufRead> Codec<R> {
...
+    pub fn get_slice(&mut self, n: usize) -> Result<&[u8], ReadError> {
+        let data = self.inner.fill_buf()?;
+        if data.len() < n {
+            return Err(ReadError::NotEnoughBytes(data.len(), n));
+        }
```
eugene-babichenko, line 78: We bump into another problem here. Having less than `n` bytes in a `BufRead` doesn't really _mean_ that there are no `n` bytes and we will never have these `n` bytes. Consider reading from a network stream wrapped into `BufReader` for example. And I do not see how we can bypass this issue without introducing some suboptimal approach. For example, we could call `BufRead::fill_buf()` until we get the required amount of bytes, but if the network stream is stalled we end up iterating forever. This is pretty much equivalent to hanging `read_exact` except the latter may trigger some timeout error in such case.

I am really starting to lean towards not removing the old implementation of deserialization from in-memory slices precisely because it accounts for all the specifics of in-memory buffers while not having to deal with the specifics of readers.

Mr-Leshiy:
@eugene-babichenko, I have updated this refactoring. Unfortunately it is not possible to use `fill_buff` as I expected, because of the reason that you have mentioned. So I have added one more `DeserializeFromSlice` trait which is in some way analogous of the previous `Readable` trait, the main difference is the usage `Codec` instead of `ReadBuf `.

chain-ser/src/deser.rs
```diff
@@ -1,28 +1,140 @@
...
 
-impl<T: Serialize> Serialize for &T {
-    type Error = T::Error;
+pub trait DeserializeFromSlice: Sized {
+    fn deserialize_from_slice(codec: &mut Codec<&[u8]>) -> Result<Self, ReadError>;
 
-    fn serialize<W: std::io::Write>(&self, writer: W) -> Result<(), T::Error> {
-        (**self).serialize(writer)
+    fn deserialize_validate_from_slice(codec: &mut Codec<&[u8]>) -> Result<(), ReadError> {
+        Self::deserialize_from_slice(codec).map(|_| ())
```
eugene-babichenko, line 108: Is there any specific need for that? It should be essentially the same as `deserialize_validate` since `&[u8]` is also `Read`.

chain-ser/src/deser.rs
```diff
@@ -1,28 +1,140 @@
...
+        }
+    }
+}
+
+impl From<std::io::Error> for ReadError {
+    fn from(err: std::io::Error) -> Self {
+        ReadError::InvalidData(err.to_string())
+    }
+}
+
```
eugene-babichenko, line 77: Maybe use `thiserror` here? It is used throughout the entire project anyway.

chain-impl-mockchain/src/transaction/transaction.rs
```diff
@@ -261,55 +264,57 @@ pub(super) struct TransactionStruct {
...
+        Witness::deserialize_validate_from_slice(&mut codec)
+            .map_err(|_| TransactionStructError::WitnessesInvalid)?;
     }
 
     // read payload auth
-    let payload_auth_pos = rb.position();
+    let payload_auth_pos = sz - codec.bytes_left();
     if P::HAS_DATA && P::HAS_AUTH {
-        if rb.is_end() {
+        if codec.bytes_left() == 0 {
```
eugene-babichenko, line 310: This value is routinely compared to zero. It looks like a good candidate to be wrapped into a function like `fn has_bytes_left(&self) -> bool`.

eugene-babichenko (commented):
I left several notes, but definitely agree with the idea in general and think we can move forward with this.

chain-ser/src/packer.rs
```diff
@@ -3,233 +3,177 @@
...
 //! data.
 
-const INITIAL_BUFFERED_CAPACITY: usize = 2048;
+use crate::deser::{ReadError, WriteError};
+use std::num::{NonZeroU32, NonZeroU64};
 
-pub struct Codec<I>(I);
+pub struct Codec<I> {
+    inner: I,
+}
```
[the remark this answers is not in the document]
eugene-babichenko: Also, `I` can be limited to `Read` here.

chain-ser/src/deser.rs
eugene-babichenko: Ok I see we are actually using `get_slice` in certain places, that makes sense.

eugene-babichenko:
A couple of thoughts for future reference. Not that we have mostly unified several traits that existed before, it is actually possible to come up with a DSL to describe these serialization format and generate `Serialize`/`Deserialize` implementations from of it. These can be implemented as either a macro language or some declarative builder-style thing. Using either of these two options should allow us to automatically generate `DeserializeFromSlice` (or rather a `&[u8]` specialization of `Deserialize`) when possible.

chain-ser/src/deser.rs
eugene-babichenko: Or can we just implement a specialized variant of `Deserialize`?

eugene-babichenko (approved):

Mr-Leshiy:
@eugene-babichenko , updated versions:
chain-wallet-libs: https://github.com/input-output-hk/chain-wallet-libs/pull/208
jormungandr: https://github.com/input-output-hk/jormungandr/pull/3712

saibatizoku (approved):

PR #698, opened 2021-12-01
Add Bech32 impl for TokenName

Mr-Leshiy:
Adding `Bech32` trait implementation for the `TokenName` struct. It will be convinient on the `jormungandr` side for different purposes

### DIFF
--- a/chain-impl-mockchain/src/tokens/name.rs
+++ b/chain-impl-mockchain/src/tokens/name.rs
@@ -5,6 +5,7 @@ use chain_core::{
     packer::Codec,
     property::Serialize,
 };
+use chain_crypto::bech32::{self, Bech32};
 use thiserror::Error;
 
 pub const TOKEN_NAME_MAX_SIZE: usize = 32;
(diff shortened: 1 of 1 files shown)


chain-impl-mockchain/src/tokens/name.rs
```diff
@@ -60,6 +61,23 @@ impl Readable for TokenName {
     }
 }
 
+impl Bech32 for TokenName {
+    const BECH32_HRP: &'static str = "token-name";
```
Zeegomo, line 65: nit: I think all other hrps use snake_case instead of kebab-case

chain-impl-mockchain/src/tokens/name.rs
```diff
@@ -60,6 +61,23 @@ impl Readable for TokenName {
...
+    const BYTES_LEN: usize = TOKEN_NAME_MAX_SIZE;
```
Zeegomo, line 66: This is not actually the size in bytes of data, but just an upper bound. Since `TokenName` is variable length I think we should use `Bech32m` (see https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki). There was some discussion already here https://github.com/input-output-hk/chain-libs/pull/630

Zeegomo (changes requested):

chain-impl-mockchain/src/tokens/name.rs
Mr-Leshiy: Got it, thanks for this catch.
 But I think an implementation of the `Bech32m` it is not a part of this update and also needs a bigger discussion. I think at this point I will just close this PR, and will use basic hex encoding.

PR #703, opened 2021-12-08
added unit tests for spending module

dkijania:
incorporated new way of dealing with spending counter
added scenario  tests

### DIFF
--- a/chain-impl-mockchain/src/accounting/account/spending.rs
+++ b/chain-impl-mockchain/src/accounting/account/spending.rs
@@ -153,3 +153,114 @@ impl From<SpendingCounter> for u32 {
         v.0
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use quickcheck::TestResult;
+
+    #[quickcheck_macros::quickcheck]
+    fn spending_counter_serialization_bijection(sc: SpendingCounter) -> TestResult {
+        let bytes = sc.to_bytes();
+        TestResult::from_bool(SpendingCounter::from_bytes(bytes) == sc)
+    }
+
+    #[test]
+    #[should_panic]
+    fn new_invalid_spending_counter() {
+        let lane: usize = (1 << LANES_BITS) + 1;
+        let counter: u32 = 1 << UNLANES_BITS;
+        SpendingCounter::new(lane, counter);
+    }
+
+    #[quickcheck_macros::quickcheck]
+    fn new_spending_counter(mut lane: usize, mut counter: u32) {
+        lane = lane % (1 << LANES_BITS);
+        counter = counter % (1 << UNLANES_BITS);
+        let sc = SpendingCounter::new(lane, counter);
+
+        assert_eq!(lane, sc.lane());
+        assert_eq!(counter, sc.unlaned_counter());
+    }
+
+    #[quickcheck_macros::quickcheck]
+    fn increment_counter(mut spending_counter: SpendingCounter) {
+        let lane_before = spending_counter.lane();
+        let counter = spending_counter.unlaned_counter();
+        spending_counter = spending_counter.increment();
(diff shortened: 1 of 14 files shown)


chain-impl-mockchain/src/accounting/account/spending.rs
```diff
@@ -153,3 +153,114 @@ impl From<SpendingCounter> for u32 {
...
+        assert_eq!(counter + 1, spending_counter.unlaned_counter());
+    }
+
+    #[quickcheck_macros::quickcheck]
+    pub fn increment_nth(mut spending_counter: SpendingCounter, n: u32) {
+        let lane_before = spending_counter.lane();
+        let counter = spending_counter.unlaned_counter();
+        spending_counter = spending_counter.increment_nth(n);
+        assert_eq!(lane_before, spending_counter.lane());
+        assert_eq!(counter + n, spending_counter.unlaned_counter());
```
Zeegomo, line 201: same as above, but we should also account for the possibility that `counter + n` does not fit in a u32 and panics in debug mode

chain-impl-mockchain/src/accounting/account/spending.rs
```diff
@@ -153,3 +153,114 @@ impl From<SpendingCounter> for u32 {
...
+        assert_eq!(counter, sc.unlaned_counter());
+    }
+
+    #[quickcheck_macros::quickcheck]
+    fn increment_counter(mut spending_counter: SpendingCounter) {
+        let lane_before = spending_counter.lane();
+        let counter = spending_counter.unlaned_counter();
+        spending_counter = spending_counter.increment();
+        assert_eq!(lane_before, spending_counter.lane());
+        assert_eq!(counter + 1, spending_counter.unlaned_counter());
```
Zeegomo, line 192: this should be a wrapping add, as that's how the spending counter deals with overflows in the counter

chain-impl-mockchain/src/accounting/account/spending.rs
```diff
@@ -153,3 +153,114 @@ impl From<SpendingCounter> for u32 {
...
+    }
+
+    #[test]
+    pub fn spending_counters_incorrect_order() {
+        let counters = vec![SpendingCounter::new(1, 0), SpendingCounter::new(0, 0)];
+        assert!(SpendingCounterIncreasing::new_from_counters(counters).is_none());
+    }
+
+    #[test]
+    pub fn spending_counters_overflow() {
```
Zeegomo, line 227: nit: I don't get the naming here

chain-impl-mockchain/src/accounting/account/spending.rs
```diff
@@ -153,3 +153,114 @@ impl From<SpendingCounter> for u32 {
...
+
+    #[test]
+    #[should_panic]
+    pub fn spending_counter_increasing_wrong_lane() {
+        let mut sc_increasing = correct_spending_counter_increasing();
+        let incorrect_sc = SpendingCounter::new(SpendingCounterIncreasing::LANES, 1);
+        assert!(sc_increasing.next_verify(incorrect_sc).is_err());
+    }
+
+    fn correct_spending_counter_increasing() -> SpendingCounterIncreasing {
```
Zeegomo, line 260: Why not `SpendingCounterIncreasing::default`?

chain-impl-mockchain/src/accounting/account/spending.rs
Zeegomo: I meant that if n is large (e.g. `u32::MAX`), adding its value to the current counter in `spending_counter.increment_nth(n)` will panic in debug mode.

dkijania: I Split test cases then to discard illegal case in quickcheck but added explicit case for overflows

Zeegomo (approved):

PR #705, opened 2021-12-13
Fix early validation issue for update proposal

Mr-Leshiy:
Fix issue for the UpdateProposal. Add validation of the `ConfigParams` during the applying update proposal into the ledger. Also update `process_proposals` function, it is not return an error anymore we assume that all update_proposals should be valid at this point.

### DIFF
--- a/chain-impl-mockchain/src/update.rs
+++ b/chain-impl-mockchain/src/update.rs
@@ -229,6 +229,7 @@ impl From<ActiveSlotsCoeffError> for Error {
 mod tests {
     use super::*;
     use crate::certificate::UpdateProposal;
+    #[cfg(test)]
     use crate::milli::Milli;
     #[cfg(test)]
     use crate::testing::serialization::serialization_bijection;
--- a/chain-impl-mockchain/src/update.rs
+++ b/chain-impl-mockchain/src/update.rs
@@ -26,6 +26,7 @@ impl UpdateState {
     ) -> Result<Self, Error> {
         let proposer_id = proposal.proposer_id();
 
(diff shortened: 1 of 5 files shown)


chain-impl-mockchain/src/update.rs
```diff
@@ -101,7 +103,9 @@ impl UpdateState {
                 // If a majority of BFT leaders voted for the
                 // proposal, then apply it.
                 if proposal_state.votes.size() > settings.bft_leaders.len() / 2 {
-                    settings = settings.apply(proposal_state.proposal.changes())?;
+                    settings = settings
+                        .try_apply(proposal_state.proposal.changes())
+                        .unwrap();
```
Zeegomo, line 108: I would add a panic warning in the function documentation

Zeegomo (approved):

chain-impl-mockchain/src/update.rs
```diff
@@ -101,7 +103,9 @@ impl UpdateState {
                 // If a majority of BFT leaders voted for the
                 // proposal, then apply it.
                 if proposal_state.votes.size() > settings.bft_leaders.len() / 2 {
-                    settings = settings.apply(proposal_state.proposal.changes())?;
+                    settings = settings
```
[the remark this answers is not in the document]
Mr-Leshiy: It is a good point, maybe for now we can leave just a comment for this situation if it is really possible

chain-impl-mockchain/src/update.rs
Mr-Leshiy: updated

eugene-babichenko (approved):

PR #707, opened 2021-12-13
add voting_token field to VotePlan

enzoc4:
unused at the time being, but it should be used to compute the tally for
the VotePlan


chain-impl-mockchain/src/testing/scenario/template/mod.rs
```diff
@@ -235,6 +239,10 @@ impl From<VotePlanDef> for VotePlan {
             proposals,
             dto.payload_type,
             dto.committee_keys,
+            TokenIdentifier {
```
Zeegomo, line 242: probably we should add this field to `VotePlanDef`

Zeegomo (approved):
I actually meant to approve it

chain-impl-mockchain/src/testing/scenario/template/mod.rs
enzoc4: I did this, not sure if your approve still holds.

eugene-babichenko (approved):

PR #712, opened 2021-12-16
Update EVM transaction error handling

Mr-Leshiy:
Return errors for EVM transaction when chain-libs is compiled without "evm" feature

### DIFF
--- a/chain-impl-mockchain/src/ledger/ledger.rs
+++ b/chain-impl-mockchain/src/ledger/ledger.rs
@@ -325,6 +325,8 @@ pub enum Error {
     UpdateVoteSignatureFailed,
     #[error("minting policy violation")]
     MintingPolicyViolation(#[from] MintingPolicyViolation),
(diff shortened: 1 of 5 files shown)


chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -325,6 +325,8 @@ pub enum Error {
     UpdateVoteSignatureFailed,
     #[error("minting policy violation")]
     MintingPolicyViolation(#[from] MintingPolicyViolation),
+    #[error("evm transactions are disabled, build with the 'evm' feature")]
```
eugene-babichenko, line 328:
```suggestion
    #[error("evm transactions are disabled, the node was built without the 'evm' feature")]
```

eugene-babichenko (changes requested):

eugene-babichenko (approved):

PR #713, opened 2021-12-16
Derive voting stake from voting token

enzoc4:
I need to re-check some things, but tests are passing now.

### DIFF
--- a/chain-impl-mockchain/src/accounting/account/mod.rs
+++ b/chain-impl-mockchain/src/accounting/account/mod.rs
@@ -276,7 +276,7 @@ mod tests {
             let voting_tokens_size = usize::arbitrary(gen);
             let arbitrary_voting_tokens = iter::from_fn(|| Some(TokenIdentifier::arbitrary(gen)))
                 .take(voting_tokens_size)
-                .collect::<Vec<_>>();
+                .collect::<HashSet<_>>();
 
             let mut ledger = Ledger::new();
 
--- a/chain-impl-mockchain/src/accounting/account/mod.rs
+++ b/chain-impl-mockchain/src/accounting/account/mod.rs
@@ -208,6 +208,16 @@ impl<ID: Clone + Eq + Hash, Extra: Clone> Ledger<ID, Extra> {
             .map_err(|e| e.into())
     }
 
+    pub fn token_get_total(&self, token: &TokenIdentifier) -> Result<Value, ValueError> {
+        let values = self
+            .0
+            .iter()
+            .filter_map(|(_, account_state)| account_state.tokens.lookup(&token))
+            .cloned();
+
+        Value::sum(values)
+    }
+
     pub fn iter(&self) -> Iter<'_, ID, Extra> {
         Iter(self.0.iter())
     }
@@ -263,6 +273,11 @@ mod tests {
                 .take(stake_pool_size)
                 .collect::<Vec<_>>();
 
+            let voting_tokens_size = usize::arbitrary(gen);
+            let arbitrary_voting_tokens = iter::from_fn(|| Some(TokenIdentifier::arbitrary(gen)))
+                .take(voting_tokens_size)
+                .collect::<Vec<_>>();
+
             let mut ledger = Ledger::new();
 
             // Add all arbitrary accounts
@@ -270,6 +285,15 @@ mod tests {
                 ledger = ledger
                     .add_account(&account_id, AverageValue::arbitrary(gen).into(), ())
                     .unwrap();
+
+                for token in &arbitrary_voting_tokens {
+                    // TODO: maybe less probability is better (for performance)
+                    if bool::arbitrary(gen) {
+                        ledger = ledger
+                            .token_add(&account_id, token.clone(), Value::arbitrary(gen))
+                            .unwrap();
+                    }
+                }
             }
 
             // Choose random subset of arbitraty accounts and delegate stake to random stake pools
--- a/chain-impl-mockchain/src/accounting/account/mod.rs
+++ b/chain-impl-mockchain/src/accounting/account/mod.rs
@@ -212,7 +212,7 @@ impl<ID: Clone + Eq + Hash, Extra: Clone> Ledger<ID, Extra> {
         let values = self
             .0
             .iter()
-            .filter_map(|(_, account_state)| account_state.tokens.lookup(&token))
+            .filter_map(|(_, account_state)| account_state.tokens.lookup(token))
             .cloned();
 
         Value::sum(values)
--- a/chain-impl-mockchain/src/ledger/ledger.rs
+++ b/chain-impl-mockchain/src/ledger/ledger.rs
@@ -307,6 +307,8 @@ pub enum Error {
     VoteTallyProofFailed,
     #[error("Vote tally decryption failed")]
     VoteTallyDecryptionFailed,
+    #[error("Vote Tally vote plan is not in the ledger")]
+    VotePlanNotFound,
     #[error("Pool update payload signature failed")]
     PoolUpdateSignatureFailed,
     #[error("Pool update last known registration hash doesn't match")]
@@ -1182,7 +1184,18 @@ impl Ledger {
             return Err(Error::VoteTallyProofFailed);
         }
 
-        let stake = StakeControl::new_with(&self.accounts, &self.utxos);
+        let voting_token = self
+            .votes
+            .plans
+            .lookup(tally.id())
+            // XXX: unsure about this.  This would be handled anyway inside
+            // `apply_commitee_result`, although with a different error.  But it needs to be
+            // checked here, unless this logic is moved to the VotePlanManager somehow.
+            .ok_or(Error::VotePlanNotFound)?
+            .plan()
+            .voting_token();
+
+        let stake = StakeControl::new_with(voting_token, &self.accounts);
 
         let mut actions = Vec::new();
 
@@ -1226,7 +1239,16 @@ impl Ledger {
             return Err(Error::VoteTallyProofFailed);
         }
 
-        let stake = StakeControl::new_with(&self.accounts, &self.utxos);
+        let voting_token = self
+            .votes
+            .plans
+            .lookup(tally.id())
+            // XXX: idem `apply_vote_tally`
+            .ok_or(Error::VotePlanNotFound)?
+            .plan()
+            .voting_token();
+
+        let stake = StakeControl::new_with(voting_token, &self.accounts);
 
         self.votes = self
             .votes
--- a/chain-impl-mockchain/src/ledger/ledger.rs
+++ b/chain-impl-mockchain/src/ledger/ledger.rs
@@ -18,9 +18,7 @@ use crate::fee::{FeeAlgorithm, LinearFee};
 use crate::fragment::{BlockContentHash, BlockContentSize, Contents, Fragment, FragmentId};
 use crate::rewards;
 use crate::setting::ActiveSlotsCoeffError;
-use crate::stake::{
-    PercentStake, PoolError, PoolStakeInformation, PoolsState, StakeControl, StakeDistribution,
-};
+use crate::stake::{PercentStake, PoolError, PoolStakeInformation, PoolsState, StakeDistribution};
 use crate::tokens::identifier::TokenIdentifier;
 use crate::tokens::minting_policy::MintingPolicyViolation;
 use crate::transaction::*;
@@ -307,8 +305,6 @@ pub enum Error {
     VoteTallyProofFailed,
     #[error("Vote tally decryption failed")]
     VoteTallyDecryptionFailed,
-    #[error("Vote Tally vote plan is not in the ledger")]
-    VotePlanNotFound,
     #[error("Pool update payload signature failed")]
     PoolUpdateSignatureFailed,
     #[error("Pool update last known registration hash doesn't match")]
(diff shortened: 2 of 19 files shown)


chain-impl-mockchain/src/accounting/account/mod.rs
```diff
@@ -208,6 +208,16 @@ impl<ID: Clone + Eq + Hash, Extra: Clone> Ledger<ID, Extra> {
             .map_err(|e| e.into())
     }
 
+    pub fn token_get_total(&self, token: &TokenIdentifier) -> Result<Value, ValueError> {
+        let values = self
+            .0
+            .iter()
+            .filter_map(|(_, account_state)| account_state.tokens.lookup(token))
+            .cloned();
```
eugene-babichenko, line 216: Don't like that we need to do `n` lookups but that's what it is, maybe this part would need a bit of refactoring.

I was thinking of keeping the total supply of tokens (and native currency as well) in a separate part of the ledger state instead of re-computing them every time. Could be either computed as new tokens are minted/burnt, or be a lazy computation, e.g. some memoization-style thing that is invalidated upon minting/burning tokens and is computed on demand.

Another possible workaround is making the data structure into

```
- Token identifier
  - Account identifier
    - Value
```

instead of the current

```
- Account identifier
  - Token identifier
    - Value
```

but we need to see how this would impact other use cases. My take on this is that it would increase load upon creating API responses and this is why the second approach was selected.

This is not critical _right now_ and we can create a task from this comment. But we will need to check how this impacts the performance before the release. 

What we can actually do _right now_ is to stop using it when it is not actually required in which we currently do when tallying.

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -1176,7 +1178,18 @@ impl Ledger {
...
         }
 
-        let stake = StakeControl::new_with(&self.accounts, &self.utxos);
+        let voting_token = self
+            .votes
+            .plans
+            .lookup(tally.id())
+            // XXX: unsure about this.  This would be handled anyway inside
+            // `apply_commitee_result`, although with a different error.  But it needs to be
+            // checked here, unless this logic is moved to the VotePlanManager somehow.
```
eugene-babichenko, line 1187: This part is looking good to me, the error still needs to be processed. Although if you are _absolutely_ sure that `.ok_or()` is never triggered you can replace it with `.expect()`. Let's agree on either of those and remove the comments.

eugene-babichenko (commented):
Fine by me mostly, although I am a bit concerned about `token_get_total` performance. I think this should never be triggered when a vote plan is create with the `OffChain` action (which is pretty much every current deployment).

Another concern is the refactoring of the stake control. First, it mentions voting tokens, while it is meant to be a generic token and outside of the voting context it should be just `token`.

Second, we just transformed something that was meant to be used with the native currency for tokens and I am a bit wary of that. At least the placement of this structure in the project looks slightly weird to me now, although I am not exactly sure what we are supposed to do with that and I am open to suggestions (including the suggestion to leave it as it is). Maybe I am overthinking it.

enzoc4:
> although I am a bit concerned about token_get_total performance.

This is only used in one test, though. For the vote tally optimization the `assigned()` function in `StakeControl` is used. It is the same problem though, because creating the `StakeControl` also loops through all the accounts, but it's only for the tally.

That said, we can avoid that too. Maybe we should ditch `StakeControl` given that we are not using `Group` addresses now.  The tally can just ask the ledger for the voting power, and we can keep track of the total distribution as you mentioned (maybe only if the vote is private, given that's only used for the optimizations I think).

chain-impl-mockchain/src/stake/controlled.rs
```diff
@@ -2,13 +2,14 @@ use crate::{
...
-use chain_addr::{Address, Kind};
 use imhamt::Hamt;
 use std::{collections::hash_map::DefaultHasher, fmt, num::NonZeroU64};
 
 #[derive(Default, Clone, Eq, PartialEq)]
+// TODO: rename this to WeightControl? VotingPowerControl?  It's a form of Stake, technically, but
+// I'm not sure about using the same term for two different things.
 pub struct StakeControl {
     assigned: Stake,
     unassigned: Stake,
```
Zeegomo, line 15: do we still have the concept of unassigned stake with tokens?

Zeegomo (commented):
Well, now from the data flow point of view `StakeControl` is just a slow way to copy a trie + indication of the voting token.

We iterate once to copy the information from all the accounts into a Hamt and then access the trie again when we actually need the information. 
Can we simply copy the root of the account state trie and lazily fetch the voting power from there?

chain-impl-mockchain/src/stake/controlled.rs
enzoc4: I don't think so, not sure if we could in the future.

enzoc4:
>  Can we simply copy the root of the account state trie and lazily fetch the voting power from there?

Yeah, that makes more sense, as I said in my other comment. Probably the whole `controlled` module is a bit redundant now. Especially if we are never going to allow tokens for utxos.

> Well, now from the data flow point of view StakeControl is just a slow way to copy a trie + indication of the voting token.

I'd argue that was already the case, tbh. Looping through the utxo ledger was necessary, but I'm not sure the loop through the accounts was ever needed.

Zeegomo:
> Especially if we are never going to allow tokens for utxos.

We shouldn't take this for granted though. It's not scheduled for the close future, but it's not like we already counted that out

enzoc4:
> Let's just revert the changes in StakeControl and have direct access to the account trie from the voting part.

I don't mind that, but what for though? AFAIK is not used for anything else, the rewards distribution and the leadership schedule both use different stake distribution data structures.

And if it is used somewhere that I couldn't find, if it's not used in the ledger I feel like the chances of that being incorrect are high. At minimum it may be a good idea to put in under #[cfg(test)] I think.

Zeegomo:
> AFAIK is not used for anything else, the rewards distribution and the leadership schedule both use different stake distribution data structures.
And if it is used somewhere that I couldn't find, if it's not used in the ledger I feel like the chances of that being incorrect are high. At minimum it may be a good idea to put in under #[cfg(test)] I think.

+1 for removing it (in another PR as well) if not needed, otherwise we end up with legacy code for which we don't remember the purpose, it's not tested and falls behind the other parts of the system and creates confusion.

enzoc4:
I also have some doubts about the _stake_ term overall, and the `Stake` type in particular. It's basically everywhere in `tally.rs`.

Should we move away to something like `Value`? Or to a brand new type... `VotingPower`, or actually, we also have a `Weight` type whose purpose is not completely clear to me. 

My opinion is that _stake_ should only refer to the things that participate in the PoS protocol, but if someone has more understanding of this part of the code I'd like to hear their opinion.

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -997,11 +995,20 @@ impl Ledger {
...
-                        return Err(Error::OwnerStakeDelegationInvalidTransaction);
+                let account_id = match tx
+                    .inputs()
+                    .iter()
+                    .map(|input| input.to_enum())
+                    .zip(tx.witnesses().iter())
+                    .next()
+                    .unwrap()
+                {
+                    (InputEnum::AccountInput(account_id, _), Witness::Account(_, _)) => account_id
```
Zeegomo, line 1006: why do we care about the witness here? I thought it was already verified in the `apply_transaction` call before

chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -1377,18 +1389,15 @@ mod tests {
             .vote(identifier.clone(), second_vote_cast.payload.clone())
             .unwrap();
 
-        let mut stake_controlled = StakeControl::new();
-        stake_controlled =
-            stake_controlled.add_to(identifier.to_single_account().unwrap(), Stake(51));
-        stake_controlled = stake_controlled.add_unassigned(Stake(49));
+        let (ledger, token) = ledger_with_tokens(identifier.clone());
```
Zeegomo, line 1392: nit: not immediately clear why this works

Also now 100% of tokens are voting in favor of the proposal, not sure if it matters here though

chain-impl-mockchain/src/accounting/account/mod.rs
```diff
@@ -263,13 +273,27 @@ mod tests {
                 .take(stake_pool_size)
                 .collect::<Vec<_>>();
 
+            let voting_tokens_size = usize::arbitrary(gen);
+            let arbitrary_voting_tokens = iter::from_fn(|| Some(TokenIdentifier::arbitrary(gen)))
```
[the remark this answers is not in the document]
enzoc4: I guess I should put a max to it, and actually thought about it, but I wasn't sure what to use for it. The accounts are also generated in the same way, though, `let account_size = std::cmp::max(usize::arbitrary(gen), 1);`, and that's been there since 2019, so I think it's probably not that risky.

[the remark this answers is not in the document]
Zeegomo: I mean, it's not too much of a problem now, the worst that can happen is that at some point it will cause a problem and we already know what to do

chain-impl-mockchain/src/ledger/ledger.rs
enzoc4: Well, I changed the ledger to only allow voting with normal accounts (and not multisig accounts), because we were not taking the stake of those into account before, and we're still not doing it now.

But account inputs are the same for both, it's the witness type the thing that discriminates which kind of account is being used.

chain-impl-mockchain/src/vote/manager.rs
enzoc4: > nit: not immediately clear why this works

what part exactly? maybe I should add an extra account that doesn't cast any vote to compensate for that `unassigned` stake? Because there is no more `unassigned` stake anymore.

chain-impl-mockchain/src/ledger/ledger.rs
Zeegomo: It feels like there's a bit of a mismatch in the domain representation if we have to look at the witness to differentiate account types that comes with different features.
Like, the witness should really just be the spending credential for whatever is the input, and not be used for anything else or carry this additional information IMHO.

Zeegomo: I looks to me like the only type to appear in public interfaces should be `UnspecifiedAccountIdentifier` or `AccountIdentifier`, as there's no particular reason to prefer any of the two in term of functionalities (except for missing implementation, but at that point it should just return an error).

chain-impl-mockchain/src/vote/manager.rs
Zeegomo: > what part exactly?

ops, forget this, it's a leftover

chain-impl-mockchain/src/ledger/ledger.rs
enzoc4: >  Like, the witness should really just be the spending credential for whatever is the input, and not be used for anything else or carry this additional information IMHO.

It seems like an optimization to save space in the accounts (so they can be 32 bytes), but I don't know what was Vincent thinking. Multisig accounts seem like a half-baked feature overall.

> I looks to me like the only type to appear in public interfaces should be UnspecifiedAccountIdentifier or AccountIdentifier, as there's no particular reason to prefer any of the two in term of functionalities (except for missing implementation, but at that point it should just return an error).

I changed things to `account::Identifier` because I think it's clearer when reading the code that it only works with normal accounts. When I found the tests using `UnspecifiedAccountIdentifier` I had to go and check if I had to include the multisig ledger or not, only to realize that it we were ignoring it.

And currently `MintToken` uses `account::Identifier` too, so I think it's more consistent overall.

Zeegomo: > And currently `MintToken` uses `account::Identifier` too, so I think it's more consistent overall.

That is only a partial implementation because of time constraints, and it could probably have been extended to `UnspecifiedAccountIdentifier` without too much effort, but there's no reason we should only restrict single accounts in the vote manager IMHO, as the public interface we use there is the same for both.

Not a blocker for this PR, but I think we should polish this a bit and uniform the way we access information

enzoc4: > That is only a partial implementation because of time constraints

Well, everything is a partial implementation because of time constraints (?). 

I still think it's better to throw an error here and remove it when (if) the feature is fully implemented.

Zeegomo (approved):

chain-impl-mockchain/src/ledger/ledger.rs
Zeegomo: > It seems like an optimization to save space in the accounts (so they can be 32 bytes), but I don't know what was Vincent thinking. Multisig accounts seem like a half-baked feature overall.

The wire format can have those optimizations, but after we've validated the witness we could incorporate the info in something like https://github.com/input-output-hk/chain-libs/blob/832f2bbc47f09d95688af8e4c2cef031bddcee74/chain-impl-mockchain/src/transaction/input.rs#L56

enzoc4: > The wire format can have those optimizations, but after we've validated the witness we could incorporate the info in something like 

Oh, yeah, sure... To be honest I was just guessing, I'm not sure what's exactly optimized. Inputs are already [tagged](https://github.com/input-output-hk/chain-libs/blob/832f2bbc47f09d95688af8e4c2cef031bddcee74/chain-impl-mockchain/src/transaction/input.rs#L67) , so I'm not sure why not have a tag there.

My guess is that it's because the multisig ledger was supposed to be unified with the normal account ledger or something like that.

eugene-babichenko (approved):

PR #716, opened 2021-12-20
Update EVM logs

Mr-Leshiy:
Add a new persistent structure for the EVM logs storaging `LogsState`.
Update evm `Ledger` struct and `VirtualMachine` struct with the usage of this new type.

### DIFF
--- a/chain-evm/src/state/trie.rs
+++ b/chain-evm/src/state/trie.rs
@@ -2,6 +2,8 @@ use imhamt::{Hamt, HamtIter, RemoveError};
 
 use std::collections::hash_map::DefaultHasher;
 use std::convert::Infallible;
+use std::error::Error;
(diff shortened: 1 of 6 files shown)


chain-evm/src/state/trie.rs
```diff
@@ -43,6 +43,19 @@ impl<K: Clone + Hash + Eq, V: Clone> Trie<K, V> {
         Self(new_state)
     }
 
+    /// Update or put the element at the key K
+    ///
+    /// If the element is not present, then V is added, otherwise the closure F is apply
```
eugene-babichenko, line 48:
```suggestion
    /// If the element is not present, then V is added, otherwise the closure F is applied
```

eugene-babichenko (approved):
Looking good for the first implementation.

PR #717, opened 2021-12-22
Add error handling during EVM transaction execution

Mr-Leshiy:
Add an EVM transaction error handling during the execution process, expose EVM errors. Also made small refactor, make a generic transaction execution function for the provided evm executor.

### DIFF
--- a/chain-evm/src/machine.rs
+++ b/chain-evm/src/machine.rs
@@ -82,11 +82,10 @@ pub enum Error {
     TransactionError(ExitError),
     #[error("transaction faral error (machine encountered an error that is not supposed to be normal EVM errors, such as requiring too much memory to execute)")]
     TransactionFatalError(ExitFatal),
-    // If the transaction reverts, there are two possible cases,
-    // it can revert because the called contract feels that it does not have enough
-    // gas left to continue, or it can revert for another reason unrelated to gas.
     #[error("transaction has been reverted (machine encountered an explict revert)")]
     TransactionRevertError(ExitRevert),
+    #[error("not enough gas")]
+    NotEnoughGas,
 }
 
 /// Top-level abstraction for the EVM with the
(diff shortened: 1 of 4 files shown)


chain-evm/src/machine.rs
```diff
@@ -98,6 +112,73 @@ fn precompiles(fork: HardFork) -> Precompiles {
...
+            &mut StackExecutor<
+                'runtime,
+                '_,
+                MemoryStackState<'_, 'runtime, VirtualMachine>,
+                Precompiles,
+            >,
+            u64,
+        ) -> (ExitReason, T),
+    {
+        let executable = |gas_limit| {
```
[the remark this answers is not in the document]
Mr-Leshiy: I have added this anonymous function to avoid code duplication, you can see that `executable` is used twice here.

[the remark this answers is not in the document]
enzoc4: Somehow missed that, sorry, you are right.

chain-evm/src/machine.rs
```diff
@@ -74,6 +76,18 @@ pub type Value = U256;
 /// The context of the EVM runtime
 pub type RuntimeContext = Context;
 
+#[derive(Debug, Error, Clone, PartialEq, Eq)]
+pub enum Error {
+    #[error("transaction error: machine returns a normal EVM error")]
```
eugene-babichenko, line 81:
```suggestion
    #[error("transaction error: machine returned a normal EVM error")]
```

chain-evm/src/machine.rs
```diff
@@ -98,6 +112,73 @@ fn precompiles(fork: HardFork) -> Precompiles {
...
+                // used to modify the accounts trie in the VirtualMachine.
+                let (values, logs) = state.deconstruct();
+
+                self.apply(values, logs, delete_empty);
+                // exit_reason
+                Ok((&self.state, &self.logs, val))
+            }
+
+            // WARNING: such approach was followed from the https://github.com/paritytech/frontier/blob/4a77d59d4640e492ae0db43f891929b75876cf46/client/rpc/src/eth.rs#L1447
+            // need to clarify it
```
eugene-babichenko, line 160: What exactly is needed to be clarified? From the docs I can see that we cannot really get an exact `ExitReason` that corresponds to `NotEnoughGas` which is, well, rather annoying. I'd rather have them make it into a separate error. I'll make it a separate task. Executing the contract twice opens up the potential for DDoS attacks.

chain-evm/src/machine.rs
```diff
@@ -98,6 +112,73 @@ fn precompiles(fork: HardFork) -> Precompiles {
...
+            ExitReason::Revert(err) => {
+                // If the user has provided a gas limit or a gas price, then we have executed
+                // with less block gas limit, so we must reexecute with block gas limit to
+                // know if the revert is due to a lack of gas or not.
+                let block_gas_limit = self.environment.block_gas_limit;
+                let ((exit_reason, _), _) = executable(block_gas_limit.as_u64());
+                match exit_reason {
+                    ExitReason::Succeed(_) => Err(Error::NotEnoughGas),
+                    _ => Err(Error::TransactionRevertError(err)),
+                }
```
eugene-babichenko, line 174: My judgement is that it is actually incorrect. The assumption here is that any transaction can fit into the `block_gas_limit`, while there is no such guarantee. Although I cannot immediately see what can be done here, probably nothing if the `evm` crate does not provide the separate out-of-gas error type.

eugene-babichenko (approved):
Noted a couple of things and there is a typo to be fixed, otherwise looking good to merge after fixing these nitpicky things.

chain-evm/src/machine.rs
Mr-Leshiy: As I mentioned before I have followed to the https://github.com/paritytech/frontier/blob/4a77d59d4640e492ae0db43f891929b75876cf46/client/rpc/src/eth.rs#L1447, where considers the same situation. Actually we do not have such assumtion that any transaction can fit with the block_gas_limit, intention of this validation to precisly determine the reason of the rejection of this execution `let ((exit_reason, val), executor) = executable(gas_limit);`.

chain-evm/src/machine.rs
Mr-Leshiy: So your suggestion just proceed with that error info that we got from the EVM, without any clarification and later just update EVM side ?

PR #726, opened 2022-01-04
Fix cfg macro for the EvmTransaction, implement full serde for EvmTransaction

Mr-Leshiy:


chain-impl-mockchain/src/evm/mod.rs
```diff
@@ -216,6 +244,48 @@ impl Readable for EvmTransaction {
...
+            1 => {
+                // CREATE2 Transaction
+                let caller = read_address(buf)?;
+                let value = read_u256(buf)?;
+                let init_code = read_bytecode(buf)?;
+                let salt = read_h256(buf)?;
+                let gas_limit = read_gas_limit(buf)?;
+                let access_list = read_access_list(buf)?;
+
+                buf.expect_end()?;
```
Zeegomo, line 257: I see this now, but I don't think this works in settings other than just the simple struct ser/de where an `EvmTransaction` is not the only element in the `ReadBuf`.
For instance, in the transaction format this is likely not the last thing in the buffer, nor we should limit binary layout because of this.

chain-impl-mockchain/src/evm/mod.rs
```diff
@@ -216,6 +244,48 @@ impl Readable for EvmTransaction {
...
+            2 => {
+                // CALL Transaction
+                let caller = read_address(buf)?;
+                let address = read_address(buf)?;
+                let value = read_u256(buf)?;
+                let data = read_bytecode(buf)?;
+                let gas_limit = read_gas_limit(buf)?;
+                let access_list = read_access_list(buf)?;
+
+                buf.expect_end()?;
```
Zeegomo, line 278: same as above (there is also one in `Create` that I cannot flag as it's already committed to the repo)

chain-impl-mockchain/src/evm/mod.rs
```diff
@@ -283,4 +353,12 @@ mod test {
             }
         }
     }
+
+    quickcheck! {
+        fn evm_transaction_serialization_bijection(b: EvmTransaction) -> bool {
```
Zeegomo, line 358: Good, we should also add `EvmTransaction` to `Fragment::arbitrary` as it would have uncovered the error above

chain-impl-mockchain/src/evm/mod.rs
```diff
@@ -83,7 +111,7 @@ fn serialize_address(
     bb: ByteBuilder<EvmTransaction>,
     caller: &Address,
 ) -> ByteBuilder<EvmTransaction> {
-    bb.u8(0).bytes(caller.as_fixed_bytes())
```
Zeegomo, line 86: do you know why was the 0 there? Is it some sort of tag?

Zeegomo (changes requested):

chain-impl-mockchain/src/evm/mod.rs
Mr-Leshiy: Not sure, I am looking how it is implemented in the `Aurora` project.

Mr-Leshiy: I have not found any similar serde and even any similar `EvmTransaction` struct that we have. So it is just our custom struct with our custom serde format. Therefore, IMO this one byte seems to be redundant for the `Address` serialization.

saibatizoku: I'm pretty sure that it was an error when refactoring the code into the `serialize_address` function. The `Address` de/serialization is handled as fixed 20 bytes (it is an alias for the `H160` primitive type), so there's no need to use tags.

saibatizoku (approved):
I think that this is ready to be merged.

chain-impl-mockchain/src/fragment/test.rs
```diff
@@ -25,23 +25,23 @@ impl Arbitrary for Fragment {
             12 => Fragment::VoteTally(Arbitrary::arbitrary(g)),
             13 => Fragment::EncryptedVoteTally(Arbitrary::arbitrary(g)),
             14 => Fragment::MintToken(Arbitrary::arbitrary(g)),
+            #[cfg(feature = "evm")]
+            15 => Fragment::Evm(Arbitrary::arbitrary(g)),
```
Zeegomo, line 33: this can panic on non-evm builds as `15` is not covered

PR #728, opened 2022-01-11
add TokenTotals to the ledger

enzoc4:
to avoid looping through all the accounts when computing the total
voting power, which is needed to compute the ratio of acceptance in the
tally.

follow up on:

https://github.com/input-output-hk/chain-libs/pull/713#discussion_r773051247

I wanted to put this in a separate PR because

- This is not strictly necessary.
- There are other ways of doing it, this is just one way.
- I haven't really measured how much of an impact it would have, and if it is indeed better.


eugene-babichenko:
> but I don't know if merging to a feature branch is a good idea.

That is generally OK, but I won't wait for this PR and merge the original one right away so that we can unblock the downstream tasks. You can then rebase this onto master.

eugene-babichenko (approved):

PR #729, opened 2022-01-12
move Environment field from EvmConfigParams to Ledger

Mr-Leshiy:
It is more relevant to store Environment field in the Ledger , because Environment describes the current state of the which is needed for the EVM.

### DIFF
--- a/chain-impl-mockchain/src/config.rs
+++ b/chain-impl-mockchain/src/config.rs
@@ -14,9 +14,7 @@ use chain_core::packer::Codec;
 use chain_core::property;
 use chain_crypto::PublicKey;
 #[cfg(feature = "evm")]
-use chain_evm::machine::{BlockCoinBase, BlockHash, Config, Environment, Origin};
-#[cfg(feature = "evm")]
-use std::convert::TryInto;
+use chain_evm::machine::Config;
 use std::{
     fmt::{self, Display, Formatter},
     io::{self, Cursor, Write},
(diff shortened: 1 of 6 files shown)


chain-impl-mockchain/src/config.rs
```diff
@@ -822,78 +792,23 @@ impl ConfigParamVariant for CommitteeId {
...
     fn from_payload(payload: &[u8]) -> Result<Self, Error> {
         let mut rb = ReadBuf::from(payload);
 
         // Read EvmConfig and match hard fork variant
-        use EvmConfig::*;
         let config = match rb.get_u8()? {
-            n if n == Istanbul as u8 => Istanbul,
-            n if n == Berlin as u8 => Berlin,
+            n if n == EvmConfig::Istanbul as u8 => EvmConfig::Istanbul,
+            n if n == EvmConfig::Berlin as u8 => EvmConfig::Berlin,
```
eugene-babichenko, line 807: Can it be transformed into just this?

```rust
        let config = match rb.get_u8()? {
            EvmConfig::Istanbul as u8 => EvmConfig::Istanbul,
            EvmConfig::Berlin as u8 => EvmConfig::Berlin,
            _ => return Err(Error::InvalidTag),
        };
```

Mr-Leshiy: I have tried, unfortunately not possible

eugene-babichenko (approved):

PR #730, opened 2022-01-18
Update modify_account function

Mr-Leshiy:
Small refactor, update `modify account` function.

### DIFF
--- a/chain-evm/src/machine.rs
+++ b/chain-evm/src/machine.rs
@@ -376,9 +376,9 @@ impl<'runtime> ApplyBackend for VirtualMachine<'runtime> {
                         for (index, value) in apply_storage {
                             account.storage = if value == crate::state::Value::default() {
                                 // value is full of zeroes, remove it
-                                account.storage.clone().remove(&index)
+                                account.storage.remove(&index)
                             } else {
-                                account.storage.clone().put(index, value)
+                                account.storage.put(index, value)
                             }
                         }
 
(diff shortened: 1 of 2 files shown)


chain-evm/src/state/account.rs
```diff
@@ -32,35 +32,23 @@ impl Account {
...
+    ///
+    /// If the element is not present, the closure F is apllied to the Default::default() value,
+    /// otherwise the closure F is applied to the found element.
+    /// If the closure returns None, then the key is deleted
+    pub fn modify_account<F>(self, address: Address, f: F) -> Self
+    where
+        F: FnOnce(Account) -> Option<Account>,
+    {
+        let account = match self.get(&address) {
+            Some(account) => account.clone(),
```
Zeegomo, line 45: nit: if we take ownership there's no need to clone here

chain-evm/src/machine.rs
```diff
@@ -361,26 +361,33 @@ impl<'runtime> ApplyBackend for VirtualMachine<'runtime> {
...
+                        if reset_storage {
+                            account.storage = Default::default();
+                        }
+
+                        // iterate over the apply_storage keys and values
+                        // and put them into the account.
+                        for (index, value) in apply_storage {
+                            account.storage = if value == crate::state::Value::default() {
+                                // value is full of zeroes, remove it
+                                account.storage.clone().remove(&index)
```
Zeegomo, line 379: nit: do we need to clone here?

chain-evm/src/state/account.rs
Mr-Leshiy: Unfortunately we do not take an ownership in this case, because `get(&address)` returns a reference of the `Account` type

Zeegomo: I though there was a `remove` method which would return the previous value

Mr-Leshiy: Not actually, `remove` method return the new state of the trie
```
pub fn remove(self, key: &K) -> Self
```

Zeegomo (approved):

PR #740, opened 2022-02-02
add EVM environment params to ledger settings

saibatizoku:
EVM environment parameters evolve along with the current block state in the EVM ledger, and are used by the virtual machine to perform calculations.

This is separate from the configurations defined by the Hard-Fork in use.

This branch is a work in progress.

### DIFF
--- a/chain-evm/src/machine.rs
+++ b/chain-evm/src/machine.rs
@@ -26,10 +26,7 @@ use crate::{
 };
 
 /// Export EVM types
-pub use evm::{
-    backend::{Log, MemoryVicinity as Environment},
-    Config, ExitReason,
-};
+pub use evm::{backend::Log, Config, ExitReason};
 
 /// An address of an EVM account.
 pub type Address = H160;
@@ -55,6 +52,9 @@ pub type BlockDifficulty = U256;
 /// A block's gas limit.
 pub type BlockGasLimit = U256;
 
+/// A block's base fee per gas.
+pub type BlockBaseFeePerGas = U256;
+
 /// A block's origin
 pub type Origin = H160;
 
@@ -69,6 +69,21 @@ pub type GasPrice = U256;
 
 /// Gas limit for EVM operations.
 pub type GasLimit = U256;
+
+#[derive(Clone, Default, Debug, PartialEq, Eq)]
+/// EVM Environment parameters needed for execution.
+pub struct Environment {
+    pub gas_price: GasPrice,
+    pub origin: Origin,
+    pub chain_id: ChainId,
+    pub block_hashes: BlockHashes,
+    pub block_number: BlockNumber,
+    pub block_coinbase: BlockCoinBase,
+    pub block_timestamp: BlockTimestamp,
+    pub block_difficulty: BlockDifficulty,
+    pub block_gas_limit: BlockGasLimit,
+    pub block_base_fee_per_gas: BlockBaseFeePerGas,
+}
 
 /// Integer of the value sent with an EVM transaction.
 pub type Value = U256;
--- a/chain-impl-mockchain/src/ledger/evm.rs
+++ b/chain-impl-mockchain/src/ledger/evm.rs
@@ -99,7 +99,11 @@ impl Ledger {
                 self.logs = new_logs.clone();
             }
         }
-        // update the environment
+
+        self.update_environment();
+        Ok(())
+    }
+    fn update_environment(&mut self) {
         let next_number = self.environment.block_number + BlockNumber::one();
         // this is a simplified block hash calculation
         let next_hash: BlockHash = <[u8; 32]>::from(next_number).into();
@@ -110,7 +114,6 @@ impl Ledger {
         );
         self.environment.block_number = next_number;
         // TODO: update block timestamp
-        Ok(())
     }
 }
 
--- a/chain-impl-mockchain/src/ledger/evm.rs
+++ b/chain-impl-mockchain/src/ledger/evm.rs
@@ -1,7 +1,7 @@
 use crate::evm::EvmTransaction;
 use crate::ledger::Error;
 use chain_evm::{
-    machine::{Config, Environment, VirtualMachine},
+    machine::{BlockHash, BlockNumber, Config, Environment, VirtualMachine},
     state::{AccountTrie, Balance, LogsState},
 };
 
@@ -62,7 +62,6 @@ impl Ledger {
                 // update ledger state
                 self.accounts = new_state.clone();
                 self.logs = new_logs.clone();
-                Ok(())
             }
             EvmTransaction::Create2 {
                 caller,
@@ -84,7 +83,6 @@ impl Ledger {
                 // update ledger state
                 self.accounts = new_state.clone();
                 self.logs = new_logs.clone();
-                Ok(())
             }
             EvmTransaction::Call {
                 caller,
@@ -99,9 +97,20 @@ impl Ledger {
                 // update ledger state
                 self.accounts = new_state.clone();
                 self.logs = new_logs.clone();
-                Ok(())
             }
         }
+        // update the environment
+        let next_number = self.environment.block_number + BlockNumber::one();
+        // this is a simplified block hash calculation
+        let next_hash: BlockHash = <[u8; 32]>::from(next_number).into();
+        self.environment.block_hashes.insert(0, next_hash);
+        assert_eq!(
+            BlockNumber::from(self.environment.block_hashes.len()),
+            next_number
+        );
+        self.environment.block_number = next_number;
+        // TODO: update block timestamp
+        Ok(())
     }
 }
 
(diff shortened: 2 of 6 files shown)


eugene-babichenko (changes requested):
Why in the ledger settings? Most of these parameters are supposed to be dynamic and either derived from the blockchain state or existing parameters if exclude the ones that cannot be defined yet.

chain-evm/src/machine.rs
```diff
@@ -70,6 +70,21 @@ pub type GasPrice = U256;
...
 pub type GasLimit = U256;
 
+#[derive(Clone, Default, Debug, PartialEq, Eq)]
+/// EVM Environment parameters needed for execution.
+pub struct Environment {
+    pub gas_price: GasPrice,
+    pub origin: Origin,
+    pub chain_id: ChainId,
+    pub block_hashes: BlockHashes,
+    pub block_number: BlockNumber,
```
eugene-babichenko, line 80: This should be taken from the ledger state and updated every time there is a new block.

chain-evm/src/machine.rs
```diff
@@ -70,6 +70,21 @@ pub type GasPrice = U256;
...
+    pub block_coinbase: BlockCoinBase,
+    pub block_timestamp: BlockTimestamp,
+    pub block_difficulty: BlockDifficulty,
+    pub block_gas_limit: BlockGasLimit,
```
eugene-babichenko, line 84: I agree that this should be taken from the settings.

chain-evm/src/machine.rs
```diff
@@ -70,6 +70,21 @@ pub type GasPrice = U256;
 /// Gas limit for EVM operations.
 pub type GasLimit = U256;
 
+#[derive(Clone, Default, Debug, PartialEq, Eq)]
+/// EVM Environment parameters needed for execution.
+pub struct Environment {
+    pub gas_price: GasPrice,
+    pub origin: Origin,
+    pub chain_id: ChainId,
+    pub block_hashes: BlockHashes,
```
eugene-babichenko, line 79: What is that supposed to be?

eugene-babichenko (changes requested):
We should have a separate settings object for EVM that would let us specify all constants that cannot be derived from either EVM or jormungandr ledger state. My understanding is that this includes `gas_price`, `block_gas_limit` and possibly `block_base_fee_per_gas`.

For the values that are constants and can be derived from the ledger (`chain_id`) they should be initialized during ledger init.

`block_number` should be updated when there is a new block.

For the rest let's put dummy values and concentrate on easy fixes.

chain-evm/src/machine.rs
```diff
@@ -70,6 +70,21 @@ pub type GasPrice = U256;
...
+    pub block_number: BlockNumber,
+    pub block_coinbase: BlockCoinBase,
+    pub block_timestamp: BlockTimestamp,
+    pub block_difficulty: BlockDifficulty,
+    pub block_gas_limit: BlockGasLimit,
+    pub block_base_fee_per_gas: BlockBaseFeePerGas,
```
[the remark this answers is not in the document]
saibatizoku: This was introduced with [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) in the London HF, changing the way that gas is consumed.

chain-evm/src/machine.rs
saibatizoku: This is a vec with a hash for each block in the chain, with the last item being the current block hash.

Same as with block_number, this needs to be updated everytime there is a new block.

chain-evm/src/machine.rs
eugene-babichenko: Fine, then go for it :)

chain-evm/src/machine.rs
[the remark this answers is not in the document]
eugene-babichenko: Judging from this

> The proposal in this EIP is to start with a base fee amount which is adjusted up and down by the protocol based on how congested the network is.

we can set to `0` or any arbitrary constant as we are not using it. Because there is no fee market in our implementation and fees are static.

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -99,9 +97,23 @@ impl Ledger {
                 // update ledger state
                 self.accounts = new_state.clone();
                 self.logs = new_logs.clone();
-                Ok(())
             }
         }
+
+        self.update_environment();
```
Mr-Leshiy, line 103: Updating environment should be for every new block applying, not for every evm transaction.

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -99,9 +97,23 @@ impl Ledger {
...
+        Ok(())
+    }
+    fn update_environment(&mut self) {
+        let next_number = self.environment.block_number + BlockNumber::one();
+        // this is a simplified block hash calculation
+        let next_hash: BlockHash = <[u8; 32]>::from(next_number).into();
```
Mr-Leshiy, line 109: I assume we can get `next_hash` from the real block hash value

Mr-Leshiy (changes requested):

chain-impl-mockchain/src/ledger/evm.rs
saibatizoku: Good point. I'm taking a look at how this can be done.

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -1709,6 +1723,9 @@ impl ApplyBlockLedger {
             }
         };
 
+        #[cfg(feature = "evm")]
+        new_ledger.evm.update_environment();
```
Mr-Leshiy, line 1727: nit, I do not think that invoke of the `update_environment();` in the `finish()` function, we need to update evm env before each transaction execution. So we need to move into the `apply_block()` ledger function, before applying all the fragments.

Also I think that we should modify `update_environment()` function to the following form:
```
fn update_environment(block_date: BlockDate, block_number: ChainLength, block_hash: Hash) {

...
} 
```
Most of these data we can get from the `metadata: &HeaderContentEvalContext` argument from the `apply_block()` function
```
pub fn apply_block(
        &self,
        ledger_params: LedgerParameters,
        contents: &Contents,
        metadata: &HeaderContentEvalContext,)
```
The last missing part `block_hash` which we can not get inside `apply_block` function currently I suggest to modify `HeaderContentEvalContext` struct to the following form:
```
pub struct HeaderContentEvalContext {
   #[cfg(feature = "evm")]
    pub(crate) content_hash: BlockContentHash,
    pub(crate) block_date: BlockDate,
    pub(crate) chain_length: ChainLength,
    pub(crate) content_hash: BlockContentHash,
    pub(crate) consensus_eval_context: ConsensusEvalContext,
}
```
And fill this field before invoking `apply_block` function to the corresponds block hash value

Mr-Leshiy (changes requested):

chain-impl-mockchain/src/ledger/ledger.rs
saibatizoku: > nit, I do not think that invoke of the `update_environment();` in the `finish()` function, we need to update evm env before each transaction execution. So we need to move into the `apply_block()` ledger function, before applying all the fragments.
> 
> Also I think that we should modify `update_environment()` function to the following form:
> 
> ```
> fn update_environment(block_date: BlockDate, block_number: ChainLength, block_hash: Hash) {
> 
> ...
> } 
> ```
> 
> Most of these data we can get from the `metadata: &HeaderContentEvalContext` argument from the `apply_block()` function
> 
> ```
> pub fn apply_block(
>         &self,
>         ledger_params: LedgerParameters,
>         contents: &Contents,
>         metadata: &HeaderContentEvalContext,)
> ```
> 
> The last missing part `block_hash` which we can not get inside `apply_block` function currently I suggest to modify `HeaderContentEvalContext` struct to the following form:
> 
> ```
> pub struct HeaderContentEvalContext {
>    #[cfg(feature = "evm")]
>     pub(crate) content_hash: BlockContentHash,
>     pub(crate) block_date: BlockDate,
>     pub(crate) chain_length: ChainLength,
>     pub(crate) content_hash: BlockContentHash,
>     pub(crate) consensus_eval_context: ConsensusEvalContext,
> }
> ```
> 
> And fill this field before invoking `apply_block` function to the corresponds block hash value

Definitely agree with you. This should go in a separate PR.

chain-impl-mockchain/src/ledger/evm.rs
saibatizoku: I changed the signature for this function, it now uses the `metadata.content_hash` value, converted into a `BlockHash` type, inserts it into the `block_hashes` Vec, and updates the `block_number`.

Mr-Leshiy (approved):
LGTM. 
@saibatizoku, can you create a ticket with `HeaderContentEvalContext` update which we have discussed to not forget to do it, or you would prefer to do it as a part to that ticket that you are working currently ?

saibatizoku:
> LGTM. @saibatizoku, can you create a ticket with `HeaderContentEvalContext` update which we have discussed to not forget to do it, or you would prefer to do it as a part to that ticket that you are working currently ?

I'll create a new ticket for it, @Mr-Leshiy

PR #743, opened 2022-02-04
Fix issue with EVM gas fees

Mr-Leshiy:


chain-impl-mockchain/src/tokens/identifier.rs
```diff
@@ -138,7 +138,7 @@ mod tests {
         let result = TokenIdentifier::read(&mut buf);
         let left = Ok(id);
         assert_eq!(left, result);
-        assert_eq!(buf.get_slice_end(), &[]);
+        assert_eq!(buf.get_slice_end(), &[] as &[u8]);
```
eugene-babichenko, line 141:
```suggestion
        assert!(buf.get_slice_end().is_empty());
```

chain-impl-mockchain/src/tokens/name.rs
```diff
@@ -95,7 +95,7 @@ mod tests {
         let result = TokenName::read(&mut buf);
         let left = Ok(token_name);
         assert_eq!(left, result);
-        assert_eq!(buf.get_slice_end(), &[]);
+        assert_eq!(buf.get_slice_end(), &[] as &[u8]);
```
eugene-babichenko, line 98:
```suggestion
        assert!(buf.get_slice_end().is_empty());
```

chain-impl-mockchain/src/tokens/policy_hash.rs
```diff
@@ -60,7 +60,7 @@ mod tests {
         let result = PolicyHash::read(&mut buf);
         let left = Ok(ph.clone());
         assert_eq!(left, result);
-        assert_eq!(buf.get_slice_end(), &[]);
+        assert_eq!(buf.get_slice_end(), &[] as &[u8]);
```
eugene-babichenko, line 63:
```suggestion
        assert!(buf.get_slice_end().is_empty());
```

eugene-babichenko (changes requested):
I am wondering if there is a better way to produce these tests. Writing all these entries that are basically links to JSON files inside of the test wrapper seems annoying. Maybe a script to generate these before compilation or a proc-macro to process test cases during compilation.

eugene-babichenko (approved):

PR #745, opened 2022-02-09
Incremental tally

enzoc4:
# Change the tally to be done incrementally (with each vote)

Some comments on a few things, briefly paraphrasing the original story. The git history is a bit messy, but I guess we'll squash anyway.

## VoteCast certificate

>    Tally is computed incrementally, e.g. the system takes the current voting power, user’s choice, multiplies them and adds to the tally.

With the current implementation the voting power is taken at the moment of the vote, initially I wanted to keep the immutable snapshot of the `TokenDistribution` inside the `VotePlanManager` from the start, but then the order of `MintToken` fragments and `VotePlan` fragments in the block0 would matter, and I think it's a bit fragile. In any case, the behavior is the same since the voting power won't change.

>    Ballots are no longer stored in the ledger state.
>    The voting ledger stores whether a user voted for the proposal or not.
>    Based on this the ledger prevents a user from casting a vote more than once for the same proposal.

There were a few tests that used the stored payloads for validation, I rewrote those so the assertions are based on the tally result instead. 

## EncryptedTally certificate

> This certificate is no longer needed and should be removed.

### Questions here

I kept the binary tag number unused, mostly because it may lead to a better error if we use a new version of the library to read and old fragment for some reason, but maybe it's better to compact things (there is a TODO for this in the code).

Also, are we updating the `abnf` file? I think we are not, so I didn't bother.

## Getting the state of vote tallying as an API user

> GET /api/v0/vote/active/plans

> The vote tally was available via this endpoint from the start, the only difference now is that it is available for the whole duration of the voting process (e.g. could be non-optional).

This needs to be done in jormungandr too, of course, but I removed the `Option` from the `VoteProposalStatus`. 

There is only one thing that I'd like to clarify here, the `total_stake` in the private tally is now taken at the moment the `statuses` are requested. I didn't want to store it from the beginning for the same thing I mentioned above, if there are `MintToken` certificates processed after the `VotePlan` then we won't consider them, unless we add some extra code to handle that case in `Ledger::new`. 

I don't expect this to be a performance concern at the moment, since we likely won't have that many tokens, but it depends on how often the statuses are requested too. Although, to be honest, I'm not quite convinced the `total_stake` should be part of the tally state, but I guess it's coupled for pragmatic reasons.


chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -1248,58 +1279,6 @@ mod tests {
...
-        };
-
-        let committee_id = match tally_proof {
-            TallyProof::Public { id, .. } => id,
-            TallyProof::Private { id, .. } => id,
-        };
-
-        assert_eq!(
-            vote_plan_manager
-                .start_private_tally(token_distribution, block_date, committee_id)
```
enzoc4, line 1294: I removed this test since `start_private_tally` is no there anymore... And I don't see this testing anything else, but it's weird to delete a test, so I would appreciate a second pair of eyes at this.

Zeegomo:
>  but then the order of MintToken fragments and VotePlan fragments in the block0 would matter, and I think it's a bit fragile

Contrasting feeling for this. On one side, we usually put starting conditions in block0 besides basic account creation, but on the other hand block0 is just a block like all others, and ordering is an important property of blocks.
We should also consider `VotePlans` submitted after block0.

> Also, are we updating the abnf file? I think we are not, so I didn't bother.

I updated it for the update proposal changes, but not sure it reflects everything. I would like it to be maintained, it's an easy way to look at chain formats without having to dig through the code.

> with current implementation the voting power is taken at the moment of the vote
> In any case, the behavior is the same since the voting power won't change

This only works because token transfers are not implemented, but they are not guaranteed to remain so forever (actually there are good reasons to implement that). Please add a big disclaimer in the code for this.

enzoc4:
> Contrasting feeling for this. On one side, we usually put starting conditions in block0 besides basic account creation, but on the other hand block0 is just a block like all others, and ordering is an important property of blocks.
We should also consider VotePlans submitted after block0.

I mean, I think the order should matter... It's just that in this case it's a bit strange, currently the `VotePlanManager` creation is tied to the `VotePlan` certificate, but to me it would make more sense if it's instanced just when it is the actual voting date. Like, to me it makes no sense that a `VotePlan` should only see the tokens that were declared before, since the actual voting will be in the future.

> This only works because token transfers are not implemented, but they are not guaranteed to remain so forever (actually there are good reasons to implement that). Please add a big disclaimer in the code for this.

I'm not sure a disclaimer about what? The issue is that I actually don't know what do we want even if we allowed transfers. My point was that currently both solutions are the same, so there is no much reason to choose one over the other right now.

Like, I think one option could be to take a 'snapshot' at the moment the voting starts, and even if you transfer after that, we would use the initial voting power. In this case I would actually change the current code.

But another option could be to use the voting power at the moment of the vote, and lock them (or lock them before, it's the same) after that. If we allow minting during the voting period then we would also have to wait for the end of the voting period to know the total stake... In this case I think the current way is closest, and I don't feel like a disclaimer adds to much information, where should we even put it? in the vote cast handler?

Zeegomo:
> Like, to me it makes no sense that a VotePlan should only see the tokens that were declared before, since the actual voting will be in the future.

Probably something like freezing when voting starts? There was this idea of forbidding transfers when a voteplan is in the voting stage


> I'm not sure a disclaimer about what?

Just a disclaimer that when we want to implement token transfers we should think of a solution for this. I was thinking to put that in the relevant part of tokens implementation

chain-impl-mockchain/src/fragment/mod.rs
```diff
@@ -64,7 +63,7 @@ pub(super) enum FragmentTag {
     VotePlan = 10,
     VoteCast = 11,
     VoteTally = 12,
-    EncryptedVoteTally = 13,
+    // TODO: change these to 13 14?
```
eugene-babichenko, line 66: Yes please.

chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -117,6 +129,9 @@ pub enum VoteError {
 
     #[error("Error during private tallying {0}")]
     PrivateTallyError(String),
+
+    #[error("It's not possible to cast a vote if the tally is already decrypted")]
+    TallyAlreadyDecrypted,
```
eugene-babichenko, line 134: Is there a specific need for such error? Votes can be cast only during the vote period and can be decrypted during tally period.

chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -146,17 +183,52 @@ impl ProposalManager {
...
+                }
+                (IncrementalTally::Private(_), ValidatedPayload::Public(_)) => {
+                    return Err(VoteError::InvalidPayloadType {
+                        received: PayloadType::Public,
+                        expected: PayloadType::Private,
+                    })
+                }
+                (IncrementalTally::Decrypted(_), _) => {
+                    return Err(VoteError::TallyAlreadyDecrypted);
+                }
```
eugene-babichenko, line 223: Reaching this match arm should be practically impossible. If we try to vote when the tally is already decrypted this should mean we already reached the end of the voting period. I think `unreachable` or `panic` with a comment is a viable option here.

chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -146,17 +183,52 @@ impl ProposalManager {
...
+            }
+        } else {
+            self.tally.clone()
```
eugene-babichenko, line 226: And here I think is a good place for an error. If the stake is zero, we are going to write a no-op transactions which opens up potential for attacks.

eugene-babichenko:
@Zeegomo @enzoc4 you expressed some good concerns, but I think that we should come back to this when minting and transfers during the blockchain run are actually possible.

eugene-babichenko:
I think we can also remove `rayon` from dependencies.

eugene-babichenko:
Ran benchmarks on my laptop, no significant impact on vote casting performance.

chain-impl-mockchain/src/vote/manager.rs
enzoc4: Now that you mention it, I just noticed that I accidentally removed the committee date validation checks for the private tally... Or well, rather, they were only done when starting the encrypted tally, and we don't do that anymore, so they need to be done when decrypting the tally, I guess.

Although, this would mean that the decrypted shared needs to be posted during the committee time, which is not the case right now (we don't have an upper bound with the current code in master)

enzoc4: Well, unless we only check the lower bound, I guess. Then the `committee_end` would have no meaning for the private tally...

I think this is another thing that we would need to think about if we remove the public variant of the `VoteTally` certificate. If we trigger the governance actions automatically on the voting period end, then I'm not sure what would be the purpose of the committee dates for public voteplans, for example.

chain-impl-mockchain/src/vote/manager.rs
enzoc4: Good point. Although technically the transactions would only be no-ops if the fees are zero, it's better to just reject them altogether.

chain-impl-mockchain/src/vote/manager.rs
Zeegomo: > Although, this would mean that the decrypted shared needs to be posted during the committee time, which is not the case right now

Mmm, interesting. I think this is an error. From an formal perspective I think it's useful to have an upper bound on committee action (like a timeout) so that we have a clear bound on when a slow committee has to be deemed unresponsive.
This information should be available on chain so that it can be used from smart contracts, scripts, etc.

eugene-babichenko (changes requested):
Could you please add a test for zero token votes? Should be good to go after that.

chain-impl-mockchain/src/vote/manager.rs
enzoc4: > We can deal with this in another PR though

With what? The smart contracts part?

eugene-babichenko (approved):

chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -344,19 +367,24 @@ impl ProposalManager {
 
         let mut result = TallyResult::new(self.options.clone());
         for (choice, &weight) in decrypted_proposal.tally_result.iter().enumerate() {
-            result.add_vote(Choice::new(u8::try_from(choice).unwrap()), weight)?;
+            result
+                .add_vote(Choice::new(u8::try_from(choice).unwrap()), weight)
+                .map_err(|error| match error {
+                    VoteError::InvalidChoice { options, choice } => {
```
Zeegomo, line 373: How can this actually happen? We build the tally with the correct number of options ourselves

enzoc4: Well, I think it can't, `Tally::verify` should take care of validating that, I think?

PR #750, opened 2022-02-16
address vit-2568, remove panic on babysteps if max_value = 0

cameron1024:
Makes the `BabyStepTable::generate` and `generate_with_balance` functions return `Option<Self>`, and also changes `ValidatedTally::decrypt_tally` take an option to a baby step table. In the case where the table is `None`, this means there were no votes cast, so we return a `Tally` with no votes

Some open questions about the `batch_decrypt` impl:
 - it requires the public and private keys of the members, is this alright? I largely copy/pasted the impl from `chain-impl-mockchain/src/testing/builders/vote.rs` but it seems odd having other people's private keys, is there something I'm missing?
 -  I'm not sure how to get the max vote count from bytes for `EncryptedVote::from_bytes`, I've left a `todo!()` there


chain-vote/src/tally.rs
```diff
@@ -223,11 +223,17 @@ impl ValidatedTally {
     pub fn decrypt_tally(
         &self,
         max_votes: u64,
-        table: &TallyOptimizationTable,
+        table: &Option<TallyOptimizationTable>,
```
eugene-babichenko, line 226:
```suggestion
        table: Option<&TallyOptimizationTable>,
```

seems more sensible

eugene-babichenko: @Zeegomo any thoughts on this?

chain-vote/src/math/babystep.rs
```diff
@@ -36,7 +36,11 @@ impl BabyStepsTable {
     ///
     /// For example, a balance of 2 means that the table will precompute 2 times more
     /// baby steps than the standard O(sqrt(n)), 1 means symmetrical steps.
-    pub fn generate_with_balance(max_value: u64, balance: u64) -> Self {
+    pub fn generate_with_balance(max_value: u64, balance: u64) -> Option<Self> {
```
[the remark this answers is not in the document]
Zeegomo: Or we could use `NonZerou64`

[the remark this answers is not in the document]
eugene-babichenko: Yeah, just using `NonZero*` types, why haven't I though about them in this place :)

chain-vote/src/tally.rs
eugene-babichenko: Thought in this?

> Another way to go there is to cover this special case with an enum and protect ourselves from such an error on the type system level. Something like max_votes_with_optimization_table: `Option<(NonZeroU64, &TallyOptimizationTable)>`

chain-vote/src/math/babystep.rs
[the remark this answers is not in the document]
eugene-babichenko: Also use the non-zero type for balance, because setting zero balance will result in the zero step size. And then we can just remove that `assert!` that was panicking.

chain-vote/src/tally.rs
cameron1024: I've made an enum that looks like:
```rust
enum TallyDecryptStrategy<'a> {
  WithVotes(&'a BabyStepTable, NonZeroU64),
  WithoutVotes,
}
```
This seems to deal with the "ability to misuse the API", as well as opting into null pointer optimization. I'm still not 100% familiar with the code, so it's possible I've chosen the wrong names for things though. API-wise, does this convey the intent better?

chain-vote/src/tally.rs
```diff
@@ -223,11 +223,17 @@ impl ValidatedTally {
...
     ) -> Result<Tally, TallyError> {
-        let r_results = self.decrypt();
-        let votes = baby_step_giant_step(r_results, max_votes, table).map_err(|_| TallyError)?;
-        Ok(Tally { votes })
+        match table {
+            None => Ok(Tally { votes: vec![] }),
```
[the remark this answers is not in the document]
cameron1024: From what I can tell, this is just `r_results.len()`, is that right?

chain-vote/src/tally.rs
Zeegomo: Right, it solves the misuse problem, but I would argue that it does not make a lot of sense to call the function with  `WithoutVotes` variant? It's already clear that the function has nothing do to in that case and the caller already knows the result.

cameron1024: Currently it calls the private `decrypt` function, which in turn accesses private fields, so it seems like there isn't a way to work out the length of the `Tally` vector from outside this module. But I'm not super clear on the model yet so maybe we can get it from something in `vote.rs`?

chain-impl-mockchain/src/testing/builders/vote.rs
```diff
@@ -48,13 +58,22 @@ pub fn decrypt_tally(
...
-            let tally = validated_tally.decrypt_tally(max_votes, &table).unwrap();
+
+            let tally = match (&table, max_votes.try_into()) {
+                (Some(table), Ok(max_votes)) => {
+                    validated_tally.decrypt_tally(table, max_votes).unwrap()
+                }
+                _ => Tally {
+                    votes: vec![0; validated_tally.len()],
+                },
+            };
```
eugene-babichenko, line 69: `decrypt_tally` is used in multiple modules across different repositories. We would have to repeat this code (and the above `TallyOptimizationTable` part) multiple times across different places. To avoid that I would suggest using an structure (let's call it `TallyDecryptor`) defined like that

```rust
struct TallyDecryptor {
    maybe_tally_optimization_table: Option<TallyOptimizationTable>
}
```

This structure should:

1. Initialize `TallyOptimizationTable` or set it to `None` in a conventional `new` method.
2. Expose `TallyDecryptor::decrypt_tally(&self, validated_tally: ValidatedTally, max_votes: u64)` that would encapsulate the code covered by this comment.

An example usage:

```rust
let tally_decryptor = TallyDecryptor::new(absolute_max_votes);
let tally = tally_decryptor.decrypt_tally(validated_tally, max_votes);
```

You can find examples of multiple usage in at least https://github.com/input-output-hk/jormungandr.

Zeegomo: Agree, I feel like currently the API is not at the correct level of abstraction, since exposing `TallyOptimizationTable` is not really relevant to downstream consumers, especially if it causes the problems outlined in this discussion.

Probably we could refine this even further, removing the need for passing `max_votes` and just exposing a function (or struct for a bit more flexibility) for decrypting a batch of `ValidatedTally`. 
E.g. something like:
```
fn batch_decrypt(tallies: &[ValidatedTally]) -> Result<Vec<Tally>> {
    let max_total_stake = tallies.iter().max_total_stake().max();
    if max_total_stake = 0 {
       ...
   }
    let table = ...
    tallies.iter()
             .map(|validate_tally| validated_tally.decrypt_tally(&table, validated_tally.max_total_stake()))
             .collect()
}
```
This would require us to keep track of the maximum total stake inside each `EncryptedTally/ValidatedTally`,  which is rather easy, we already do that but at another level https://github.com/input-output-hk/chain-libs/blob/2744ba2d3980e025b728293cb2ec08372d5f0526/chain-impl-mockchain/src/vote/tally.rs#L74). And we could also have a much more granular control on this value (by tracking https://github.com/input-output-hk/chain-libs/blob/2744ba2d3980e025b728293cb2ec08372d5f0526/chain-vote/src/tally.rs#L120) instead of using the total available stake https://github.com/input-output-hk/chain-libs/blob/2744ba2d3980e025b728293cb2ec08372d5f0526/chain-impl-mockchain/src/vote/manager.rs#L316.

chain-vote/src/tally.rs
```diff
@@ -381,55 +389,47 @@ impl Tally {
...
-        Some(Ok(absolute_max_stake)) => {
+pub fn batch_decrypt(validated_tallies: &[ValidatedTally]) -> Result<Vec<Tally>, TallyError> {
+    let absolute_max_stake = validated_tallies.iter().map(|tally| tally.max_stake).max();
+
+    let absolute_max_stake = match absolute_max_stake {
+        Some(stake) => stake,
+        None => return Ok(vec![]),
+    };
+
+    match absolute_max_stake.try_into() {
```
Zeegomo, line 400: nit: I would say it's a bit less readable to have this `try_into` without explicit type (which I guess is `NonZeroU64`) compared to match for 0 and non zero like
```
match absolute_max_stake {
    0 => 
    non_zero => Table::generate(non_zero.try_into().unwrap())
```

chain-vote/src/math/babystep.rs
```diff
@@ -72,9 +71,10 @@ pub struct MaxLogExceeded;
 /// Solve the discrete log on ECC using baby step giant step algorithm
 pub fn baby_step_giant_step(
     points: Vec<GroupElement>,
-    max_log: u64,
+    max_log: NonZeroU64,
```
Zeegomo, line 74: this can actually be 0

chain-vote/src/tally.rs
```diff
@@ -245,9 +275,11 @@ impl std::ops::Add for EncryptedTally {
             .zip(rhs.r.iter())
             .map(|(left, right)| left + right)
             .collect();
+        let max_stake = std::cmp::max(self.max_stake, rhs.max_stake);
```
Zeegomo, line 278: we need to add this as well

chain-vote/src/tally.rs
```diff
@@ -222,8 +252,8 @@ impl ValidatedTally {
     /// of each resulting plaintext.
     pub fn decrypt_tally(
         &self,
-        max_votes: u64,
         table: &TallyOptimizationTable,
+        max_votes: NonZeroU64,
```
Zeegomo, line 256: this can be zero as well

chain-vote/src/tally.rs
```diff
@@ -357,6 +389,47 @@ impl Tally {
...
+        Ok(absolute_max_stake) => {
+            let table = TallyOptimizationTable::generate_with_balance(
+                absolute_max_stake,
+                1.try_into().unwrap(),
+            );
+
+            let mut result = Vec::with_capacity(validated_tallies.len());
+
+            for validated_tally in validated_tallies {
+                let tally = match validated_tally.max_stake.try_into() {
```
Zeegomo, line 412: nit: by relaxing unnecessary bounds in the `decrypt_tally` function we could rewrite this without the need for this match arm

Zeegomo (changes requested):
We could remove `total_stake` in https://github.com/input-output-hk/chain-libs/blob/6fdab6bf0a4893bb1cf38361f25529cd7b999e9d/chain-impl-mockchain/src/vote/tally.rs#L35 as I don't think it's needed anymore, but it's better in another PR

chain-vote/src/tally.rs
cameron1024: I guess I went for this approach to avoid the unwrap call. What's your preference between the following:
```rust
match absolute_max_stake {
   0 => {},
  non_zero => Table::generate(non_zero.try_into().unwrap()),
}
// or
match NonZeroU64::try_from(absolute_max_stake) {
  Err(_) => },
  Ok(non_zero) => Table::generate(non_zero),
}
```

Zeegomo: Personally probably the first, but it doesn't really matter. Having the explicit cast is already a good step forward, whatever you prefer between these two is fine

chain-vote/src/tally.rs
cameron1024: Looks like this would end up being, `validated_tally.decrypt_tally(&table, validated_tally.max_stake)`. Is it worth moving the second parameter inside the function body as `self.max_stake`?

Zeegomo: yup, good idea

chain-vote/src/tally.rs
```diff
@@ -357,6 +386,42 @@ impl Tally {
...
+    };
+
+    match NonZeroU64::try_from(absolute_max_stake) {
+        // if absolute_max_stake == 0, all stakes are zero, so safe to do a trivial conversion
+        Err(_) => Ok(validated_tallies.iter().map(trivial_convert).collect()),
+        Ok(absolute_max_stake) => {
+            let table = TallyOptimizationTable::generate(absolute_max_stake);
+            let result = validated_tallies
+                .iter()
+                .map(|tally| tally.decrypt_tally(&table).unwrap())
```
Zeegomo, line 408: Let's propagate an error here instead of unwrapping

eugene-babichenko (approved):

Zeegomo (approved):
Good work!

minikin:
@eugene-babichenko @Zeegomo Do we usually have this situation with windows on CI?

Zeegomo:
It's not common, but happened from time to time. I've restarted all jobs just to be sure

minikin:
@cameron1024 @eugene-babichenko @Zeegomo 
Should we set up auto-merge in case all checks are passed?

cameron1024:
I usually don't like auto-merge by default, but if it's something you can press to say "merge this once the checks pass" that could be good. ATM I think I don't have merge permissions anyways, so I think I wouldn't be able to trigger that anyways :/

enzoc4:
Automerge could complicate things if we don't have a PR ready in jormungandr (or any other crate, tbh, since chain-libs is used basically everywhere), which could potentially block other tasks.

That said, if we know we didn't change the public api it should be fine.

PR #751, opened 2022-02-17
Update EVM block timestamp params

saibatizoku:
Uses the `BlockDate` and ledger settings such as `slots_per_epoch`, and `slot_duration`, as well as the `Block0Date` value to calculate the appropriate timestamp each time `apply_block` is called.

### DIFF
--- a/chain-impl-mockchain/src/ledger/ledger.rs
+++ b/chain-impl-mockchain/src/ledger/ledger.rs
@@ -365,7 +365,7 @@ impl Ledger {
             token_totals: TokenTotals::default(),
         };
         #[cfg(feature = "evm")]
-        let ledger = ledger.set_evm_block0();
+        let ledger = ledger.set_evm_block0().set_evm_environment();
         ledger
     }
 
@@ -450,8 +450,7 @@ impl Ledger {
                     Block0Error::InitialMessageNoConsensusLeaderId,
                 ));
             }
-            let ledger = Ledger::empty(settings, static_params, era, pots);
-            ledger
+            Ledger::empty(settings, static_params, era, pots)
         };
 
         let params = ledger.get_ledger_parameters();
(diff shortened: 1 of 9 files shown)


chain-impl-mockchain/src/evm/mod.rs
```diff
@@ -16,7 +16,36 @@ use crate::{
 };
 
 #[cfg(feature = "evm")]
-pub use chain_evm::machine::{BlockGasLimit, Config, Environment, GasPrice};
+pub use chain_evm::machine::{
+    BlockGasLimit, BlockTimestamp, ChainId, Config, Environment, GasPrice,
+};
+
+/// Static parameters for EVM
```
Zeegomo, line 23: this is static but contains `BlockEpoch` which is a dynamic parameter, have I missed something?

saibatizoku: Just a typo, thanks for pointing it out.

Zeegomo (approved):

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -815,10 +824,10 @@ impl Ledger {
         #[cfg(feature = "evm")]
         {
             // Set EVM environment values derived from block0 values
-            new_ledger.evm.environment.chain_id =
-                <[u8; 32]>::from(new_ledger.static_params.block0_initial_hash).into();
-            // TODO: set Origin, coinbase, and set block timestamp when
-            // they are defined. Using default values meanwhile.
+            new_ledger.evm.environment.chain_id = new_ledger.settings.evm_block_params.chain_id;
+            new_ledger.evm.environment.block_timestamp =
```
Mr-Leshiy, line 828: I think it is better to move it inside `update_evm_block()` function

saibatizoku: This has been completely re-worked to go inside a method.

Mr-Leshiy (approved):

minikin:
@saibatizoku  @Mr-Leshiy  
Can we resolve issues here to merge it, please? Or do we have some other dependencies ?

PR #753, opened 2022-02-18
Exclude evm-tests submodule from cargo

Mr-Leshiy:


eugene-babichenko:
The issue with builds was fixed with the appropriate system configuration.

PR #756, opened 2022-02-22
remove redundant stake field from private tally

cameron1024:
As mentioned in #750 , the `total_stake` field in `PrivateTallyState::Encrypted` is redundant, since we track it in `EncryptedTally` now. This PR removes the unnecessary field


chain-impl-mockchain/src/vote/tally.rs
```diff
@@ -98,15 +90,11 @@ impl Tally {
         }
     }
 
-    pub fn private_encrypted(&self) -> Result<(&EncryptedTally, &Value), TallyError> {
+    pub fn private_encrypted(&self) -> Result<(&EncryptedTally, Value), TallyError> {
```
Zeegomo, line 93: is the stake return value used somewhere? I don't think so, as we don't have such thing for the public tally

chain-vote/src/tally.rs
```diff
@@ -116,6 +116,10 @@ impl EncryptedTally {
         }
     }
 
+    pub fn max_stake(&self) -> u64 {
```
Zeegomo, line 119: I don't think this is necessary and exposing this only for `EncryptedTallies` seems a bit strange to me

Zeegomo (changes requested):

chain-impl-mockchain/src/vote/tally.rs
cameron1024: It's used in `ProposalManager` somewhere, I guess it could probably just return a `Result<&EncryptedTally>`

chain-vote/src/tally.rs
```diff
@@ -50,7 +50,7 @@ impl From<(&ElectionPublicKey, &Crs)> for ElectionFingerprint {
 pub struct EncryptedTally {
     r: Vec<Ciphertext>,
     fingerprint: ElectionFingerprint,
-    max_stake: u64,
+    pub max_stake: u64,
```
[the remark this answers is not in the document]
cameron1024: I thought that was what you were suggesting on the `pub fn max_stake(&self) -> u64` above. I guess this way makes me a bit uneasy, since it makes it possible to corrupt the internal state, was there something else you had in mind? FWIW, the only use of the field outside the module is a single function in `ProposalManager`

chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -347,7 +347,7 @@ impl ProposalManager {
             result.add_vote(Choice::new(u8::try_from(choice).unwrap()), weight)?;
         }
 
-        if self.check((*total_stake).into(), governance, &result) {
+        if self.check(Stake(encrypted_tally.max_stake), governance, &result) {
```
Zeegomo, line 350: I think it was a mistake also in the previous implementation, this should be the total available voting power, not the one that participated in the election (which can be obtained just by looking at the results)

Zeegomo (changes requested):

chain-impl-mockchain/src/vote/manager.rs
```diff
@@ -347,7 +347,7 @@ impl ProposalManager {
             result.add_vote(Choice::new(u8::try_from(choice).unwrap()), weight)?;
         }
 
-        if self.check((*total_stake).into(), governance, &result) {
+        if self.check(result.participation(), governance, &result) {
```
Zeegomo, line 350: Sorry if I was misleading, I meant to say we should use something like `token_distribution.get_total()`, which btw is what we do on master. Rebasing should fix this

Zeegomo (commented):
Please fix the lint, otherwise good to merge

PR #761, opened 2022-02-25
Limit ethereum balances to 64 bits

saibatizoku:
To keep compatibility with Jormungadr balances, we need to cap ethereum account balances to 64 bits.


chain-evm/src/machine.rs
```diff
@@ -398,7 +398,9 @@ impl<'runtime> ApplyBackend for VirtualMachine<'runtime> {
                     // If reset_storage is set, the account's balance is
                     // set to be Default::default().
                     self.state = self.state.clone().modify_account(address, |mut account| {
-                        account.balance = balance;
+                        account.balance = balance
+                            .try_into()
+                            .expect("Account balance cannot exceed 64 significant bits");
```
Zeegomo, line 403: Shouldn't we return an error here?
Actually also up in the call stack we unwrap like in https://github.com/input-output-hk/chain-libs/blob/0496dd601aab0c827b7437807fda47612b3091af/chain-evm/src/machine.rs#L194 Am I reading the code wrong or should we return errors in these cases instead of panicking?

Zeegomo (changes requested):

chain-evm/src/machine.rs
saibatizoku: You are right @Zeegomo , this should be returning an error, not panicking.

saibatizoku: Handling these situations as runtime errors is left for future issues.

PR #763, opened 2022-03-03
replace pack_pool_registration with serde traits

cameron1024:
remove extra bytes from `pack_pool_registration` serialization


Zeegomo (approved):
FYI, there are not serde traits but our own (see https://github.com/input-output-hk/chain-libs/tree/master/chain-ser/)

cameron1024:
Ahh yeah, I've got into the bad habit of shortening "serialisation + deserialization" to "serde" generally, not just the crate

PR #764, opened 2022-03-04
Add new SignedCertificate EvmMapping

Mr-Leshiy:


Zeegomo:
Is this supposed to be a one-way or two-way mapping? If the latter (e.g. if the goal is to establish a shared balance between the two mappings) I think it's necessary to include a proof that you own the evm address, otherwise you could link to an arbitrary evm address and drain its balance.

Mr-Leshiy:
@Zeegomo  good point!
At the current implementation I assume we will follow one-way mapping and it should be enough.

- Scenario 1. Mapping already exists for this Jormungandr account:
**returns error**
- Scenario 2. Mapping already exists for this Ethereum address
**returns error**
- Scenario 3. Mapping to the existing Ethereum account with balance but without account mapping
** it should be on the responsibility on the users why they send some assets on the not mapped account.**

So the secure flow for any `EVM` user will be: 
first create `jormungandr` and `EVM` accounts, then map their, then you can interact via the `EVM`.

But we need to make additional research how we should do with the smart-contracts account addresses, is it possible to trigger any transaction for him in the meaning "outside EVM". So maybe we will need to add that type of mapping and validation that you have described.

By the way we even do not have a `Ethereum` crypto signing algorithms, so I assume we can process as it is and later update it.

Zeegomo:
>  Mapping already exists for this Jormungandr account: returns error

why not overwrite? After all it's signed on the Jormungandr side (kind of like a new catalyst registration overwrites a old one) 

> ** it should be on the responsibility on the users why they send some assets on the not mapped account.**

I'm basing my observations off of https://docs.google.com/document/d/1dSGEMWb8Yy-3TfMp49Ju3KRKGYbgejbcSGlCdCWGSEc/edit#, which mentions a shared balance between the two accounts systems. In this case I would not need to do anything on the EVM side. I think it would be sufficient to:
* find a evm address with some balance (possibly not mapped to any jormungandr account)
* create a new jormungandr account with 0 balance
* map the jorm account to the evm address
* drain the evm balance using this shared balance mechanism and for example send all the funds to another jormungandr account I control

Mr-Leshiy:
@Zeegomo 

>find a evm address with some balance (possibly not mapped to any jormungandr account)
create a new jormungandr account with 0 balance
map the jorm account to the evm address

Yes it will be possible but I am not sure that It should be a problem or vulnerability because on the not mapped evm address somehow should appear these funds, so someone was not to attentive and receive their funds on the incorrect account.

>drain the evm balance using this shared balance mechanism and for example send all the funds to another jormungandr account I control

Such situation should be impossible because it will be possible to make transfers evm-to-evm and jor-to-jor accounts

saibatizoku (approved):
Seems good to put the foot in the door.

PR #766, opened 2022-03-07
EVM balance overflows return runtime error

saibatizoku:
Handle account balance overflow by returning a runtime error.

Currently, only external account balances can be checked before transaction execution. Contract account balances remain `U256` values.

If contract account balances need to be restricted, we need to have our custom implementation of the `StackState` trait from SputnikVM

### DIFF
--- a/chain-evm/src/machine.rs
+++ b/chain-evm/src/machine.rs
@@ -19,7 +19,7 @@ use thiserror::Error;
 
 use crate::{
     precompiles::Precompiles,
-    state::{Account, AccountTrie, ByteCode, Error as StateError, Key, LogsState},
+    state::{Account, ByteCode, Error as StateError, Key},
 };
 
 /// Export EVM types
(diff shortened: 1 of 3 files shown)


Mr-Leshiy (commented):
But why not to leave an existing `Balance` struct which was also wrapper for the `U256`, what is the purpose to create a new one `CappedU256` ?

saibatizoku:
> But why not to leave an existing `Balance` struct which was also wrapper for the `U256`, what is the purpose to create a new one `CappedU256` ?

No particular reason at this point, I can revert it to being a wrapper for `U256`. The reason for creating a capped version of U256 was to have multiple aliased types for it, as I tried to limit balances inside the executor. I abandoned this approach, but didn't revert the code.

Mr-Leshiy:
@saibatizoku  IMO `Balance` naming is more clear in this case, so I would prefer to stay with this naming

saibatizoku:
> @saibatizoku IMO `Balance` naming is more clear in this case, so I would prefer to stay with this naming

Agreed. Just pushed the last changes for this, @Mr-Leshiy

Mr-Leshiy (approved):

PR #768, opened 2022-03-08
Update EVM VirtualMachine struct, remove redundant usage of the evm state fields 

Mr-Leshiy:
This update is one from the "EVM mapping" feature implementation.
Purpose of this PR is to remove redundant evm state fields as `AccountTrie`, `LogsState`, `Environment` and leave them only inside `Ledger` struct:
```
pub struct Ledger {
    pub(crate) accounts: AccountTrie,
    pub(crate) logs: LogsState,
    pub(crate) environment: Environment,
    pub(crate) current_epoch: BlockEpoch,
}
```

### DIFF
--- a/chain-evm/src/machine.rs
+++ b/chain-evm/src/machine.rs
@@ -223,6 +223,7 @@ impl<'a, State: EvmState> VirtualMachine<'a, State> {
     }
 
     /// Execute a CREATE transaction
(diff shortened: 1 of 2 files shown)


saibatizoku (approved):
Cool, this refactoring was long overdue. I'll want to use these changes with my branch.

PR #769, opened 2022-03-09
Refactor evm Ledger

Mr-Leshiy:
- Update `EvmState` trait, remove explicit info of the accounts and logs storing, it should be a part of the implementation, not a part of the interface.
- Implement `EvmState` for the `ledger::Ledger` struct, it is needed to get an access to the `account::Ledger` which will be needed for the evm mapping.


saibatizoku (approved):
LGTM

PR #772, opened 2022-03-11
Add EvmMapping fragment handling

Mr-Leshiy:


Mr-Leshiy:
Need firstly merge https://github.com/input-output-hk/chain-libs/pull/773

Mr-Leshiy:
Better firstly merge https://github.com/input-output-hk/chain-libs/pull/772

Zeegomo (changes requested):

chain-evm/src/state/account.rs
```diff
@@ -21,44 +21,54 @@ impl TryFrom<U256> for Balance {
...
     }
 }
 
 impl Balance {
     /// Zero (additive identity) of this type.
     pub fn zero() -> Self {
-        Balance(U256::zero())
+        Balance(0)
     }
     /// Checked addition of `U256` types. Returns `Some(balance)` or `None` if overflow
```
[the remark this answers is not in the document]
Mr-Leshiy: fixed

chain-evm/src/state/account.rs
```diff
@@ -21,44 +21,54 @@ impl TryFrom<U256> for Balance {
...
     /// occurred.
     pub fn checked_add(self, other: U256) -> Option<Balance> {
-        match self.0.checked_add(other) {
-            Some(U256([v, 0, 0, 0])) => Some(Balance::from(v)),
-            _ => None,
+        if let U256([val, 0, 0, 0]) = other {
```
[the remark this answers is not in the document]
Mr-Leshiy: fixed

chain-evm/src/machine.rs
```diff
@@ -129,7 +127,7 @@ pub trait EvmState {
 
     fn contains(&self, address: Address) -> bool;
 
-    fn modify_account<F>(&mut self, address: Address, f: F)
+    fn modify_account<F>(&mut self, address: Address, f: F) -> Result<(), String>
```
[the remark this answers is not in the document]
Mr-Leshiy: fixed

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -1,23 +1,78 @@
...
+
+impl AddressMapping {
+    fn new() -> Self {
+        Self {
+            evm_to_jor: Hamt::new(),
+            jor_to_evm: Hamt::new(),
+        }
+    }
+
+    fn jor_address(&self, evm_id: EvmAddress) -> Option<JorAddress> {
```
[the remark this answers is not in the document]
Mr-Leshiy: fixed

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -1,23 +1,78 @@
...
+        self.evm_to_jor.lookup(&evm_id).cloned()
+    }
+
+    fn del_accounts(&mut self, jor_id: JorAddress) {
```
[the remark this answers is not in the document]
Mr-Leshiy: fixed

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -1,23 +1,78 @@
...
+    }
+}
 
 #[derive(Clone, PartialEq, Eq)]
 pub struct Ledger {
-    pub(crate) accounts: AccountTrie,
     pub(crate) logs: LogsState,
     pub(crate) environment: Environment,
     pub(crate) current_epoch: BlockEpoch,
+    pub(crate) adress_mapping: AddressMapping,
```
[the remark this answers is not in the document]
Mr-Leshiy: fixed

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -326,9 +328,11 @@ pub enum Error {
     MintingPolicyViolation(#[from] MintingPolicyViolation),
     #[error("evm transactions are disabled, the node was built without the 'evm' feature")]
     DisabledEvmTransactions,
+    // #[error("Transaction for EvmMapping is invalid. expecting 1 input, 1 witness and 0 output")]
+    // EvmMappingInvalidTransaction,
```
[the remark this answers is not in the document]
Mr-Leshiy: fixed

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -1,22 +1,68 @@
...
+use imhamt::Hamt;
+use std::collections::hash_map::DefaultHasher;
+use thiserror::Error;
+
+#[derive(Debug, Error, Clone, PartialEq, Eq)]
+pub enum Error {
+    #[error(
+        "for the provided jormungandr account: {} or evm account: {} mapping is already exist", .0.to_string(), .1.to_string()
+    )]
+    ExistedMapping(JorAddress, EvmAddress),
```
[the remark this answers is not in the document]
Mr-Leshiy: I am following to the spec, so this behavior has been documented. Of course later we can update it and override existing mapping with the new data, but at least it does not seems that we essentially need this functionality at this moment, so we are following the simplest approach.

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -31,19 +86,59 @@ impl EvmState for super::Ledger {
...
+            .unwrap_or_else(|_| JorAccount::new(Value::zero(), ()));
+
+        match f(EvmAccount {
+            balance: account.value.0.into(),
+            state: account.evm_state,
+        }) {
+            Some(account) => {
+                self.accounts = self
+                    .accounts
+                    .evm_update(
```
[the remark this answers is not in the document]
Mr-Leshiy: Yes, you are right, it is not used in this PR, but such option will be used later with the evm contract mapping implementation.

chain-impl-mockchain/src/accounting/account/account_state.rs
```diff
@@ -84,6 +84,8 @@ pub struct AccountState<Extra> {
     pub value: Value,
     pub tokens: Hamt<DefaultHasher, TokenIdentifier, Value>,
     pub last_rewards: LastRewards,
+    #[cfg(feature = "evm")]
+    pub evm_state: chain_evm::state::AccountState,
     pub extra: Extra,
```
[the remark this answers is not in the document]
Mr-Leshiy: It is also an option, but I would prefer to add a new field because it does not leads to the huge update related for replacing in all parts usage of the `Account<chain_evm::state::AccountState>` with the `Account<()>` as it is used currently.
And generally I want to suggest to remove this optional `extra` field, as it is not used and defined as `()` type.

chain-impl-mockchain/src/ledger/evm.rs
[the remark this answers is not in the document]
Zeegomo: Maybe it could be useful to add two different methods so that invocations like this one are cleaner?

chain-impl-mockchain/src/accounting/account/account_state.rs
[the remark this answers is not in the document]
Mr-Leshiy: > Would it be easier to treat all those states as the Extra instead of adding an hardcoded field for each one?

I am not sure that it will be easier, because still we will have a separate struct with which we will define this `Extra` generic type. But again we will be need to correctly update AccountState typing in all places where it is initialized, that is does not look like as an easy update.
Instead of adding just a new field and modify a few places.

chain-impl-mockchain/src/ledger/evm.rs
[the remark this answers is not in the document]
Mr-Leshiy: fixed

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -31,19 +86,59 @@ impl EvmState for super::Ledger {
...
-        F: FnOnce(Account) -> Option<Account>,
+        F: FnOnce(EvmAccount) -> Option<EvmAccount>,
     {
-        self.evm.accounts = self.evm.accounts.clone().modify_account(address, f);
+        let address = self.evm.adress_mapping.jor_address(address).ok_or(format!(
+            "Can not find corresponding jormungadr account for the evm account: {}",
+            address
+        ))?;
+
+        let account = self
```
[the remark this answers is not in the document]
Mr-Leshiy: It is no appropriate to use `self.account(address)` at this place because this function returns `EvmAccount` instead of `JorAccount` which is needed at this place.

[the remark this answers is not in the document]
Zeegomo: But ultimately you need the `EvmAccount` in line 121, don't you?

chain-impl-mockchain/src/accounting/account/account_state.rs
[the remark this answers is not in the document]
Zeegomo: > AccountState typing in all places where it is initialized

I would expect that to be just a type definition in one place and then compiler inference all the way down, but I may be wrong. 
To be clear, I don't think adding a new field is a wrong solution, I'm just a bit worried of bloating the struct with many fields, and I'm wondering why do we even have that extra field if we don't use it.

chain-impl-mockchain/src/ledger/evm.rs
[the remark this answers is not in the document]
Mr-Leshiy: >Yes, you are right, it is not used in this PR, but such option will be used later with the evm contract mapping implementation.

Sorry, was mistaken and confused you. We probably can create a new account because of 
```
        let account = self
            .accounts
            .get_state(&address)
            .cloned()
            .unwrap_or_else(|_| JorAccount::new(Value::zero(), ()));
```

chain-impl-mockchain/src/ledger/evm.rs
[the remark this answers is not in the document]
Mr-Leshiy: Yes, you are right, have missed that part. But the most crucial part is to treat an error in this place.
```
        let address = self
            .evm
            .address_mapping
            .jor_address(&address)
            .ok_or_else(|| {
                ExitError::Other(
                    format!(
                        "Can not find corresponding jormungadr account for the evm account: {}",
                        address
                    )
                    .into(),
                )
            })?
            .clone();

```
And it is something different that we want inside the `fn account(&self, evm_address: &EvmAddress)`

chain-impl-mockchain/src/accounting/account/account_state.rs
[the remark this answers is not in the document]
Mr-Leshiy: Actually I am not sure, what was an intention of the `extra`, so I am also wondering what is the purpose of this field :)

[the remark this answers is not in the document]
Mr-Leshiy: I am also wondering what is the purpose of the `extra` field for the `AccountState` 😊

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -1,23 +1,78 @@
...
+        if let Some(evm_id) = self.jor_to_evm.lookup(&jor_id) {
+            self.evm_to_jor = self.evm_to_jor.remove(evm_id).unwrap();
+            self.jor_to_evm = self.jor_to_evm.remove(&jor_id).unwrap();
+        }
+    }
+
+    fn map_accounts(&mut self, jor_id: JorAddress, evm_id: EvmAddress) -> Result<(), Error> {
+        (!self.evm_to_jor.contains_key(&evm_id) && !self.jor_to_evm.contains_key(&jor_id))
+            .then(|| ())
+            .ok_or_else(|| Error::ExistedMapping(jor_id.clone(), evm_id))?;
```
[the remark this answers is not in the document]
Mr-Leshiy: fixed

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -1,23 +1,78 @@
...
+
+        self.evm_to_jor = self.evm_to_jor.insert(evm_id, jor_id.clone()).unwrap();
+        self.jor_to_evm = self.jor_to_evm.insert(jor_id, evm_id).unwrap();
```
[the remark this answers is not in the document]
Mr-Leshiy: fixed

chain-evm/src/machine.rs
```diff
@@ -467,25 +486,30 @@ impl<'a, State: EvmState> StackState<'a> for VirtualMachine<'a, State> {
...
-            });
+            self.state
+                .modify_account(*address, |_| {
+                    if self.delete_empty && account.is_empty() {
+                        None
+                    } else {
+                        Some(account)
+                    }
+                })
+                .map_err(|e| e)?;
```
Zeegomo, line 505: this is a no-op

chain-evm/src/machine.rs
```diff
@@ -467,25 +486,30 @@ impl<'a, State: EvmState> StackState<'a> for VirtualMachine<'a, State> {
...
         }
 
         // delete account
         for address in self.substate.deletes.iter() {
-            self.state.modify_account(*address, |_| None);
+            self.state
+                .modify_account(*address, |_| None)
+                .map_err(|e| e)?;
```
Zeegomo, line 512: no-op

chain-evm/src/machine.rs
```diff
@@ -568,19 +592,27 @@ impl<'a, State: EvmState> StackState<'a> for VirtualMachine<'a, State> {
...
+        self.substate.account_mut(address, self.state).state.code = code;
     }
 
     fn transfer(&mut self, transfer: Transfer) -> Result<(), ExitError> {
         let source = self.substate.account_mut(transfer.source, self.state);
 
-        source.balance = match source.balance.checked_sub(transfer.value) {
+        source.balance = match source
+            .balance
+            .checked_sub(transfer.value.try_into().map_err(|_| ExitError::OutOfGas)?)
```
Zeegomo, line 603: Can the failed `try_into()` be caused by value being bigger than what a balance can hold? In that case I think it's better to have a more descriptive error message, since I'm sure this can happen in a lot of other places as well

chain-evm/src/machine.rs
```diff
@@ -568,19 +592,27 @@ impl<'a, State: EvmState> StackState<'a> for VirtualMachine<'a, State> {
...
             None => return Err(ExitError::OutOfGas),
         };
 
         let target = self.substate.account_mut(transfer.target, self.state);
-        target.balance = match target.balance.checked_add(transfer.value) {
+        target.balance = match target.balance.checked_add(
+            transfer
+                .value
+                .try_into()
+                .map_err(|_| ExitError::Other("Balance overflow".into()))?,
```
Zeegomo, line 614: Please be consistent with error codes if the cause is the same, we can just bubble up the conversion error from `Balance` in this case

chain-evm/src/state/account.rs
```diff
@@ -21,44 +21,48 @@ impl TryFrom<U256> for Balance {
...
     }
-    /// Checked addition of `U256` types. Returns `Some(balance)` or `None` if overflow
-    /// occurred.
-    pub fn checked_add(self, other: U256) -> Option<Balance> {
-        match self.0.checked_add(other) {
-            Some(U256([v, 0, 0, 0])) => Some(Balance::from(v)),
-            _ => None,
+    /// Returns `Some(balance)` or `None` if overflow occurred.
+    pub fn checked_add(self, other: Balance) -> Option<Balance> {
+        if let Some(res) = self.0.checked_add(other.0) {
```
Zeegomo, line 55: nit: `self.0.checked_add(other.0).map(Self)`

chain-evm/src/state/account.rs
```diff
@@ -21,44 +21,48 @@ impl TryFrom<U256> for Balance {
...
     }
-    /// Checked substraction of `U256` types. Returns `Some(balance)` or `None` if overflow
-    /// occurred.
-    pub fn checked_sub(self, other: U256) -> Option<Balance> {
-        match self.0.checked_sub(other) {
-            Some(U256([v, 0, 0, 0])) => Some(Balance::from(v)),
-            _ => None,
+    /// Returns `Some(balance)` or `None` if overflow occurred.
+    pub fn checked_sub(self, other: Balance) -> Option<Balance> {
+        if let Some(res) = self.0.checked_sub(other.0) {
```
Zeegomo, line 62: nit: `self.0.checked_sub(other.0).map(Self)`

chain-evm/src/state/account.rs
```diff
@@ -68,19 +72,28 @@ pub type ByteCode = Vec<u8>;
 /// A represantation of an EVM account.
 #[derive(Clone, Default, PartialEq, Eq)]
 pub struct Account {
-    /// Account nonce. A number of value transfers from this account.
-    pub nonce: Nonce,
```
Zeegomo, line 72: Not that it's crucial, but I would like to understand the reasoning behind nonce and balance being split in different structs, is it because of the shared balance stuff?

chain-evm/src/state/account.rs
```diff
@@ -68,19 +72,28 @@ pub type ByteCode = Vec<u8>;
...
     /// Account balance.
     pub balance: Balance,
+    /// Account state.
+    pub state: AccountState,
+}
+
+/// A represantation of an EVM account state.
```
Zeegomo, line 81: typo: representation

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -1,23 +1,84 @@
...
+use imhamt::Hamt;
+use std::collections::hash_map::DefaultHasher;
+use thiserror::Error;
+
+#[derive(Debug, Error, Clone, PartialEq, Eq)]
+pub enum Error {
+    #[error(
+        "for the provided jormungandr account: {} or evm account: {} mapping is already exist", .0.to_string(), .1.to_string()
+    )]
+    ExistedMapping(JorAddress, EvmAddress),
```
Zeegomo, line 28: typo: `ExistingMapping`

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -31,19 +92,68 @@ impl EvmState for super::Ledger {
...
-        F: FnOnce(Account) -> Option<Account>,
+        F: FnOnce(EvmAccount) -> Option<EvmAccount>,
     {
-        self.evm.accounts = self.evm.accounts.clone().modify_account(address, f);
+        let address = self
+            .evm
+            .address_mapping
+            .jor_address(&address)
+            .ok_or_else(|| {
+                ExitError::Other(
```
Zeegomo, line 121: we can provide a better typed error here

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -31,19 +92,68 @@ impl EvmState for super::Ledger {
...
+                    format!(
+                        "Can not find corresponding jormungadr account for the evm account: {}",
+                        address
+                    )
+                    .into(),
+                )
+            })?
+            .clone();
```
Zeegomo, line 129: is this clone needed?

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -52,12 +162,24 @@ impl EvmState for super::Ledger {
...
+            return Err(Error::EvmMappingSignatureFailed);
+        }
+
+        self.evm
+            .address_mapping
+            .map_accounts(tx.account_id().clone(), *tx.evm_address())?;
+        Ok(self)
+    }
+
+    pub fn run_transaction(mut self, contract: EvmTransaction) -> Result<Self, Error> {
```
Zeegomo, line 181: should this at the end remove funds from the `JorAccount` if the balance is shared?

Zeegomo (changes requested):

chain-evm/src/state/account.rs
Mr-Leshiy: fixed

chain-evm/src/state/account.rs
Mr-Leshiy: fixed

chain-evm/src/state/account.rs
Mr-Leshiy: fixed

chain-evm/src/state/account.rs
Mr-Leshiy: Yes, you are correct, `AccountState` also used for the jormungandr `AccountState<Extra>` as a field

chain-evm/src/machine.rs
Mr-Leshiy: fixed

chain-evm/src/machine.rs
Mr-Leshiy: fixed

chain-impl-mockchain/src/ledger/evm.rs
Mr-Leshiy: fixed

Mr-Leshiy: Yes it is needed to avoid borrowing of the `self.evm.address_mapping` though because of the later usage in `self.evm.address_mapping.del_accounts(&address)`

chain-impl-mockchain/src/ledger/evm.rs
Mr-Leshiy: All accounts mutation is appeared inside `modify_account` directly

chain-impl-mockchain/src/ledger/evm.rs
Mr-Leshiy: fixed

chain-evm/src/machine.rs
Mr-Leshiy: Yes, it is possible, we can throw something similar with `ExitError::Other("Balance overflow".into()))`, but the purpose of throwing `ExitError::OutOfGas` it is if value which we want to spend is overflowed it is obviously also means that you are trying to spend more than you have ``ExitError::OutOfGas` error in this case. 
So IMO it is not a huge difference between those two erros types, but I agree `ExitError::Other("Balance overflow".into()))` reflects the core problem, so lets throw such error.

chain-impl-mockchain/src/ledger/evm.rs
Mr-Leshiy: Actually no, `ExitError` it is a `evm-library` error type, so it is the only way to provide a custom error message.
Or do you mean something else ?

Zeegomo: Ah sorry, forget this then

chain-evm/src/machine.rs
Zeegomo: >  it is if value which we want to spend is overflowed it is obviously also means that you are trying to spend more than you have

Then wouldn't `OutOfFund` be more appropriate? `OutOfGas` has a different meaning

PR #776, opened 2022-03-21
Add StackState implementation for the VirtualMachine

Mr-Leshiy:
Add a `StackState` trait implementation for the `VirtualMachine` struct which empowers better error handling and future performance  improvements of the vm state

### DIFF
--- a/chain-evm/src/machine.rs
+++ b/chain-evm/src/machine.rs
@@ -254,11 +254,6 @@ where
             // apply changes to the state, this consumes the executor
             let vm = executor.into_state();
             println!("pre gas: {}", vm.substate.metadata.gasometer().gas());
-            // // Next, we consume the stack state and extract the values and logs
-            // // used to modify the accounts trie in the VirtualMachine.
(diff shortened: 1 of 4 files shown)


saibatizoku (approved):
LGTM, it's great to see balances throwing overflow errors at runtime! This implementation makes better abstractions. Fantastic!

Only thing missing from my perspective is to use proper machine errors instead of text messages, I think.

Mr-Leshiy:
> Only thing missing from my perspective is to use proper machine errors instead of text messages, I think.
Can you a little bit elaborate in which places do we need it ?
Because we have the following errors:
```
pub enum Error {
    #[error("transaction error: machine returned a normal EVM error")]
    TransactionError(ExitError),
    #[error("transaction fatal error: machine encountered an error that is not supposed to be normal EVM errors, such as requiring too much memory to execute")]
    TransactionFatalError(ExitFatal),
    #[error("transaction has been reverted: machine encountered an explict revert")]
    TransactionRevertError(ExitRevert),
}
```

PR #775, opened 2022-03-16
begin migration from quickcheck -> proptest in `chain-impl-mockchain`

cameron1024:
This PR adds proptest implementations to a few key types, and also a few tests.
A later PR will add implementations for EVM types, I tried adding them into this PR but it got quite big quite quickly, so trying to leave that to its own PR

PR #779, opened 2022-03-24
More proptest migrations

cameron1024:
This PR changes a few more tests from quickcheck -> proptest, specifically avoiding any which would benefit from the following PRs:
https://github.com/input-output-hk/proptest/pull/4
https://github.com/input-output-hk/proptest/pull/3

Depends on: https://github.com/input-output-hk/chain-libs/pull/775


Mr-Leshiy:
Also need to make a corresponding update inside https://github.com/input-output-hk/jormungandr repo if it is needed.

cameron1024:
I think it shouldn't affect jormungandr? This is just changing some tests and adding `proptest` impls, but we're keeping the old `quickcheck` impls. I'm going to remove those impls in a later PR

Mr-Leshiy:
@cameron1024  I am not sure, but it is possible that jormungandr tests can rely on the `quickcheck` implementation for some types currently. That is why it is better to firstly check that jormungandr will work fine with this update and then merge this PR.

PR #781, opened 2022-03-25
Add new SerializedSize trait 

Mr-Leshiy:
- Add new `SerializedSize` trait, and implement it for all types which implements  `Serialization` trait. This trait allows to cheaply determine size of the serialized object avoiding serialization procedure.
- Modify the usage of the Fragment::serialized_size() function
- Remove redundant `HeaderRaw` struct, and update `Header` serde implementation.


chain-addr/src/lib.rs
```diff
@@ -418,6 +416,21 @@ impl std::str::FromStr for AddressReadable {
     }
 }
 
+impl SerializedSize for Address {
+    fn serialized_size(&self) -> usize {
+        0_u8.serialized_size()
```
enzoc4, line 421: this looks a bit weird, tbh... I know it makes no difference, but maybe `self.to_kind_value().serialized_size()` reads a bit better.

chain-impl-mockchain/src/ledger/recovery.rs
```diff
@@ -1120,14 +1120,19 @@ fn unpack_entries(codec: &mut Codec<&[u8]>) -> Result<Vec<EntryOwned>, ReadError
     Ok(res)
 }
 
+impl SerializedSize for Ledger {
+    fn serialized_size(&self) -> usize {
+        unimplemented!()
```
enzoc4, line 1125: what's the problem with this?

chain-ser/src/deser.rs
```diff
@@ -84,3 +94,75 @@ impl<const N: usize> Deserialize for [u8; N] {
...
+    fn serialized_size(&self) -> usize {
+        std::mem::size_of::<u64>()
+    }
+}
+
+impl SerializedSize for u128 {
+    fn serialized_size(&self) -> usize {
+        std::mem::size_of::<u128>()
+    }
+}
```
enzoc4, line 126: I wonder if could simplify these by implementing

`impl<T: Copy> SerializedSize for T`

but maybe it is a bit hacky?

chain-ser/src/deser.rs
```diff
@@ -26,8 +26,18 @@ pub enum WriteError {
...
+
+impl<T: SerializedSize> SerializedSize for &T {
+    fn serialized_size(&self) -> usize {
+        (*self).serialized_size()
+    }
+}
+
 /// Define that an object can be written to an `std::io::Write` object.
-pub trait Serialize {
+pub trait Serialize: SerializedSize {
```
enzoc4, line 40: I'm not sure about this bound, is it really necessary? We could just request for `Serialize + SerializedSize` at some point if we need both methods.

enzoc4 (commented):
There are a few things here that make me wonder if we shouldn't invest some time in writing a derive macro for this, but I never did it before so I don't know how complicated that would be. I think it could be less error prone, though.

Also, I imagine you already considered this, but why not just have:

```rust
pub trait Serialize {
    fn serialize<W: std::io::Write>(&self, codec: &mut Codec<W>) -> Result<(), WriteError>;

    fn serialize_as_vec(&self) -> Result<Vec<u8>, WriteError> {
        let mut data = Vec::new();
        self.serialize(&mut Codec::new(&mut data))?;
        Ok(data)
    }

    fn serialized_size(&self) -> usize;
}
```

Especially considering that if we require `Serialize: SerializedSize` it's more or less the same.

Mr-Leshiy:
@enzoc4 , I have avoided such implementation, because for the primitives types as `u8`, `u16` etc. we shouldn't have a serialization operations.
```
pub trait Serialize {
    fn serialize<W: std::io::Write>(&self, codec: &mut Codec<W>) -> Result<(), WriteError>;

    fn serialize_as_vec(&self) -> Result<Vec<u8>, WriteError> {
        let mut data = Vec::new();
        self.serialize(&mut Codec::new(&mut data))?;
        Ok(data)
    }

    fn serialized_size(&self) -> usize;
}
```

chain-ser/src/deser.rs
Mr-Leshiy: Unfortunately it is not possible and will not correct, because for any type which will implement `Copy` trait will have `std::mem::size_of` implementation which is not correct.

_The types *const T, &T, Box<T>, Option<&T>, and Option<Box<T>> all have the same size. If T is Sized, all of those types have the same size as usize._

chain-ser/src/deser.rs
Mr-Leshiy: Actually not it is not necessary so can be update and therefore `impl SerializedSize for Ledger {` can be removed.

chain-impl-mockchain/src/ledger/recovery.rs
Mr-Leshiy: I have not added implementation for the `Ledger` type at this moment because it needs a huge effort to implement it for most of dependent types. Thus it is not actually used for the `Ledger` type I left it `unimplemented!()`.

chain-addr/src/lib.rs
Mr-Leshiy: fixed

chain-ser/src/deser.rs
enzoc4: Oh right, I forgot about reference types, good point. Probably something like `zerocopy::AsBytes` would be a better fit, but that would require a dependency, so yeah, ignore this.

enzoc4:
> because for the primitives types as u8, u16 etc. we shouldn't have a serialization operations.

I'm not sure I see the reason for that, since they are indeed serializable. The implementations can be just a single `codec.put_*`. 

What's the disadvantage? You can always decide to cast the type, or just not use the trait's function if you need more flexibility for a specific field. Besides, I find it weird to have a type that says it has a `SerializedSize` but can't be serialized.

Mr-Leshiy:
@enzoc4 we have a two possible options how two serde these primitive types: little endian encoding or big endian encoding, so it leads to confusing in the default implementation. Or if we will even choose what kind of default implementation we will have little endian or big endian, it can leads for confusing in the Serialize trait usage in my opinion.
But generally I don't have a strong position about it.
What do you think ?

enzoc4:
>@enzoc4 we have a two possible options how to implement serde these primitive types: little endian encoding or big endian encoding, so it leads to confusing in the default implementation. Or if we will even choose what kind of default implementation we will have little endian or big endian, it can leads for confusing in the Serialize trait usage in my opinion.
But generally I don't have a strong position about it.
What do you think ?

Yeah, that's a good point, I don't think a default implementation is a good idea. OTOH we _do_ implement `Serialize` for things like `Value`, which is not really that different than a `u64`.

I'm still not too convinced that having things `SerializedSize` but not `Serialize` makes too much sense, because the things are pretty much coupled, but I'm not just how to unify them neither... To be honest I think I would rather just write `std::mem::size_of_val` or `std::mem::size_of`  at this point. Or maybe even have something like this:

```rust
pub trait StaticSerializedSize: Sized {
    const SERIALIZED_SIZE: usize = std::mem::size_of::<Self>();

    fn serialized_size(&self) -> usize {
        Self::SERIALIZED_SIZE
    }
}
```

But to be fair, I'm not sure that provides any advantages, and it may be bikeshedding.

Mr-Leshiy:
@enzoc4  I would prefer to avoid the usage of the `std::mem::size_of_val` or `std::mem::size_of` functions in this case, because it is not really reflect that meaning that we need. We need to get a serialized size of the provided object, not a size of the struct or type (for example for the enum types std::mem::size_of_val, std::mem::size_of will return a maximum sized enum option + 4 bytes for the alignment)
```
enum B {
    V1(u32),
    V2(u32, u32),
}

fn main() {
    let b1 = B::V1(0);
    let b2 = B::V2(0, 0);
    println!("Size of B: {}", std::mem::size_of::<B>());
    println!("Size of val B: {}", std::mem::size_of_val(&b1));
    println!("Size of val B: {}", std::mem::size_of_val(&b2));
}
```
Returns
```
Size of B: 12
Size of val B: 12
Size of val B: 12
```

chain-ser/src/deser.rs
```diff
@@ -84,3 +94,75 @@ impl<const N: usize> Deserialize for [u8; N] {
...
+    }
+}
+
+impl SerializedSize for u64 {
+    fn serialized_size(&self) -> usize {
+        std::mem::size_of::<u64>()
+    }
+}
+
+impl SerializedSize for u128 {
```
Zeegomo, line 122: How can you define the serialized size without knowing the `Serialize` implementation or the format used? Big-endian or little-endian are not the only possible formats, what if the implementor decide to use something like https://docs.rs/bincode/latest/bincode/config/struct.VarintEncoding.html?

chain-impl-mockchain/src/header/header.rs
```diff
@@ -298,20 +298,25 @@ use chain_core::{
...
+    fn serialized_size(&self) -> usize {
+        0_u16.serialized_size() + self.as_slice().serialized_size()
+    }
+}
+
 impl Serialize for Header {
     fn serialize<W: std::io::Write>(&self, codec: &mut Codec<W>) -> Result<(), WriteError> {
-        codec.put_bytes(self.as_slice())
+        let bytes = self.as_slice();
+        codec.put_be_u16(bytes.len() as u16)?;
```
Zeegomo, line 310: I think `Header` is fixed size, no need to add length here. If deserializing without knowing the version (i.e. Praos, BFT, ...) is infeasible we could use a tag to define which `BlockVersion` this is

chain-ser/src/deser.rs
Zeegomo: we could use macros though

Zeegomo (commented):
IMO, I'm leaning towards what Enzo suggested and have `serialized_size` as a method of `Serialize`. `serialize_size` is entirely dependent on the format used, and it does not make sense for me to be able to define it without having a `Serialize` implementation laid out (see my comment about integer serialization).

I know it kind of defeats the purpose of a fast computation, but this could very well be an optional method that just calls `serialize_as_vec().len()` as the default behavior

Mr-Leshiy:
@enzoc4 , @Zeegomo , Thanks for all your suggestions and comments, very appreciated !
I have updated with all that parts that you have mentioned, hope currently it fits with that you have meant.

chain-impl-mockchain/src/certificate/delegation.rs
```diff
@@ -56,10 +56,16 @@ impl StakeDelegation {
 }
 
 impl Serialize for OwnerStakeDelegation {
+    fn serialized_size(&self) -> usize {
+        let delegation_buf =
+            serialize_delegation_type(&self.delegation, ByteBuilder::new()).finalize_as_vec();
+        delegation_buf.len()
```
enzoc4, line 62: I think this is basically the same thing the default implementation does, is it not? Not too important though.

chain-impl-mockchain/src/header/version.rs
```diff
@@ -151,7 +151,7 @@ mod tests {
 
     #[quickcheck]
     pub fn conversion_u16(block_version: AnyBlockVersion) -> TestResult {
```
enzoc4, line 153:
```suggestion
    pub fn conversion_u8(block_version: AnyBlockVersion) -> TestResult {
```

enzoc4 (approved):

chain-addr/src/lib.rs
```diff
@@ -419,6 +417,17 @@ impl std::str::FromStr for AddressReadable {
 }
 
 impl Serialize for Address {
+    fn serialized_size(&self) -> usize {
+        std::mem::size_of::<u8>()
```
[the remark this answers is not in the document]
Mr-Leshiy: fixed

chain-impl-mockchain/src/header/version.rs
Mr-Leshiy: fixed

chain-impl-mockchain/src/certificate/delegation.rs
Mr-Leshiy: Generally yes, you agree, but I would prefer to left such implementation only for one purpose - to have a reminder that if we will change the `serialize` implementation, to also update `serialized_size` impl.

PR #783, opened 2022-03-29
Remove primitive_types dependency

Mr-Leshiy:


saibatizoku (approved):
Bye bye primitive-types, see you as aliases.

PR #785, opened 2022-03-31
Add tests to leadership module

nicopado:


chain-impl-mockchain/src/leadership/genesis/mod.rs
```diff
@@ -295,6 +295,49 @@ mod tests {
...
+        let cb = ConfigBuilder::new()
+            .with_slots_per_epoch(leader_election_parameters.slots_per_epoch)
+            .with_active_slots_coeff(leader_election_parameters.active_slots_coeff_as_milli());
+
+        let mut ledger = LedgerBuilder::from_config(cb)
+            .build()
+            .expect("cannot build test ledger")
+            .ledger;
+
+        let mut pools = HashMap::<PoolId, (SecretKey<RistrettoGroup2HashDh>, u64, Stake)>::new();
```
dkijania, line 312: i think you don't need to specify full hashMap signature. it is sufficient to just:
```
let mu pools = HashMap::new()
```

chain-impl-mockchain/src/leadership/genesis/mod.rs
```diff
@@ -295,6 +295,49 @@ mod tests {
...
+
+        for _i in 0..leader_election_parameters.pools_count {
```
dkijania, line 314: nitpick:

this can be written as generator:

```
let mut pools: HashMap<_>  = std::iter::from_fn(|| {
      let (pool_id, pool_vrf_private_key) = make_pool(&mut ledger);
      (
                pool_id.clone(),
                (pool_vrf_private_key, 0, leader_election_parameters.value),
            )
 }).take(leader_election_parameters.pools_count).collect();
```

and you dont need line 312 i guess

chain-impl-mockchain/src/leadership/genesis/mod.rs
```diff
@@ -295,6 +295,49 @@ mod tests {
...
+        }
+
+        let date = ledger.date();
+
+        let selection = make_leadership_with_pools(&ledger, &pools);
+
+        let (invalid_pool_id, pool_vrf_private_key) = make_pool(&mut ledger);
+
+        for pool in pools{
+            let leader = selection.leader(&pool.0, &pool.1.0, date);
```
dkijania, line 329: leader is not used. maybe replace this line to:

```
            let _ = selection.leader(&pool.0, &pool.1.0, date);
```

and this entire loop can be written simpler :

```
 pools.iter().for_each(|pool| selection.leader(&pool.0, &pool.1.0, date));
```

chain-impl-mockchain/src/leadership/genesis/mod.rs
```diff
@@ -295,6 +295,49 @@ mod tests {
...
+        }
+
+        // Act
+        let invalid_leader = selection.leader(&invalid_pool_id, &pool_vrf_private_key, date);
+
+        println!("Invalid Leader: {:?}", invalid_leader);
+
+        // Assert
+        assert!(true);
```
dkijania, line 338: i guess this assertion is only tbd ?

chain-impl-mockchain/src/leadership/genesis/mod.rs
```diff
@@ -295,6 +295,49 @@ mod tests {
...
+        let selection = make_leadership_with_pools(&ledger, &pools);
+
+        let (invalid_pool_id, pool_vrf_private_key) = make_pool(&mut ledger);
+
+        for pool in pools{
+            let leader = selection.leader(&pool.0, &pool.1.0, date);
+        }
+
+        // Act
+        let invalid_leader = selection.leader(&invalid_pool_id, &pool_vrf_private_key, date);
```
dkijania, line 333: if i remember correctly method should return Option<_>. Maybe just put assertion:
```
 assert!(selection.leader(&invalid_pool_id, &pool_vrf_private_key, date).in_none())
```

chain-impl-mockchain/src/leadership/genesis/mod.rs
dkijania: or if you just want to assert that for correct leader selection will return something. you could write:

```
assert!(pools.iter().all(|pool| selection.leader(&pool.0, &pool.1.0, date).is_some()));
```

chain-impl-mockchain/src/leadership/genesis/mod.rs
```diff
@@ -295,6 +295,49 @@ mod tests {
...
+
+        for _i in 0..leader_election_parameters.pools_count {
+            let (pool_id, pool_vrf_private_key) = make_pool(&mut ledger);
+            pools.insert(
+                pool_id.clone(),
+                (pool_vrf_private_key, 0, leader_election_parameters.value),
+            );
+        }
+
+        let date = ledger.date();
```
dkijania, line 322: this is only my opinion but i think date variable should be daclared as late as possible. 
 https://stackoverflow.com/questions/10204828/declare-local-variables-as-late-as-possible-or-at-the-nearest-curly-brace-they-b

dkijania (changes requested):

chain-impl-mockchain/src/leadership/genesis/mod.rs
```diff
@@ -295,6 +295,64 @@ mod tests {
...
+            for (pool_id, (pool_vrf_private_key, _, _)) in pools.iter_mut(){
+                
+                let leader = selection.leader(pool_id, pool_vrf_private_key, date);
+                match leader.unwrap()
+                {
+                    None => {}
+                    Some(_) => {
+                        println!("Leader Slot #{:?}, date: {:?}", _i, date);
+                        
+                        let invalid_leader = selection.leader(&invalid_pool_id, &invalid_pool_vrf_private_key, date);
```
Zeegomo, line 339: Why is this run only on slots for which other some other pool is selected?

chain-impl-mockchain/src/leadership/genesis/mod.rs
```diff
@@ -536,6 +594,33 @@ mod tests {
     use crate::fragment::Contents;
     use crate::header::{BlockVersion, HeaderBuilderNew};
 
+    #[test]
+    pub fn leadership_verify_wrong_proof() {
```
Zeegomo, line 598: could you be more specific on what is the property we're testing?

chain-impl-mockchain/src/leadership/genesis/mod.rs
```diff
@@ -295,6 +295,64 @@ mod tests {
...
+        // Act
+        let mut date = ledger.date();
+        let mut flag = true;
+
+        for _i in 00..leader_election_parameters.slots_per_epoch{
+            println!("Slot #{:?}, date: {:?}", _i, date);
+            for (pool_id, (pool_vrf_private_key, _, _)) in pools.iter_mut(){
+                
+                let leader = selection.leader(pool_id, pool_vrf_private_key, date);
+                match leader.unwrap()
```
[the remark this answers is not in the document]
nicopado: The logic will be revisited almost surely :)
The reason behind this is that this method will return the same result (Ok(None)) for both a valid and an invalid pool in most of the slots, and I needed a way to be sure that I wasn't getting a false positive

chain-impl-mockchain/src/leadership/genesis/mod.rs
nicopado: that is most a placeholder name for now as I'm still figuring it out :)
that test is meant to give coverage to line 149, in which the verify method is throwing a verification error

chain-impl-mockchain/src/leadership/genesis/mod.rs
Zeegomo: The two events are not related. Whether some pool is selected for a slot does not influence other pools. An invalid pool should get `None` all of the times, the others *might* be selected with a certain probability deriving from their stakes and some chain parameters.

chain-impl-mockchain/src/leadership/genesis/mod.rs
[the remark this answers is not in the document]
Zeegomo: See my comment below, those two events are independent

chain-impl-mockchain/src/leadership/genesis/mod.rs
nicopado: I see your point now, thanks for the clarification! I removed the check on the valid pools and checked only that the invalid one was never returning the unexpected value within one epoch.
Let me know if you see further improvements :)

chain-impl-mockchain/src/leadership/genesis/mod.rs
```diff
@@ -295,6 +295,49 @@ mod tests {
...
+            ))
+        }).take(leader_election_parameters.pools_count).collect();
+
+        let selection = make_leadership_with_pools(&ledger, &pools);
+
+        let (invalid_pool_id, invalid_pool_vrf_private_key) = make_pool(&mut ledger);
+
+        // Act
+        let mut date = ledger.date();
+        let mut flag = true;
```
dkijania, line 326: I don't particularly like this approach with flag. maybe we could have a block like this:

```
let invalid_leader_elected_for_epoch = {
      for _ in 00..leader_election_parameters.slots_per_epoch{
            let invalid_leader = selection.leader(&invalid_pool_id, &invalid_pool_vrf_private_key, date).unwrap();
             if invalid_leader.is_some() {
                   return true;
            }
            date = date.next(ledger.era());
        }
        return false;
 }
 
 assert!(!invalid_leader_elected_for_epoch)
```

dkijania: In above it's clear what we are checking

dkijania (changes requested):

chain-impl-mockchain/src/leadership/genesis/mod.rs
Zeegomo: asserting on the return value of the `leader` function directly would be even better

dkijania (approved):

PR #786, opened 2022-03-31
RLP encoding for EVM transactions

saibatizoku:
Replacing codec from chain-libs with RLP encoding from Ethereum for EVM Transactions


chain-impl-mockchain/src/evm/mod.rs
```diff
@@ -48,66 +47,132 @@ pub enum EvmTransaction {
...
+        }
+    }
+}
+
+impl EvmTransaction {
+    /// Serialize the contract into a `ByteBuilder`.
+    pub fn serialize_in(&self, _bb: ByteBuilder<Self>) -> ByteBuilder<Self> {
+        #[cfg(feature = "evm")]
+        {
+            _bb.bytes(&self.rlp_bytes())
```
enzoc4, line 171:
```suggestion
            let bytes = &self.rlp_bytes();
            _bb.u64(bytes.len() as u64).bytes(&bytes)
```

chain-impl-mockchain/src/evm/mod.rs
```diff
@@ -126,152 +191,39 @@ pub fn serialize_u256<T>(bb: ByteBuilder<T>, value: &U256) -> ByteBuilder<T> {
...
-#[cfg(feature = "evm")]
-fn read_bytecode(codec: &mut Codec<&[u8]>) -> Result<ByteCode, ReadError> {
-    match codec.get_be_u64()? {
-        n if n > 0 => Ok(ByteCode::from(codec.get_slice(n.try_into().unwrap())?)),
-        _ => Ok(ByteCode::default()),
+impl Serialize for EvmTransaction {
+    fn serialize<W: std::io::Write>(&self, _codec: &mut Codec<W>) -> Result<(), WriteError> {
+        #[cfg(feature = "evm")]
+        {
+            _codec.put_bytes(self.rlp_bytes().as_ref())?;
```
enzoc4, line 203:
```suggestion
            let bytes = self.rlp_bytes();
            _codec.put_be_u64(bytes.len() as u64)?;
            _codec.put_bytes(&bytes)?;
```

chain-impl-mockchain/src/evm/mod.rs
```diff
@@ -126,152 +191,39 @@ pub fn serialize_u256<T>(bb: ByteBuilder<T>, value: &U256) -> ByteBuilder<T> {
...
-                    access_list,
-                })
-            }
-            n => Err(ReadError::UnknownTag(n as u32)),
+impl Deserialize for EvmTransaction {
+    fn deserialize<R: std::io::Read>(_codec: &mut Codec<R>) -> Result<Self, ReadError> {
+        #[cfg(feature = "evm")]
+        {
+            let mut rlp_bytes = vec![];
+            _codec.read_to_end(&mut rlp_bytes)?;
```
enzoc4, line 219: `read_to_end` is what causes the deser errors, since the rest of the transaction fields are _after_ the payload, and this is consuming everything. So the expiry date is not really invalid, it's just that there isn't anything else to read on the buffer.

```suggestion
            let len = _codec.get_be_u64()?;
            let rlp_bytes = _codec.get_bytes(len as usize)?;
```

enzoc4 (commented):
Maybe `u64` is more than needed here (I don't have a clue what `rlp` is), but this seems to fix the test failure.

chain-impl-mockchain/src/evm/mod.rs
saibatizoku: Thanks. RLP is the encoding that Ethereum uses, in hashes that are 256-bit long. I think that `u64` len should be enough.

Mr-Leshiy:
@saibatizoku , mostly looks good, but we need to ensure that this new serde implementation for the EvmTransaction converges to the original Ethereum implementation that is the original goal of implementing Rlp encoding.
So need to add additional test with the hardcoded EvmTransaction hex encoded data.
https://www.trustology.io/insights-events/decoding-an-ethereum-transaction-its-no-secret-its-just-smart

chain-impl-mockchain/src/evm/mod.rs
```diff
@@ -301,15 +256,43 @@ impl Payload for EvmTransaction {
...
+    };
     use quickcheck::Arbitrary;
 
     impl Arbitrary for EvmTransaction {
         fn arbitrary<G: quickcheck::Gen>(g: &mut G) -> Self {
             let caller = [u8::arbitrary(g); H160::len_bytes()].into();
             let value = u128::arbitrary(g).into();
             let gas_limit = Arbitrary::arbitrary(g);
-            let access_list = Vec::new();
+            let access_list: AccessList = match u8::arbitrary(g) % 5 {
```
Mr-Leshiy, line 270:
```suggestion
            let access_list: AccessList = match u8::arbitrary(g) % 2 {
```

chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -366,9 +366,14 @@ impl Ledger {
...
             token_totals: TokenTotals::default(),
         };
+        #[cfg(not(feature = "evm"))]
+        {
+            ledger
+        }
         #[cfg(feature = "evm")]
-        let ledger = ledger.set_evm_block0().set_evm_environment();
-        ledger
+        {
```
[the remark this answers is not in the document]
saibatizoku: Clippy complained, so I removed the lint. I think it's not that bad looking, and explicitly handles the `not(` feature, which can't hurt.

saibatizoku:
> @saibatizoku , mostly looks good, but we need to ensure that this new serde implementation for the EvmTransaction converges to the original Ethereum implementation that is the original goal of implementing Rlp encoding. So need to add additional test with the hardcoded EvmTransaction hex encoded data. https://www.trustology.io/insights-events/decoding-an-ethereum-transaction-its-no-secret-its-just-smart

I agree, there are also some possible test values in the `ethereum` crate that we might use. I'll take a look, in any case, adding tests will have to be a new PR, this one has grown too long already.

Mr-Leshiy (approved):

chain-evm/src/transaction.rs
```diff
@@ -0,0 +1,49 @@
...
+
+impl EthereumTransaction {
+    /// Serialize the contract into a `ByteBuilder`.
+    pub fn serialize_in(&self, bb: ByteBuilder<Self>) -> ByteBuilder<Self> {
+        let bytes = self.rlp_bytes();
+        bb.u64(bytes.len() as u64).bytes(&bytes)
+    }
+}
+
+impl Decodable for EthereumTransaction {
```
Mr-Leshiy, line 21: I think in case that we do not want to propagate all these Ethereum traits as `Decodable` and `Encodable`, I would prefer to add just a `from_bytes(&[u8]) -> Self` and `to_bytes(self) -> &[u8]` methods for the `EthereumTransaction` it seems to be more convenient for the further usage

chain-evm/src/transaction.rs
```diff
@@ -0,0 +1,49 @@
...
+    }
+}
+
+impl Encodable for EthereumTransaction {
+    fn rlp_append(&self, s: &mut RlpStream) {
+        self.0.rlp_append(s);
+    }
+}
+
+impl Serialize for EthereumTransaction {
```
Mr-Leshiy, line 34: Actually I am not sure that these trait implementation as `Serialize`, `Deserialize` will also be needed for us, maybe we can just remove it at this point ?

chain-evm/src/transaction.rs
```diff
@@ -0,0 +1,49 @@
...
+    packer::Codec,
+    property::{Deserialize, ReadError, Serialize, WriteError},
+};
+use ethereum::TransactionV2;
+use rlp::{decode, Decodable, DecoderError, Encodable, Rlp, RlpStream};
+use typed_bytes::ByteBuilder;
+
+/// Wrapper type for `ethereum::TransactionV2`, which includes the `EIP1559Transaction`, `EIP2930Transaction`, `LegacyTransaction` variants.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct EthereumTransaction(TransactionV2);
```
Mr-Leshiy, line 11: I assume we should also add implementation of the `TryFrom`/`TryInto` for the conversion into the `EvmTransaction` type

chain-evm/src/transaction.rs
saibatizoku: I think that we need these traits for Fragment-related ser/de, but I'm not sure.

chain-evm/src/transaction.rs
saibatizoku: I'll add these functions, although the `Serialize`/`Deserialize` traits end up providing the same, and make use of the Ethereum traits under the hood, so I think that it's equivalent.

chain-impl-mockchain/src/evm/mod.rs
```diff
@@ -47,133 +48,66 @@ pub enum EvmTransaction {
...
-    }
-}
-
-#[cfg(feature = "evm")]
-impl Encodable for EvmTransaction {
-    fn rlp_append(&self, s: &mut RlpStream) {
-        use EvmTransaction::*;
+impl EvmTransaction {
+    /// Serialize the contract into a `ByteBuilder`.
+    pub fn serialize_in(&self, _bb: ByteBuilder<Self>) -> ByteBuilder<Self> {
```
Mr-Leshiy, line 57: I would prefer to implement `Serialize` trait, it will allow to use `serialization_bijection` function in tests and it will follow to the full serde interface. (as `Deserialize` trait has been implemented)

Mr-Leshiy (changes requested):

chain-impl-mockchain/src/evm/mod.rs
saibatizoku: I restored the `Serialize` trait, and kept the RLP encoding because it handles things like `U256` values and list of items better than we were doing it. If anything, I would move the EvmTransaction type into `chain-evm` to clean up dependencies a bit, but that would be in a future PR.

Mr-Leshiy (approved):

PR #789, opened 2022-04-04
remove txbuilder

cameron1024:
This whole file (including tests) were never being compiled, because they weren't included in the module tree. It's 2-3 years old and out of date, so probably best to remove it


enzoc4 (approved):
And we have the builder in `transaction/builder.rs`. I guess this is a previous version of that.

PR #790, opened 2022-04-04
Add transactions related scenarios

nicopado:
Adding scenario tests:
valid transaction
invalid transaction
correct fees when casting votes
correct fees when tallying


chain-impl-mockchain/src/testing/e2e/fees.rs
```diff
@@ -239,3 +244,93 @@ fn verify_total_funds_after_transaction_with_fee(fee: u64) {
         .pots()
         .has_fee_equals_to(&Value(fee));
 }
+
+#[quickcheck]
+pub fn vote_cast_fees(constant_fee: u64, coefficient: u64, certificate: u64) {
+    let expected_fees = constant_fee + coefficient + certificate;
+    let favorable = Choice::new(1);
+    let voting_token = TokenName::try_from(vec![0u8; TOKEN_NAME_MAX_SIZE]).unwrap();
```
dkijania, line 252: i think we got `TestGen::token_name()` static method for this purpose

chain-impl-mockchain/src/testing/e2e/fees.rs
```diff
@@ -239,3 +244,93 @@ fn verify_total_funds_after_transaction_with_fee(fee: u64) {
         .pots()
         .has_fee_equals_to(&Value(fee));
 }
+
+#[quickcheck]
+pub fn vote_cast_fees(constant_fee: u64, coefficient: u64, certificate: u64) {
```
dkijania, line 249: I'm pretty sure that `LinearFee` implements Arbitrary, so you can write directly:

`pub fn vote_cast_fees(linear_fee: LinearFee) {`

chain-impl-mockchain/src/testing/e2e/fees.rs
```diff
@@ -239,3 +244,93 @@ fn verify_total_funds_after_transaction_with_fee(fee: u64) {
...
+    alice.confirm_transaction();
+
+    LedgerStateVerifier::new(ledger.clone().into())
+        .info("fee pot is filled with expected fees amount")
+        .pots()
+        .has_fee_equals_to(&Value(expected_fees));
+
+    LedgerStateVerifier::new(ledger.into())
+        .info("account balance is correct")
+        .address_has_expected_balance(alice.as_account_data(), Value(1000 - expected_fees));
```
dkijania, line 292: It would be better to assign some variable to express 1000. Because when reading test it's not 100% clear that this 1000 is initial account balance or initial rewards

dkijania: And it would also be better to have different values for :

- rewards
- voting tokens
- balance

to avoid any confusion

chain-impl-mockchain/src/testing/e2e/fees.rs
```diff
@@ -239,3 +244,93 @@ fn verify_total_funds_after_transaction_with_fee(fee: u64) {
...
+}
+
+#[quickcheck]
+pub fn vote_tally_fees(constant_fee: u64, coefficient: u64, certificate: u64) {
```
dkijania, line 296: the same as in line 254. you can use `LinearFee`

chain-impl-mockchain/src/testing/e2e/fees.rs
```diff
@@ -239,3 +244,93 @@ fn verify_total_funds_after_transaction_with_fee(fee: u64) {
...
+    let expected_fees = constant_fee + coefficient + certificate;
```
dkijania, line 297: And as a result expected `expected_fees` can be calculated :
```
let expected_fees = linear_fee.constant + linear_fee.coefficient +  linear_fee.certificate
```

chain-impl-mockchain/src/testing/e2e/fees.rs
```diff
@@ -239,3 +244,93 @@ fn verify_total_funds_after_transaction_with_fee(fee: u64) {
...
+    alice.confirm_transaction();
+
+    LedgerStateVerifier::new(ledger.clone().into())
+        .info("fee pot is filled with expected fees amount")
+        .pots()
+        .has_fee_equals_to(&Value(expected_fees));
+
+    LedgerStateVerifier::new(ledger.into())
+        .info("account balance is correct")
+        .address_has_expected_balance(alice.as_account_data(), Value(1000 - expected_fees));
```
dkijania, line 335: the same as comment in line 292

chain-impl-mockchain/src/testing/e2e/transactions.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    controller
+        .transfer_funds(&alice, &bob, &mut ledger, amount + total_fees)
+        .unwrap();
+    alice.confirm_transaction();
+
+    LedgerStateVerifier::new(ledger.into())
+        .address_has_expected_balance(bob.as_account_data(), Value(bob_initial_balance + amount));
+
+    TestResult::passed()
```
dkijania, line 51: i think, you don't need to explicitly set TestResult to pass. It would be pass automatically if none of assertion failed etc.

chain-impl-mockchain/src/testing/e2e/transactions.rs
```diff
@@ -0,0 +1,91 @@
...
+    let mut rng = rand::thread_rng();
+
+    let (fee, total_fees) = (LinearFee::new(1, 1, 1), 3);
+    let (alice_initial_balance, bob_initial_balance) =
+        (amount + total_fees + 1, rng.gen_range(1..10));
+
+    let (mut ledger, controller) = prepare_scenario()
+        .with_config(
+            ConfigBuilder::new()
+                .with_discrimination(Discrimination::Test)
```
dkijania, line 30: is test discrimination important for this test case? If not, it's better to remove it, because reader can think that it has something to do with fees

chain-impl-mockchain/src/testing/e2e/transactions.rs
```diff
@@ -0,0 +1,91 @@
...
+        .unwrap();
+
+    let alice = controller.wallet("Alice").unwrap();
+    let bob = controller.wallet("Bob").unwrap();
+
+    controller
+        .transfer_funds(&alice, &bob, &mut ledger, amount + total_fees)
+        .unwrap();
+
+    TestResult::from_bool(
```
dkijania, line 86: It would be good to add check that alice bob and fees are the same after and before this transaction, as test cases says `validate state after invalid transaction`

chain-impl-mockchain/src/testing/e2e/transactions.rs
```diff
@@ -0,0 +1,91 @@
...
+use quickcheck_macros::quickcheck;
+use rand::Rng;
+
+#[quickcheck]
+pub fn validate_ledger_state_after_transaction(amount: u64) -> TestResult {
+    if amount == 0 {
+        return TestResult::discard();
+    };
+
+    let mut rng = rand::thread_rng();
```
dkijania, line 21: it is not very quickcheckish way of getting random number. I would write a bit more code but it would be nicely incorporated into framework.

First let's create arbitrary for u64 with value constraint (1..10):
```
struct Random1to10(u64);
```
and then implement arbitrary
```
impl Arbitrary for Random1to10 {
            Self(u64::arbitrary(g) % 10 + 1)
     }
}
```

Finally use it in quickcheck:

```
pub fn validate_ledger_state_after_transaction(amount: u64, random: Random1to10 ) -> TestResult {
```

chain-impl-mockchain/src/testing/e2e/transactions.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    let (fee, total_fees) = (LinearFee::new(1, 1, 1), 3);
+    let (alice_initial_balance, bob_initial_balance) =
+        (amount + total_fees + 1, rng.gen_range(1..10));
```
dkijania, line 25: magic number (1), please name it so we know why it's 1

chain-impl-mockchain/src/testing/e2e/transactions.rs
```diff
@@ -0,0 +1,91 @@
...
+
+#[quickcheck]
+pub fn validate_ledger_state_after_transaction(amount: u64) -> TestResult {
+    if amount == 0 {
+        return TestResult::discard();
+    };
+
+    let mut rng = rand::thread_rng();
+
+    let (fee, total_fees) = (LinearFee::new(1, 1, 1), 3);
```
dkijania, line 23: maybe linear fee would also be arbitrary value ? and another magic number : 3

chain-impl-mockchain/src/testing/e2e/transactions.rs
```diff
@@ -0,0 +1,91 @@
...
+
+#[quickcheck]
+pub fn validate_ledger_state_after_invalid_transaction(amount: u64) -> TestResult {
+    if amount == 0 {
+        return TestResult::discard();
+    };
+
+    let mut rng = rand::thread_rng();
+
+    let (fee, total_fees) = (LinearFee::new(1, 1, 1), 3);
```
dkijania, line 62: the same as in method above:
- linear fees 
- random arbitrary number
- no magic numbers

dkijania (changes requested):

dkijania:
Overall code looks good. It just need to have some readability improvements

chain-impl-mockchain/src/testing/e2e/fees.rs
nicopado: Not working for this purpose as TestGen::token_name() returns an incompatible type of token.
As discussed I created a ticket to track this desiderata: https://input-output.atlassian.net/browse/NPG-1217

chain-impl-mockchain/src/testing/e2e/fees.rs
nicopado: not working as is, implemented as:
`let expected_fee = linear_fee.calculate(None, 1, 1)`
after review

chain-impl-mockchain/src/testing/e2e/transactions.rs
nicopado: As discussed, fee checks are already covered by tests in 
`chain-impl-mockchain/src/testing/e2e/fee.rs`

chain-impl-mockchain/src/testing/e2e/transactions.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    LedgerStateVerifier::new(ledger.into()).address_has_expected_balance(
+        bob.as_account_data(),
+        Value(bob_initial_balance + valid_transaction_amount),
+    );
+}
+
+#[quickcheck]
+pub fn validate_ledger_state_after_invalid_transaction(amount: Random1to10, linear_fee: LinearFee) {
+    let total_fees = linear_fee.constant + linear_fee.coefficient + linear_fee.certificate;
```
dkijania, line 53: Shouldn't this be the same like in LINE: 18

dkijania (approved):

PR #794, opened 2022-04-05
Update add_account() function interface

Mr-Leshiy:
Remove implicit `clone()` execution inside, so in some places we can get rid of uneeded copy operations.


Mr-Leshiy:
@enzoc4 , Will try do it but maybe in the separate PR

PR #804, opened 2022-05-06
try cargo nextest in CI

cameron1024:


cameron1024:
somehow slower than cargo test 🤷

PR #808, opened 2022-05-16
Evm mapping scenario test

kukkok3:


chain-impl-mockchain/src/testing/e2e/evm_mapping.rs
```diff
@@ -0,0 +1,56 @@
+use crate::testing::scenario::wallet;
+use crate::testing::TestGen;
+use crate::testing::{scenario::prepare_scenario, verifiers::LedgerStateVerifier, ConfigBuilder};
+use chain_evm::Config;
+
+const ALICE: &str = "Alice";
+const BOB: &str = "Bob";
+
+#[test]
+pub fn evm_mapping() {
```
dkijania, line 10: this is very good test but i think it's testing two things:

- user can send correct evm mapping tx sending
- user cannot override evm mapping once its set

IMHO we should have 2 test cases

chain-impl-mockchain/src/testing/verifiers/ledger_verifier.rs
```diff
@@ -189,6 +192,26 @@ impl LedgerStateVerifier {
     pub fn votes(&self) -> VotesVerifier {
         VotesVerifier::new(self.ledger.active_vote_plans())
     }
+
+    #[cfg(feature = "evm")]
+    pub fn is_mapped_to_evm(&self, evm_mapping: &EvmMapping) -> &Self {
```
dkijania, line 197: Nitpick:

It would be great if evm related verifications would be done on child level of LedgerVerifier, as this struct was mainly design to just be a container of child verifiers. In essence i would like to call:

```
  let ledger_verifier = ...;
  ledger_verifier.evm().is_mapped(bob_mapping)
```

and LedgerVerifier should have sub verifier for emv only:

```
 impl LedgerVerifier {
  
 ...
 
 
 pub fn evm(self) -> EvmVerifier {
    EvmVerifier::new(...)
 }
```
So we will be prepared for adding more evm related verifications.

dkijania: Also we will can remove one `#[cfg(feature = "evm")]` for free!

dkijania (approved):

dkijania:
good stuff!

Mr-Leshiy (commented):
Just want to clarify, what is the difference of these test cases with the tests which we have there https://github.com/input-output-hk/chain-libs/blob/b6590318da760d223022bf4e3b1c42e1ee697242/chain-impl-mockchain/src/ledger/evm.rs#L532 ?

chain-impl-mockchain/src/testing/ledger.rs
```diff
@@ -104,7 +104,7 @@ impl ConfigBuilder {
             consensus_version: ConsensusVersion::Bft,
             transaction_max_expiry_epochs: None,
             #[cfg(feature = "evm")]
-            evm_params: chain_evm::Config::Istanbul,
+            evm_params: chain_evm::Config::default(),
```
Mr-Leshiy, line 107: Maybe we should also need to somehow configure evm config, and run tests cases on the different versions

kukkok3: The tests in [chain-libs/chain-impl-mockchain/src/ledger/evm.rs](https://github.com/input-output-hk/chain-libs/blob/b6590318da760d223022bf4e3b1c42e1ee697242/chain-impl-mockchain/src/ledger/evm.rs#L532) are unit tests. This ones instead are scenario test, "one level up" of those, and meant to test that all the create fragment and apply/run transaction works as well

Yes we have a way to configure the evm config here https://github.com/input-output-hk/chain-libs/blob/7f94a61365c36a30d23108e6ec81780729dff990/chain-impl-mockchain/src/testing/e2e/evm_mapping.rs#L16 
I didnt make different tests for each evm config because in the case of evm_mapping we dont use any of the evm parameters, but it will probably be the case when we test evm_transaction

Mr-Leshiy: yea, sure. sorry I have confused with the `EvmTransaction`, indeed for the `EvmMapping` we don't actually need to test on different evm configs. Sorry 😅

kukkok3: No worries, better safe than sorry 😄

PR #809, opened 2022-05-18
Add SignedEvmMapping type

saibatizoku:
Adds support for signing `EvmMapping`s, and verifying that the signing key corresponds to the `evm_address` field.

### DIFF
--- a/chain-impl-mockchain/src/certificate/evm_mapping.rs
+++ b/chain-impl-mockchain/src/certificate/evm_mapping.rs
@@ -15,7 +15,7 @@ use chain_evm::{
     },
     ethereum::TransactionSignature,
     ethereum_types::{H256, U256},
-    rlp::{self, Decodable, DecoderError, Encodable, Rlp, RlpStream},
+    rlp::{self, decode, Decodable, DecoderError, Encodable, Rlp, RlpStream},
     transaction::SIGNATURE_BYTES,
     util::{decode_h256_from_u256, sign_data_hash, Secret},
     Address, Error,
@@ -40,16 +40,6 @@ impl EvmMapping {
             account_id,
             evm_address,
         }
-    }
-
-    #[cfg(feature = "evm")]
-    pub fn evm_address(&self) -> &Address {
-        &self.evm_address
-    }
-
-    #[cfg(feature = "evm")]
-    pub fn account_id(&self) -> &Identifier {
-        &self.account_id
     }
 
     pub fn serialize_in(&self, bb: ByteBuilder<Self>) -> ByteBuilder<Self> {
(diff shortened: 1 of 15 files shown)


chain-impl-mockchain/src/certificate/evm_mapping.rs
```diff
@@ -127,6 +163,101 @@ impl DeserializeFromSlice for EvmMapping {
...
+    fn rlp_append(&self, s: &mut RlpStream) {
+        s.begin_list(2);
+        s.append(&U256::from_big_endian(self.account_id.as_ref().as_ref()));
+        s.append(&self.evm_address);
+    }
+}
+
+/// Represents a signed mapping between a Jormungandr account and an EVM account.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SignedEvmMapping {
```
Mr-Leshiy, line 197: Why is it needed to create a separate struct ? Do we really need signed and unsigned types of the `EvmMapping` transaction ? I suppose that we can leave only `SignedEvmMapping`.

chain-impl-mockchain/src/certificate/evm_mapping.rs
```diff
@@ -127,6 +166,123 @@ impl DeserializeFromSlice for EvmMapping {
...
```
Mr-Leshiy, line 200: It is better to reuse original `EvmMapping` for the `account_id` and `evm_address` fields

Mr-Leshiy, line 200: We also should replace the current usage of the `EvmMapping` type with the `SIgnedEvmMapping` inside the certificate

chain-impl-mockchain/src/certificate/evm_mapping.rs
```diff
@@ -127,6 +166,123 @@ impl DeserializeFromSlice for EvmMapping {
...
+            account_id: other.account_id.clone(),
+            evm_address: other.evm_address,
+        }
+    }
+}
+
+#[cfg(feature = "evm")]
+impl SignedEvmMapping {
+    /// Verifies that the signing key corresponds to the `evm_address`.
+    pub fn verify(&self) -> Result<bool, Error> {
```
Mr-Leshiy, line 226: We should use this validation inside `apply_map_accounts` fn (or any other place, but during the transaction processing inside the ledger)

Mr-Leshiy (changes requested):

chain-impl-mockchain/src/certificate/evm_mapping.rs
saibatizoku: This is done.

saibatizoku:
closing this PR in favor of https://github.com/input-output-hk/chain-libs/pull/832

PR #810, opened 2022-05-19
Add nonce field for the EvmTransaction

Mr-Leshiy:


chain-impl-mockchain/src/evm/mod.rs
```diff
@@ -115,56 +101,38 @@ impl Decodable for EvmTransaction {
...
-            Create {
-                caller,
-                value,
-                init_code,
-                gas_limit,
-                access_list,
-            } => {
+        match &self.action_type {
+            EvmActionType::Create { init_code } => {
                 s.begin_list(6);
```
Zeegomo, line 106: I think there are 7 values below, but I would expect this to show up in tests. Maybe it's not covered?

chain-impl-mockchain/src/ledger/evm.rs
```diff
@@ -192,8 +196,24 @@ impl<'a> EvmState for EvmStateImpl<'a> {
...
+        accounts: &account::Ledger,
+        transaction: &EvmTransaction,
+    ) -> Result<(), Error> {
+        let jor_address = evm.address_mapping.jor_address(&transaction.caller);
+        let expected_nonce = match accounts.get_state(&jor_address) {
+            Ok(account_state) => account_state.evm_state.nonce,
+            Err(LedgerError::NonExistent) => 0,
+            Err(_) => unimplemented!("other errors are not expected"),
+        };
+        (expected_nonce == transaction.nonce)
```
Zeegomo, line 210: nit: simple if would be more readable IMO

Mr-Leshiy: fixed

chain-impl-mockchain/src/evm/mod.rs
Mr-Leshiy: As I see, generally it will not a mistake to put less value than actually needed, It something similar as `Vec::with_capacity();`.

PR #829, opened 2022-06-28
Chain time improvment

kukkok3:
I added serde deserialize to a couple of struct in chain time because it is needed in the explorer. Also I fixed a unused import warning in chain-libs


kukkok3:
Found another way to make it work, probably this is not needed. Closing for now

PR #830, opened 2022-06-30
Refactor ledger config, remove redundant LedgerParameters

Mr-Leshiy:
We have LedgerParameters struct [chain-libs/ledger.rs at 8129ec87d68de7013c5c542ebae1136868912ab6 · input-output-hk/chain-libs](https://github.com/input-output-hk/chain-libs/blob/8129ec87d68de7013c5c542ebae1136868912ab6/chain-impl-mockchain/src/ledger/ledger.rs#L53)  and Settings struct [chain-libs/setting.rs at ff539c523cfd5b9dfd9c42ff128c60c3cbd25b27 · input-output-hk/chain-libs](https://github.com/input-output-hk/chain-libs/blob/ff539c523cfd5b9dfd9c42ff128c60c3cbd25b27/chain-impl-mockchain/src/setting.rs#L23)  . Those two structs have the same fields as 
```
    /// Fees expected for transactions and certificates
2    pub fees: LinearFee,
3    /// Tax cut of the treasury which is applied straight after the reward pot
4    /// is fully known
5    pub treasury_tax: rewards::TaxType,
6    /// Reward contribution parameters for this epoch
7    pub reward_params: rewards::Parameters,
8    /// the block content's max size in bytes
9    pub block_content_max_size: BlockContentSize,
10    /// the epoch stability parameter, the depth, number of blocks, to which
11    /// we consider the blockchain to be stable and prevent rollback beyond
12    /// that depth.
13    pub epoch_stability_depth: u32,
14    /// Where the fees get transfered to during the rewards
15    pub fees_goes_to: setting::FeesGoesTo,
16    /// List of committee members
17    pub committees: Arc<[CommitteeId]>,
```
and these two structs are fields of the ApplyBlockLedger [chain-libs/ledger.rs at 8129ec87d68de7013c5c542ebae1136868912ab6 · input-output-hk/chain-libs](https://github.com/input-output-hk/chain-libs/blob/8129ec87d68de7013c5c542ebae1136868912ab6/chain-impl-mockchain/src/ledger/ledger.rs#L101) , (Settings struct is a field of the Ledger struct) so it is possible to have an inconsistency of these values.

So intention is to simplify code, and remove duplicated `LedgerParameters` struct with its usage

### DIFF
--- a/chain-impl-mockchain/src/fee.rs
+++ b/chain-impl-mockchain/src/fee.rs
@@ -5,7 +5,7 @@ use std::num::NonZeroU64;
 
 /// Linear fee using the basic affine formula
 /// `COEFFICIENT * bytes(COUNT(tx.inputs) + COUNT(tx.outputs)) + CONSTANT + CERTIFICATE*COUNT(certificates)`.
-#[derive(PartialEq, Eq, PartialOrd, Debug, Clone, Copy)]
+#[derive(PartialEq, Eq, PartialOrd, Debug, Clone)]
 pub struct LinearFee {
     pub constant: u64,
     pub coefficient: u64,
--- a/chain-impl-mockchain/src/leadership/bft.rs
+++ b/chain-impl-mockchain/src/leadership/bft.rs
@@ -2,7 +2,6 @@ use crate::block::{BftProof, BlockDate, Header, Proof};
 use crate::{
     key::BftLeaderId,
     leadership::{Error, ErrorKind, Verification},
-    ledger::Ledger,
 };
 use std::sync::Arc;
 
@@ -17,13 +16,13 @@ pub struct LeadershipData {
 
 impl LeadershipData {
     /// Create a new BFT leadership
-    pub fn new(ledger: &Ledger) -> Option<Self> {
-        if ledger.settings.bft_leaders.len() == 0 {
+    pub fn new(leaders: &Arc<[BftLeaderId]>) -> Option<Self> {
+        if leaders.len() == 0 {
             return None;
         }
 
         Some(LeadershipData {
-            leaders: Arc::clone(&ledger.settings.bft_leaders),
+            leaders: Arc::clone(leaders),
         })
     }
 
@@ -88,7 +87,7 @@ mod tests {
     use crate::header::HeaderBuilderNew;
     use crate::leadership::tests::generate_ledger_with_bft_leaders;
     use crate::leadership::tests::generate_ledger_with_bft_leaders_count;
-    use crate::ledger::Pots;
+    use crate::ledger::{Ledger, Pots};
     use crate::setting::Settings;
     use crate::testing::data::AddressData;
     use crate::testing::TestGen;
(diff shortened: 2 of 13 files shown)


chain-impl-mockchain/src/fee.rs
```diff
@@ -5,7 +5,7 @@ use std::num::NonZeroU64;
 
 /// Linear fee using the basic affine formula
 /// `COEFFICIENT * bytes(COUNT(tx.inputs) + COUNT(tx.outputs)) + CONSTANT + CERTIFICATE*COUNT(certificates)`.
-#[derive(PartialEq, Eq, PartialOrd, Debug, Clone, Copy)]
+#[derive(PartialEq, Eq, PartialOrd, Debug, Clone)]
```
Zeegomo, line 8: why?

chain-impl-mockchain/src/leadership/mod.rs
```diff
@@ -204,12 +201,6 @@ impl Leadership {
         &self.inner
     }
 
-    /// access the ledger parameter for the current leadership
-    #[inline]
-    pub fn ledger_parameters(&self) -> &LedgerParameters {
```
Zeegomo, line 209: I think this could be useful to calculate in advance the leadership schedule without having an actual ledger available, have you thought about this?

chain-impl-mockchain/src/setting.rs
```diff
@@ -310,6 +307,10 @@ impl Settings {
         params
     }
 
+    pub fn to_treasury_params(&self) -> TaxType {
```
Zeegomo, line 310: remove `to_`, I think the convention around `to_` methods is that they take `self`

chain-impl-mockchain/src/fee.rs
Mr-Leshiy: I am not in strong position with this change, but IMO `LinearFee` seems a complex structure with a lot of fields, so I would prefer to explicitly `clone` it where it is needed, so to have a better view how we can optimize our code.

chain-impl-mockchain/src/leadership/mod.rs
Mr-Leshiy: Good point, need to think how it is better organize, taking a look...

Mr-Leshiy: Updated, I have modified `LeadershipData` generation, but `Leadership` remain the same, if we will pass all this arguments separately will have `clippy:: too_many_arguments` warning.

chain-impl-mockchain/src/leadership/bft.rs
```diff
@@ -17,13 +16,13 @@ pub struct LeadershipData {
 
 impl LeadershipData {
     /// Create a new BFT leadership
-    pub fn new(ledger: &Ledger) -> Option<Self> {
-        if ledger.settings.bft_leaders.len() == 0 {
+    pub fn new(leaders: &Arc<[BftLeaderId]>) -> Option<Self> {
```
Zeegomo, line 19: nit: removing the `&` in `&Arc` would avoid double clone if the caller already had an `Arc`

Zeegomo (approved):

chain-impl-mockchain/src/leadership/bft.rs
Mr-Leshiy: updated

PR #831, opened 2022-07-15
fix and refactor validate_ledger test

kukkok3:
Fixing validate_ledger_test. We initialized alice balance with amount.0 + valid_transaction_amount and valid_transaction_amount is equal to total_fees.0 + amount.0. So looks like alice got twice the amount we are expected in test. Fixed sending all the initial alice balance in the first transaction, leaving alice with 0 balance, and therefore making the second trasaction fail


Mr-Leshiy (approved):
LGTM

